PR #1, opened 2017-09-15
Fix test for composeAndComposeSimple function

psibi:
Not the best way to fix test code and also involves the usage of
partial function, but I guess that's fine for test code.


harendra-kumar:
Looks good. Thanks!

PR #6, opened 2018-03-10
Prelude functions

Abhiroop:
Added a couple of prelude functions: `scanl` and `tail`

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -1,10 +1,8 @@
 {-# LANGUAGE BangPatterns              #-}
 {-# LANGUAGE FlexibleContexts          #-}
 {-# LANGUAGE FlexibleInstances         #-}
-{-# LANGUAGE GeneralizedNewtypeDeriving#-}
 {-# LANGUAGE MultiParamTypeClasses     #-}
 {-# LANGUAGE RankNTypes                #-}
-{-# LANGUAGE StandaloneDeriving        #-}
 {-# LANGUAGE UndecidableInstances      #-} -- XXX
 
 -- |
@@ -20,8 +18,9 @@
 module Streamly.Prelude
     (
     -- * Construction
-      cons
-    , nil
+      nil
+    , cons
+    , (.:)
     , unfoldr
     , unfoldrM
     , each
@@ -30,6 +29,7 @@ module Streamly.Prelude
     -- ** General Folds
     , foldr
     , foldrM
+    , scan
     , foldl
     , foldlM
     , uncons
@@ -73,7 +73,7 @@ module Streamly.Prelude
     )
 where
 
-import           Control.Monad               (liftM)
+import           Control.Monad (void)
 import           Control.Monad.IO.Class      (MonadIO(..))
 import           Data.Semigroup              (Semigroup(..))
 import           Prelude hiding              (filter, drop, dropWhile, take,
@@ -82,7 +82,7 @@ import           Prelude hiding              (filter, drop, dropWhile, take,
                                               sum, product, elem, notElem,
                                               maximum, minimum, head, last,
                                               length)
-import qualified Prelude as Prelude
+import qualified Prelude
 import qualified System.IO as IO
 
 import           Streamly.Core
@@ -96,7 +96,7 @@ import           Streamly.Streams
 unfoldr :: Streaming t => (b -> Maybe (a, b)) -> b -> t m a
 unfoldr step = fromStream . go
     where
-    go s = Stream $ \_ stp yld -> do
+    go s = Stream $ \_ stp yld ->
         case step s of
             Nothing -> stp
             Just (a, b) -> yld a (Just (go b))
@@ -115,11 +115,11 @@ unfoldrM step = fromStream . go
 -- | Same as @foldWith (<>)@ but more efficient.
 {-# INLINE each #-}
 each :: (Streaming t, Foldable f) => f a -> t m a
-each xs = Prelude.foldr cons nil xs
+each = Prelude.foldr cons nil
 
 -- | Read lines from an IO Handle into a stream of Strings.
 fromHandle :: (Streaming t, MonadIO m) => IO.Handle -> t m String
-fromHandle h = fromStream $ go
+fromHandle h = fromStream go
   where
   go = Stream $ \_ stp yld -> do
         eof <- liftIO $ IO.hIsEOF h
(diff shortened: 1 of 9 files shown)


src/Streamly/Prelude.hs
```diff
@@ -303,6 +315,10 @@ head m =
         yield a _ = return (Just a)
     in (runStream (toStream m)) Nothing stop yield
 
+-- | Returns the stream minus the head.
+tail :: Streaming t => t m a -> t m a
```
Abhiroop: The `tail` might be unnecessary. I was thinking of the standard Prelude. Just checked Pipes Prelude and Conduits, nothing like this exists.

harendra-kumar:
`tail` will have to either return a `Maybe` or will be a partial function when the stream is empty. Let's not have `tail` for now or at least in this commit. Let me take a look at the `scanl`.

src/Streamly/Prelude.hs
```diff
@@ -166,6 +168,16 @@ foldl step begin done m = go begin (toStream m)
             yield a (Just x) = go (step acc a) x
          in (runStream m1) Nothing stop yield
 
+-- | Scan left. A strict left fold which accumulates the result of its reduction steps inside a stream, from left.
+scanl :: (Streaming t, Monad m) => (b -> a -> b) -> b -> t m a -> t m b
```
harendra-kumar: Let's make the signature on the lines of `foldl`, having a `done` function (`x -> b` in foldl) as well so that this works well with the `foldl` package. When writing a new prelude functions it may be a good idea to take a look at how the `streaming` package types it or implements it. `streaming` is pretty close to `streamly`.

src/Streamly/Prelude.hs
```diff
@@ -166,6 +168,16 @@ foldl step begin done m = go begin (toStream m)
...
+scanl f seed m = cons seed $ fromStream $ go (toStream m) seed
```
harendra-kumar: Name the args on the lines of `foldl` i.e. `step`, `begin`, `done` and `m`. `scanl` and `foldl` are almost the same except that scanl keeps the intermediate steps.

src/Streamly/Prelude.hs
```diff
@@ -166,6 +168,16 @@ foldl step begin done m = go begin (toStream m)
...
+    where
+    go m1 !s = Stream $ \_ stp yld ->
```
harendra-kumar: call it `acc` (accumulator) instead of `s` to be consistent with `foldl` implementation.

src/Streamly/Prelude.hs
```diff
@@ -166,6 +168,16 @@ foldl step begin done m = go begin (toStream m)
...
+        let stop = stp
+            yield a Nothing  = return (f s a) >> stp
```
harendra-kumar: The semantics of `yield a Nothing` is equivalent to yielding `a` and then calling `stp` so we should not call `stp` explicitly in this case. In general, the stream can end in two ways, by calling `stp` or by yielding `Nothing` as the next element.

src/Streamly/Prelude.hs
```diff
@@ -166,6 +168,16 @@ foldl step begin done m = go begin (toStream m)
...
+            yield a (Just x) = return (f s a) >>= \b -> yld b (Just (go x (f s a)))
```
harendra-kumar: This can be written as `yld (f s a) (Just (go...`

harendra-kumar:
Great way to start @Abhiroop ! Thanks for the PR.

I just pushed a few pending changes and it includes some changes to tests. You can write a test for `scanl` very easily (just one line) by just comparing the results with the Data.List `scanl`. See `transformOps` function in `test/Main.hs`.

harendra-kumar:
I just added a test for `foldl`, see at the end of test/Main.hs. scanl test should be similar.

harendra-kumar:
BTW, to run perf tests using your local streamly repo you will have to add the local package location like this in the `stack.yaml` of `streaming-benchmarks`:

```
packages:
- '.'
- location: ../streamly
  extra-dep: true
```

Abhiroop:
I ran an `hindent` pass on the `Prelude.hs` and it added a bunch of new indentations. Let me know if you are fine with it or I will revert it. I am using the Johan Tibell guide for hindent, let me know if you are running any other guide then I will change to that for this project.

harendra-kumar:
Some procedural stuff, in general:

* Never mix different functionalities in a single commit, it may be ok in separate commits the same PR but separate PRs are preferred. This is for ease of reviewing and change tracking later in the history.

* Styling changes (hindent) should always be separate. It becomes difficult to reason about what changed and there are a bunch of unrelated changes lumped together. Just change what is necessary for that one logical change. Other changes if necessary can be committed separately.

* I do not use fully automatic indenting on the whole file, I do not like all the changes done by hindent. Some of them are not visually appealing. I mostly indent manually and use hindent sometimes on selected lines as a tooling aid to perform it. For example, if you want to indent the scan function, in vim you can visually select a number lines (using Shit-v and moving the cursor down or up) and then use "gq" to format (assuming you have set the formatprog to hindent with Johan Tibell style). This is what I use in vim:

```
let &formatprg="hindent --style johan-tibell --line-length 79 --indent-size 4 --sort-imports -XQuasiQuotes"
```
* The hindent changes have changed everything to 2 space indentation whereas all of the project is formatted using 4 space indentation, that is my preferred style, clearer to see than 2 space.

* I see that you merged master in your branch. Usually what you do is a rebase instead of a merge. Use `git rebase master` when on your branch to rebase your changes on top of master.

harendra-kumar:
Todo for this PR: undo the hindent changes, squash all your changes into a single commit using `git rebase -i`. Everything else (scan implementation and test) looks fine.

If you want to hindent the whole file you can push that as a separate change, I can review it and we can discuss if there is anything specifically that I may not like in it. In general the hindent style and my manual style mostly agrees. Just make sure to use 4 spaces and 79 columns rule. I guess we can even use a hindent config file to make some exceptions according to our taste, not sure if hindent allows it, I think stylish-haskell allows it.

harendra-kumar:
The indentation for scan is still inconsistent with the rest of the file. It is using 2 spaces, change that to 4 spaces. Also, I start the `where` clause contents aligned in the same column as `where` itself. This saves some indentation which can be used better inside the contents of the clause. Like this:

```
    where

    x = ...
```

Also most likely we need an `INLINE` pragma on scan.

src/Streamly/Prelude.hs
```diff
@@ -155,6 +156,20 @@ foldrM step acc m = go (toStream m)
...
         in (runStream m1) Nothing stop yield
 
+-- | Scan left. A strict left fold which accumulates the result of its reduction steps inside a stream, from left.
+{-# INLINE scan #-}
+scan :: Streaming t => (x -> a -> x) -> x -> (x -> b) -> t m a -> t m b
+scan step begin done m = cons (done begin) $ fromStream $ go (toStream m) begin
+    where
+    go m1 !acc = Stream $ \_ stp yld ->
+        let
+          stop = stp
```
harendra-kumar: Need 4 space indent here as well as in the nested let clause. Also, I prefer to write multiline let clauses like this (not hard and fast rule though):

```
let stop = stp
    yield a Nothing = ...
```

The first line of code comes right after the `let` keyword.

Sorry for being a pain in the neck about style and indent :-)

Abhiroop: Oh! I actually write `let` like this too. I copied your style in the `foldl` function.

harendra-kumar:
Looks good to go. Can you squash it into a single commit? Did you check the effect of `INLINE` on performance benchmark? My guess is it would have improved but sometimes it worsens, so its better to check it.

Abhiroop:
Oh yes, there is a marked improvement after inlining :)

```
benchmarked elimination/scan/streamly
time                 27.43 ms   (26.80 ms .. 27.92 ms)
                     0.998 R²   (0.997 R² .. 0.999 R²)
mean                 28.01 ms   (27.60 ms .. 28.48 ms)
std dev              1.013 ms   (722.0 μs .. 1.341 ms)
```

harendra-kumar:
That's nice to see! I guess it is beating streaming now? Thanks for all the work!

Abhiroop:
@harendra-kumar I had some issues when squashing the merge. Github has a squash and merge feature right? Can you try that?

harendra-kumar:
I can do that. I think you are merging master into your branch instead of rebasing. You need to rebase your branch against master and then push just your change.

harendra-kumar:
When on your branch use `git rebase master` and it will replay your changes on top of master.

harendra-kumar:
Don't worry, I will take care of this one.

Abhiroop:
I think I was able to squash. @harendra-kumar Can you verify once?

harendra-kumar:
It has included a lot of commits, everything that went since you created your branch. I think the problem is you kept merging master into your branch instead of rebasing your branch and now it may be difficult to rebase as well because of the merges, I am not sure. One way is to discard this branch, paste your diff in a new branch and push that one.

harendra-kumar:
You can also use `git cherry-pick` to transfer the changes from old branch to the new one.

harendra-kumar:
I did a rebase on your branch, resolved some merge conflicts and finally the changes are rebased on master, now I can squash and commit, I guess.

harendra-kumar:
Merged! Thanks!

harendra-kumar:
This PR won't show it as merged as I rebased the changes before merging and merged it from command line.

Issue #11, opened 2018-03-11
Missing prelude operations


harendra-kumar:
Some of the `base/Data.List` operations that are missing and we can add to Streamly:

## Easy to implement

These operations are fairly easy or in some cases trivial or mechanical to implement. They can be modeled on the implementations of similar existing operations. I will be happy to help in bootstrapping if someone wants to pick up some of these.

basic:

- [ ] tail
- [ ] init
- [ ] null
- [ ] reverse
- [ ] intersperse

Infinite streams:

- [ ] iterate
- [ ] repeat
- [ ] replicate
- [ ] cycle

Splitting:

- [ ] splitAt
- [ ] span
- [ ] break
- [ ] stripPrefix

SubStreams:

- [ ] isPrefixOf
- [ ] isInfixOf
- [ ] isSuffixOf 
- [ ] isSubsequenceOf

Lookups:

- [ ] lookup
- [ ] find
- [ ] partition

Indices:

- [ ] elemIndex
- [ ] elemIndices
- [ ] findIndex
- [ ] findIndices

Strings:

- [ ] lines
- [ ] words
- [ ] unlines
- [ ] unwords

Folds:

- [ ] foldl1
- [ ] foldr1
- [ ] scanl1
- [ ] scanr1
- [ ] mapAccumL
- [ ] mapAccumR
- [ ] and
- [ ] or

Set operations/Ordered list operations/"By" operations:

- [ ] Set operations
- [ ] Ordered list operations
- [ ] "By" operations

## Advanced

Involving two layers of containers, we need to decide on how to represent layers (Foldable/MonoFoldable/Streams) in general before attempting these:

- [ ] concat
- [ ] concatMap
- [ ] intercalate
- [ ] transpose
- [ ] subsequences
- [ ] permutations
- [ ] group
- [ ] inits
- [ ] tails

Veladus:
I'd love to tackle this as my first PR to a Haskell project, but I'm totally busy this week. Would it be possible to leave this issue open for me.

harendra-kumar:
Surely @Veladus ! This is more of a task list here listing many operations to be implemented and each one could be a separate PR in itself. Take one at a time, there are plenty of them. Even if someone else takes a stab at some of them, its not going anywhere in a week's time.

harendra-kumar:
[...]

PR #23, opened 2018-03-13
Easiest Prelude functions

Veladus:
I added `Prelude.tail`, `Prelude.null` and `Prelude.reverse`.
Refers to #11

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -39,10 +39,13 @@ module Streamly.Prelude
     , all
     , any
     , head
+    , tail
     , last
+    , null
     , length
     , elem
     , notElem
+    , reverse
     , maximum
     , minimum
     , sum
@@ -82,7 +85,7 @@ import           Prelude hiding              (filter, drop, dropWhile, take,
                                               mapM, mapM_, sequence, all, any,
                                               sum, product, elem, notElem,
                                               maximum, minimum, head, last,
-                                              length)
+                                              tail, length, null, reverse)
 import qualified Prelude
 import qualified System.IO as IO
 
@@ -333,10 +336,25 @@ head m =
         yield a _ = return (Just a)
     in (runStream (toStream m)) Nothing stop yield
 
+-- | Extract all but the first element of the stream, if any.
+tail :: (Streaming t, Monad m) => t m a -> m (Maybe (t m a))
+tail m =
+    let stop             = return Nothing
+        yield _ Nothing  = return $ Just nil
+        yield _ (Just t) = return $ Just $ fromStream t
+    in (runStream (toStream m)) Nothing stop yield
+
 -- | Extract the last element of the stream, if any.
 {-# INLINE last #-}
 last :: (Streaming t, Monad m) => t m a -> m (Maybe a)
 last = foldl (\_ y -> Just y) Nothing id
+
+-- | Determine whether the stream is empty.
+null :: (Streaming t, Monad m) => t m a -> m Bool
+null m =
+    let stop      = return True
+        yield _ _ = return False
+    in (runStream (toStream m)) Nothing stop yield
 
 -- | Determine whether an element is present in the stream.
 elem :: (Streaming t, Monad m, Eq a) => a -> t m a -> m Bool
@@ -361,6 +379,19 @@ notElem e m = go (toStream m)
 -- | Determine the length of the stream.
 length :: (Streaming t, Monad m) => t m a -> m Int
 length = foldl (\n _ -> n + 1) 0 id
+
+-- | Returns the elements of the stream in reverse order. 
+-- The stream must be finite.
+reverse :: (Streaming t) => t m a -> t m a
+reverse m = fromStream $ go Nothing (toStream m)
+    where
+    go rev rest = Stream $ \svr stp yld ->
+        let stop = case rev of
+                Nothing ->  stp
+                Just str -> runStream str svr stp yld
+            yield a Nothing  = runStream (a `scons` rev) svr stp yld
+            yield a (Just x) = runStream (go (Just $ a `scons` rev) x) svr stp yld
+         in runStream rest svr stop yield
 
 -- XXX replace the recursive "go" with continuation
 -- | Determine the minimum element in a stream.
--- a/test/Main.hs
+++ b/test/Main.hs
@@ -7,6 +7,7 @@ import Control.Concurrent (threadDelay)
 import Control.Monad (replicateM)
 import Data.Foldable (forM_)
 import Data.List (sort)
+import Data.Maybe (fromJust)
 import Test.Hspec
 
 import Streamly
@@ -675,7 +676,9 @@ streamOperations (stream, list, len) = do
     it "dropWhile < some" $ transform (A.dropWhile (< (len `div` 2)))
                                       (dropWhile (< (len `div` 2)))
 
+    -- Transformations
     it "scan left"  $ transform (A.scan (+) 0 id) (scanl (+) 0)
+    it "reverse" $ transform A.reverse reverse
 
     -- Elimination
     it "foldl" $ elimination (A.foldl (+) 0 id) (foldl (+) 0)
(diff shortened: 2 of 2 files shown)


harendra-kumar:
That's pretty awesome @Veladus for a first Haskell PR! I will take better look at the changes tonight. In the meanwhile you can take a look at the `streamOperations` function at the bottom of `test/Main.hs` and add your two three lines there. It is trivial to add tests for these operations.

Also, a couple of tests are currently failing in the build (`product` and `last`). I will fix those soon or if you want to fix those as well you can ahead and fix. The `product` implementation is using 0 as a seed which is incorrect. The `last` implementation should be threading around the last value in the `go` function and return it in case the `stop` continuation is called.

👍 thank you!

src/Streamly/Prelude.hs
```diff
@@ -333,6 +336,14 @@ head m =
         yield a _ = return (Just a)
     in (runStream (toStream m)) Nothing stop yield
 
+-- | Extract all but the first element of the stream, if any.
+tail :: (Streaming t, Monad m) => t m a -> m (Maybe (t m a))
```
harendra-kumar: The return type should be `Maybe (t m a)`.

src/Streamly/Prelude.hs
```diff
@@ -343,6 +354,13 @@ last m = go (toStream m)
             yield _ (Just x) = go x
         in (runStream m1) Nothing stop yield
 
+-- | Determine wheter the stream is empty
```
harendra-kumar: wheter => whether, end with a full stop.

src/Streamly/Prelude.hs
```diff
@@ -367,6 +385,9 @@ notElem e m = go (toStream m)
 length :: (Streaming t, Monad m) => t m a -> m Int
 length = foldl (\n _ -> n + 1) 0 id
 
+reverse :: (Streaming t, Monad m) => t m a -> m (t m a)
```
harendra-kumar: -- | Returns the elements of the stream in a reverse order. The stream must be finite.

The return type must be `t m a`.

Abhiroop:
@Veladus you might want to add a test for the `reverse` function that you added. Its extremely simple to add tests. You can see the `streamOperations` function in `tests/Main.hs` and add just the following line: `it "reverse"  $ transform A.reverse reverse` which basically compares the result of the streamly reverse function with the Prelude reverse function.

src/Streamly/Prelude.hs
Veladus: Actually I don't think this is possible consider the following example. I only assumed that I can create a function `tail` with the given return type and thereby created a function of type `String -> Int` which gives the number of bytes in a file.
```haskell
import Data.Maybe
import Streamly
import Streamly.Prelude ((.:),nil)

-- tail (nil) ~ Nothing
-- tail (pure 3) ~ Just nil
-- tail (pure 3 <> m) ~ Just m
sTail :: Streaming t => t m a -> Maybe (t m a)
sTail = undefined

-- Therefore I can easily define null as follows
-- which already escapes the monadic context
sNull :: Streaming t => t m a -> Bool
sNull m = isNothing $ sTail m

-- And when I have this I can do interesting things.
-- e.g. defining length without a monadic result.
sLength :: Streaming t => t m a -> Int
sLength m = go 0 (Just m)
    where
        go _ Nothing = 0
        go acc (Just m)
            | sNull m = acc
            | otherwise = go (1+acc) (sTail m)
          
-- Now I can easily escape arbitrary data from its `IO` context.
-- Lets just get the number of bytes in a File
numBytes :: String -> Int
numBytes fName = sLength . serially $ byteStream
    where 
      byteStream :: StreamT IO Char
      byteStream = do
        bytes <- liftIO $ readFile fName
        let toStreaming = foldr (.:) nil
        toStreaming bytes

main = print "tail therefore can't be t m a -> Maybe (t m a)"
```

harendra-kumar: I am sorry to have said that, it is obviously incorrect to have the type as `Maybe (t m a)`.

src/Streamly/Prelude.hs
```diff
@@ -332,11 +335,26 @@ head m =
         yield a _ = return (Just a)
     in (runStream (toStream m)) Nothing stop yield
 
+-- | Extract all but the first element of the stream, if any.
+tail :: (Streaming t, Monad m) => t m a -> m (Maybe (t m a))
+tail m =
+    let stop             = return Nothing
+        yield _ Nothing  = return $ Just nil
```
harendra-kumar: It should be `return Nothing`, we do not want to return an empty stream as tail when there is no tail. I see you have missed the test for tail.

src/Streamly/Prelude.hs
```diff
@@ -361,6 +379,20 @@ notElem e m = go (toStream m)
 length :: (Streaming t, Monad m) => t m a -> m Int
 length = foldl (\n _ -> n + 1) 0 id
 
+-- | Returns the elements of the stream in reverse order. 
+-- The stream must be finite.
+reverse :: (Streaming t) => t m a -> t m a
+reverse m = fromStream $ go Nothing (toStream m)
+    where
+      go rev rest = Stream $ \s stp yld ->
```
harendra-kumar: Align this like elsewhere:
```
    where
    go
```

harendra-kumar: We can call the `Svar` as `svr` just like `stp` and `yld`, that will provide a bit more hint about what it is. Historically I called it `ctx` everywhere but we can change that now.

src/Streamly/Prelude.hs
```diff
@@ -361,6 +379,20 @@ notElem e m = go (toStream m)
...
+          let
```
harendra-kumar: make the formatting like this `let stop = case rev of`

src/Streamly/Prelude.hs
```diff
@@ -361,6 +379,20 @@ notElem e m = go (toStream m)
...
+              stop             = case rev of
+                  Nothing -> runStream snil s stp yld
```
harendra-kumar: Why can't this be just `Nothing -> stp`?

src/Streamly/Prelude.hs
```diff
@@ -361,6 +379,20 @@ notElem e m = go (toStream m)
...
+                  Just str -> runStream str s stp yld
+              yield a Nothing  = runStream (a `scons` rev) s stp yld
+              yield a (Just x) = runStream (go (Just $ a `scons` rev) x) s stp yld
+           in runStream rest Nothing stop yield
```
harendra-kumar: We need to pass on the `SVar` `s` here instead of using `Nothing`. I need to figure out a way to add quick tests for such concurrency related stuff.

test/Main.hs
```diff
@@ -668,6 +668,9 @@ streamOperations (stream, list, len) = do
 
     it "scan left"  $ transform (A.scan (+) 0 id) (scanl (+) 0)
 
+    -- Transformations
```
harendra-kumar: `scan left` above is also a transformation.

test/Main.hs
```diff
@@ -686,16 +689,39 @@ streamOperations (stream, list, len) = do
...
 
     where
+    expectedTail [] = Nothing
+    expectedTail (_:xs) = Just xs
 
     -- XXX run on empty stream as well
     transform streamOp listOp =
         (A.toList $ streamOp stream) `shouldReturn` listOp list
 
+    transformMaybe streamOp listOp = do
```
harendra-kumar: How about something like this?
```
case list of
    _ : _ : _ -> it "tail" $ A.tail stream >>= toListSerial . fromJust `shouldReturn` tail list 
    _ -> it "tail empty or singleton" $ A.tail stream `shouldReturn` Nothing
```

Veladus: Could you that but I wanted something, which I could reuse for `init`

src/Streamly/Prelude.hs
```diff
@@ -332,11 +335,26 @@ head m =
         yield a _ = return (Just a)
     in (runStream (toStream m)) Nothing stop yield
 
+-- | Extract all but the first element of the stream, if any.
+tail :: (Streaming t, Monad m) => t m a -> m (Maybe (t m a))
+tail m =
+    let stop             = return Nothing
+        yield _ Nothing  = return Nothing
```
harendra-kumar: On second thoughts, I think the previous version of this was the correct one. We should differentiate between empty tail and no tail. That's the behavior for lists as well.

test/Main.hs
harendra-kumar: We can use a similar two line code for `init` as well. I think they are better understood when separate instead of using one monolithic function for both. You can just put these tests in the same block where we have other tests like `head` and `last`, it will keep the test code consistent for all tests.

test/Main.hs
```diff
@@ -686,14 +689,19 @@ streamOperations (stream, list, len) = do
         it "last empty" $ A.last stream `shouldReturn` Nothing
         it "maximum empty" $ A.maximum stream `shouldReturn` Nothing
         it "minimum empty" $ A.minimum stream `shouldReturn` Nothing
+        it "null empty" $ A.null stream `shouldReturn` True
+        it "tail empty" $ A.tail stream >>= 
+            \t -> if isNothing t then return () else expectationFailure "failed"
```
harendra-kumar: Why not write it as:

```
it "tail empty" $ A.tail stream `shouldReturn` Nothing
```

Veladus: because `Maybe (t m a)` doesn't have a `Show` instance, which is required by `shouldReturn`

harendra-kumar: Ah, ok. In that case you can write as:

```
it "tail empty" $ A.tail stream >>= maybe (return ()) (expectationFailure "failed")
```

harendra-kumar:
Ah, it seems you merged master in your branch whenever master changed. It has pulled other changes as well. Usually you should always rebase your branch using `git rebase master` so that you are always on top of master. Anyway, let me know if you want me to take care of this during merge.

Veladus:
Yeah, this would be awesome. I tried rewriting the history myself, but I've never done a rewrite of more than two or three commits. Thanks in advance :+1:

harendra-kumar:
Thanks @Veladus ! I merged your commits. I appreciate your time and effort, looking forward to more contributions.

Issue #9, opened 2018-03-10
last is broken


harendra-kumar:
The implementation of `last` is broken if the stream ends abruptly using the `stop` continuation. In that case even though there may have been a last element we would not return it. Consider this:

```
    go m1 =
        let stop            = return Nothing
            yield a Nothing = return (Just a)
            yield _ (Just x) = go x
        in (runStream m1) Nothing stop yield
```

As soon as stop is called we return `Nothing` which is incorrect. We should remember and thread around the last element via `go` and in case `stop` is called we should yield the last one.

harendra-kumar:
We need tests that include streams that stop abruptly using the `stop` continuation instead of using `yield a Nothing`.

harendra-kumar:
Fixed by #25, thanks @Abhiroop !

PR #25, opened 2018-03-13
Try a different implementation of last

Abhiroop:


src/Streamly/Prelude.hs
```diff
@@ -335,13 +335,10 @@ head m =
 
 -- | Extract the last element of the stream, if any.
 last :: (Streaming t, Monad m) => t m a -> m (Maybe a)
-last m = go (toStream m)
-    where
-    go m1 =
-        let stop            = return Nothing
```
Abhiroop: @harendra-kumar The problem in this implementation was that stop is always called after the last `yield a Nothing` as a result in the sequence of say `1 .: 2 .: 3`,  `Just 3` is yielded as the last element but finally `Nothing` is also returned. I was not sure how to stop `Nothing` from being called so wrote a new simple implementation. Literally wrote this in a couple of minutes, so please let me know if you have any better suggestions.

harendra-kumar: > The problem in this implementation was that stop is always called after the last yield a Nothing

`stop` cannot be called if `yield a Nothing` is called, because `last` terminates if `yield a Nothing` is called. The problem is that in some cases `stop` is called INSTEAD OF `yield a Nothing`.

harendra-kumar:
Please update the changelog as well.

Abhiroop:
@harendra-kumar The current implementation of `last` is much faster. Benchmarked multiple times and got this average:

```
benchmarked elimination/last/streamly
time                 17.58 ms   (17.03 ms .. 18.49 ms)
                          0.988 R²   (0.976 R² .. 0.997 R²)
mean                 18.47 ms   (18.11 ms .. 18.92 ms)
std dev              991.5 μs   (758.8 μs .. 1.235 ms)
```

harendra-kumar:
Everything looks good. Can you do a `git rebase -i` with `reword` for the first commit and provide an appropriate commit message and use `fixup` for the rest of the commits to squash them?

harendra-kumar:
Fixes #9

PR #26, opened 2018-03-15
Add forM, forM_ and sequence_ functions

psibi:


harendra-kumar:
Thanks @psibi ! The change looks good. Can you also add tests for these in `streamOperations` at the bottom of `test/Main.hs`? It should be pretty simple to add. Also, add a short note in changelog.

harendra-kumar:
@psibi I hope you have not already implemented a `replicateM`, I just implemented it as I needed to create some property tests.

psibi:
@harendra-kumar Nope, not yet. I'm just refreshing the internals of Streamly. So picked few basic things to warm up.

psibi:
@harendra-kumar What kind of test do you think would be apt for `_` operation like `mapM_` and `sequence_`?

harendra-kumar:
@psibi One way would be to use an IORef and each action in the stream can append an element to a list in the IORef. Finally you can check the IORef to see whether it has the full list that you expected. This is just like other tests except that the output is produced in the IORef (a side effect) instead of as a return value.

harendra-kumar:
@psibi this has been inactive for a while. Let me know if you are going to pick this up.

psibi:
@harendra-kumar Unfortunately, I don't think I would be able to give time to this in the coming months. So, closing the PR. Sorry for the bitrot.

PR #35, opened 2018-03-22
fix: Zip/AsyncZip applicative instances

### LINKED ISSUE #34: Equivalent code producing different output
I have found that two equivalent codes are producing different output.
* the first **flow** will produce `[(1, 1), (2, 2), (3, 3)]`
* while **the second** one will produce a single tuple `[(1, 1)]`

Maybe I am stupid but I can't find the answer to this behavior... ?

Example:
```haskell
import           Control.Monad    (join)
import           Streamly

main = mapM_ print $ join (S.toList . zipping $ flow)
  where
    flow = (,) <$> stream <*> stream
    -- flow = pure (,) <*> stream <*> stream
    stream = S.each [1, 2, 3]
```

hussein-aitlahcen:
This PR fixes #34 by correcting Zip and AsyncZip applicative instances.

### DIFF
--- a/src/Streamly/Streams.hs
+++ b/src/Streamly/Streams.hs
@@ -751,7 +751,7 @@ instance Monad m => Functor (ZipStream m) where
         in m Nothing stp yield
 
 instance Monad m => Applicative (ZipStream m) where
-    pure a = ZipStream $ scons a Nothing
+    pure = ZipStream . srepeat
     (<*>) = zipWith id
 
 instance Streaming ZipStream where
@@ -838,7 +838,7 @@ instance Monad m => Functor (ZipAsync m) where
         in m Nothing stp yield
 
 instance MonadAsync m => Applicative (ZipAsync m) where
-    pure a = ZipAsync $ scons a Nothing
+    pure = ZipAsync . srepeat
     (<*>) = zipAsyncWith id
 
 instance Streaming ZipAsync where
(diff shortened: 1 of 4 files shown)


src/Streamly/Streams.hs
```diff
@@ -751,7 +751,7 @@ instance Monad m => Functor (ZipStream m) where
         in m Nothing stp yield
 
 instance Monad m => Applicative (ZipStream m) where
-    pure a = ZipStream $ scons a Nothing
+    pure a = ZipStream $ let x = scons a (Just x) in x
```
harendra-kumar: it may be a good idea to factor out this pattern in a function called `srepeat` that will construct a stream by repeating a pure element just like `scons` constructs a singleton stream, and use `srepeat` at both these places. You can put that function next to `scons` definition.

harendra-kumar:
@hussein-aitlahcen thanks for finding and quickly fixing it as well! Can you also add a changelog entry in `Changelog.md`? Make a section called `Unreleased` on top and put it under that as a bug fix entry.

hussein-aitlahcen:
@harendra-kumar I just edited the code to make it pointfree since it's trivial. Will add the `Changelog.md` section.

hussein-aitlahcen:
@harendra-kumar made the final changes

hussein-aitlahcen:
Is this okay ?

harendra-kumar:
Merged, thanks again and looking to forward to more from you :-)

harendra-kumar:
Merged, thanks again and looking to forward to more from you :-)

PR #36, opened 2018-03-22
Iterate

Abhiroop:
Needed these functions for Mandelbrot etc

`iterateM` does not exist in `base`. There is an `iterateM` function in `streaming` and also an `iterateM_` in `monad-loops`

I was not sure whether to INLINE so did not add that pragma.

How to test `iterateM`?

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -126,16 +126,16 @@ each = Prelude.foldr cons nil
 -- | Iterate a pure function from a seed value, streaming the results forever
 iterate :: Streaming t => (a -> a) -> a -> t m a
 iterate step = fromStream . go
-  where
-  go s = scons s (Just (go (step s)))
+    where
+    go s = scons s (Just (go (step s)))
 
 -- | Iterate a monadic function from a seed value, streaming the results forever
-iterateM :: (Streaming t, Monad m) => (a -> m a) -> m a -> t m a
+iterateM :: (Streaming t, Monad m) => (a -> m a) -> a -> t m a
 iterateM step = fromStream . go
-  where
-  go ms = Stream $ \_ _ yld -> do
-    s <- ms
-    yld s (Just (go (step s)))
+    where
+    go s = Stream $ \_ _ yld -> do
+       a <- step s
+       yld s (Just (go a))
 
 -- | Read lines from an IO Handle into a stream of Strings.
 fromHandle :: (Streaming t, MonadIO m) => IO.Handle -> t m String
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -122,10 +122,12 @@ unfoldrM step = fromStream . go
 each :: (Streaming t, Foldable f) => f a -> t m a
 each = Prelude.foldr cons nil
 
+-- | Iterate a pure function from a seed value, streaming the results forever
 iterate :: Streaming t => (a -> a) -> a -> t m a
-iterate f = fromStream . go
+iterate step = fromStream . go
   where
-    go s = scons s (Just (go (f s)))
+    go s = scons s (Just (go (step s)))
+
 
 -- | Read lines from an IO Handle into a stream of Strings.
 fromHandle :: (Streaming t, MonadIO m) => IO.Handle -> t m String
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -24,6 +24,7 @@ module Streamly.Prelude
     , unfoldr
     , unfoldrM
     , each
+    , iterate
 
     -- * Elimination
     -- ** General Folds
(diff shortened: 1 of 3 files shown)


src/Streamly/Prelude.hs
```diff
@@ -120,6 +123,20 @@ unfoldrM step = fromStream . go
...
 each = Prelude.foldr cons nil
 
+-- | Iterate a pure function from a seed value, streaming the results forever
+iterate :: Streaming t => (a -> a) -> a -> t m a
+iterate step = fromStream . go
+  where
+  go s = scons s (Just (go (step s)))
+
+-- | Iterate a monadic function from a seed value, streaming the results forever
+iterateM :: (Streaming t, Monad m) => (a -> m a) -> m a -> t m a
```
harendra-kumar: The seed must be pure, and not monadic. Only the step should be monadic. See `unfoldrM` which is very similar to `iterateM`.

src/Streamly/Prelude.hs
```diff
@@ -120,6 +123,20 @@ unfoldrM step = fromStream . go
 each :: (Streaming t, Foldable f) => f a -> t m a
 each = Prelude.foldr cons nil
 
+-- | Iterate a pure function from a seed value, streaming the results forever
+iterate :: Streaming t => (a -> a) -> a -> t m a
+iterate step = fromStream . go
+  where
```
harendra-kumar: Use 4 space indentation everywhere. The `fromHandle` function below is an aberration. The rest of the file is using 4 spaces.

src/Streamly/Prelude.hs
```diff
@@ -120,6 +123,20 @@ unfoldrM step = fromStream . go
 each :: (Streaming t, Foldable f) => f a -> t m a
 each = Prelude.foldr cons nil
 
+-- | Iterate a pure function from a seed value, streaming the results forever
+iterate :: Streaming t => (a -> a) -> a -> t m a
```
harendra-kumar: BTW, in general, for any pure style functions we can use the list functions and convert that to a stream e.g. `each $ iterate f x` should be equivalent to `iterate f x`. I am not sure if the native iterate is significantly more efficient compared to that?

test/Main.hs
```diff
@@ -651,6 +651,12 @@ streamOperations (stream, list, len) = do
             return $ str == lst
         `shouldReturn` True
 
+    it "iterate" $ do
+            let lst = take len $ iterate (+ 1) 0
+            str <- A.toList . serially . (A.take len) $ (A.iterate (+ 1) (0 :: Int))
+            return $ str == lst
+        `shouldReturn` True
```
harendra-kumar: Since `lst` is not monadic we write that directly on the RHS of `shouldReturn`:

```
A.toList . serially . (A.take len) $ (A.iterate (+ 1) (0 :: Int)) `shouldReturn`  take len $ iterate (+ 1) 0
```

src/Streamly/Prelude.hs
harendra-kumar: I see that `streaming` is using a monadic seed. But I am not sure if there is a clear advantage in doing that. We can always run the action separately and pass the seed to this function. So a pure seed should be enough to achieve all that a monadic seed can. Pure seed is simpler. I would go with pure unless there is some clear benefit proven that impure seed gives us.

harendra-kumar:
You can test `iterateM` by using `Identity` monad, in that case `iterateM` should be the same as `iterate` and you can use a similar test.

src/Streamly/Prelude.hs
Abhiroop: In the other monadic fold functions in the prelude `streamly` uses a monadic seed. Is there any particular advantage to that in folds?

harendra-kumar: The general rule is: inputs are pure and outputs are monadic, remember the `a -> m b` in bind. That's what has been chosen as convention for convenient composition. In `unfoldr` and `iterate` it is a `seed` i.e. an input, in folds it is an `accumulator` i.e. an output, that's why it is monadic. We cannot make it pure without making the function itself pure.

test/Main.hs
```diff
@@ -651,6 +651,16 @@ streamOperations (stream, list, len) = do
...
         `shouldReturn` True
 
+    it "iterate" $
+            (A.toList . serially . (A.take len) $ (A.iterate (+ 1) (0 :: Int)))
+            `shouldReturn` (take len $ iterate (+ 1) 0)
+
+    it "iterateM" $ do
+              let addM = (\ y -> return (y + 1))
+              A.toList . serially . (A.take len) $ A.iterateM addM (0 :: Int)
+           `shouldReturn` (take len $ iterate (+ 1) 0)
```
harendra-kumar: Can be indented a bit more.

harendra-kumar:
LGTM.

PR #37, opened 2018-03-23
(WIP) docs: typos

hussein-aitlahcen:
This PR improve artificially the documentation !

### DIFF
--- a/src/Streamly/Core.hs
+++ b/src/Streamly/Core.hs
@@ -150,7 +150,7 @@ data SVar m a =
 --
 -- data Stream m a = Stop | Yield a (Maybe (Stream m a))
 --
--- Our goal is to be able to represent finite as well infinite streams and
+-- Our goal is to be able to represent finite as well as infinite streams and
 -- being able to compose a large number of small streams efficiently. In
 -- addition we want to compose streams in parallel, to facilitate that we
 -- maintain a local state in an SVar that is shared across and is used for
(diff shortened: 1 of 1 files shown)


src/Streamly/Core.hs
```diff
@@ -150,10 +150,10 @@ data SVar m a =
 --
 -- data Stream m a = Stop | Yield a (Maybe (Stream m a))
 --
--- Our goal is to be able to represent finite as well infinite streams and
+-- Our goal is to be able to represent finite as well as infinite streams and
 -- being able to compose a large number of small streams efficiently. In
 -- addition we want to compose streams in parallel, to facilitate that we
--- maintain a local state in an SVar that is shared across and is used for
+-- maintain a local state in a SVar that is shared across and is used for
```
harendra-kumar: I pronounce this as "ess-var", so based on the sound "an" sounds appropriate :-)

hussein-aitlahcen: I just saw that in many place, was going to cancel that change :P

harendra-kumar: BTW, this documentation is not user visible. Though it should be correct nevertheless, but we do not have to as stringent as we have to be for user visible docs.

hussein-aitlahcen: Contributors are users :P, we learn your architecture by reading the comments aswell ^^

PR #40, opened 2018-03-24
Foldr m

Abhiroop:


harendra-kumar:
@Abhiroop can you rebase the change, the changelog file has changed.

harendra-kumar:
I just realized, we should not have changed the foldrM signature. The old signature was the right signature, because we want foldr to be lazy. With the new changes the laziness/strictness of foldrM depends on the bind of the underlying monad. For example, when using IO monad foldrM becomes unnecessarily strict. We should revert back to the old signature.

harendra-kumar:
BTW, vector also has the incorrect signature which is not lazy enough.

PR #41, opened 2018-03-25
rearrange some functions in prelude

harendra-kumar:
According to the correct categories they belong to.

Let me know if it badly affects someone's pending changes.

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -23,18 +23,20 @@ module Streamly.Prelude
     , (.:)
     , unfoldr
     , unfoldrM
-    , each
     , iterate
     , iterateM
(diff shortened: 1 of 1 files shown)


Veladus:
I currently have some pending work done in Prelude. This PR wouldn't effect my work too badly, but I suppose a similar PR would be needed afterwards.

harendra-kumar:
Let me know if it is better for you to go first and then do it at once, I will hold off in that case.

Veladus:
I think doing the cleanup in one go would be better 😃

harendra-kumar:
Yeah, that will be more efficient. Thanks for letting me know!

harendra-kumar:
This is obsolete now, dropping it.

Issue #39, opened 2018-03-24
RFC: Nomenclature for fold functions


harendra-kumar:
Currently we have `foldl` and `foldlM` functions whose signatures are more general than the standard ones and inline with the `foldl` library. It means that they are slightly different from the standard fold functions. To keep full compatibility with the nomenclature of standard lists and Foldable we should perhaps use different names for the functions created for the purposes of `foldl`. 

The names `foldl`, `foldl'`, `foldl1` are all taken. The name `fold` is used by `Foldable` typeclass. If we are not going to have a `Foldable` instance we can use `fold`, `foldM` for these general functions. Note that we also have `scan` that is a general version of `scanl`.

harendra-kumar:
Instead making it a breaking change we can also deprecate the old function names.

harendra-kumar:
We should perhaps use `foldl'` for strict left fold to be inline with `Foldable`. Do we need a lazy left fold?

harendra-kumar:
It might make sense to provide a `fold` and a `foldMap` function in line with the `Foldable` typeclass. So those names should probably not be reused. 

How about `foldx` and `foldxM` where the `x` can be thought of as a mnemonic for e`x`traction function required in the `foldl` library compatible fold functions?

PR #43, opened 2018-03-27
Prelude functions

Veladus:
#11 
Adds many but not quite all of the requested functions. In the next days I consider to add a few more but in the meantime I'd appreciate your opinion.

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -206,19 +206,14 @@ foldr step acc m = go (toStream m)
             yield a (Just x) = go x >>= \b -> return (step a b)
         in (runStream m1) Nothing stop yield
 
--- | Right fold, with no starting value. Returns Nothing if the list is empty.
+-- | Right fold, for non-empty streams, using first element as the starting
+-- value. Returns 'Nothing' if the stream is empty.
 foldr1 :: (Streaming t, Monad m) => (a -> a -> a) -> t m a -> m (Maybe a)
-foldr1 step m = go (toStream m)
+foldr1 step = foldr go Nothing
     where
-    go m1 =
-        let stop = return Nothing
-            yield a Nothing = return $ Just a
-            yield a (Just x) = go x >>= \b ->
-                return $ Just $ maybe_ a (step a) b
-        in runStream m1 Nothing stop yield
-
-    maybe_ def _ Nothing = def
-    maybe_ _ f (Just v)  = f v
+        go x m = Just (case m of
+                            Nothing -> x
+                            Just y -> step x y)
 
 -- | Right fold with a monadic step function.  See 'toList' for an example use.
 {-# INLINE foldrM #-}
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -33,9 +33,11 @@ module Streamly.Prelude
     -- * Elimination
     -- ** General Folds
     , foldr
+    , foldr1
     , foldrM
     , scan
     , foldl
+    , foldl1
     , foldlM
     , uncons
 
@@ -59,6 +61,8 @@ module Streamly.Prelude
     , lookup
     , find
     , partition
+    , and
+    , or
 
     -- * Splitting
     , splitAt
@@ -116,7 +120,8 @@ import           Prelude hiding              (filter, drop, dropWhile, take,
                                               tail, length, null, reverse,
                                               iterate, repeat, replicate,
                                               cycle, lookup, splitAt, span,
-                                              break, init)
+                                              break, init, foldr1, foldl1, and,
+                                              or)
 import qualified Prelude
 import qualified System.IO as IO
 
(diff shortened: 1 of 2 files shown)


harendra-kumar:
@Veladus this PR looks awesome, I specially like the fact that you added test cases for everything. I will take a look at the code by tomorrow.

harendra-kumar:
A small nitpick, no change required in this PR but something to keep in mind for later, we have a convention to start the commit messages in imperative mood e.g. use `add Prelude.intersperse` instead of `added Prelude.intersperse`.

src/Streamly/Prelude.hs
```diff
@@ -138,6 +142,31 @@ iterateM step = fromStream . go
...
+replicate n a = take n $ repeat a
+
+-- | @cycle m@ creates an infinite stream by starting at the first element of
+-- @m@ after exhausting all other elements.
+-- The given stream may NOT be empty.
+cycle :: (Streaming t) => t m a -> t m a
+cycle m = fromStream $ goIfNotNull $ toStream m
+    where
+    goIfNotNull m1 = Stream $ \svr _ yld ->
+        let stop = error "Can't cycle an empty Stream"
```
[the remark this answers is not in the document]
harendra-kumar: Partial functions are to be avoided. But I did not comment  on this because I am not in favor of having the `cycle` function in the first place as there is a `cycle1` in `Data.Semigroup` that we can use for the same purpose.

src/Streamly/Prelude.hs
```diff
@@ -440,6 +612,155 @@ maximum m = go Nothing (toStream m)
...
+        mbUncons1 <- uncons m1
+        mbUncons2 <- uncons m2
+        case (mbUncons1, mbUncons2) of
+            (Nothing, Nothing) -> return True
+            (Just (h1, m1'), Just (h2, m2')) | h1 == h2 -> equiv m1' m2'
+            _ -> return False
+
+-- | Takes two streams and determines if the first stream is contained
+-- continously in the second.
+isInfixOf :: (Streaming t, Monad m, Eq a) => t m a -> t m a -> m Bool
```
harendra-kumar: This is also going through the streams more than once.

src/Streamly/Prelude.hs
```diff
@@ -440,6 +612,155 @@ maximum m = go Nothing (toStream m)
...
+                    current <- if headM == headP then p' `isInfixOf` m'
+                                                 else return False
+                    return $ current || next
+                Nothing -> return False
+
+-- | Takes two streams and determines if the elements of the first stream
+-- occur in the second in the right order, but opposed to @isInfixOf@ they
+-- don't have to be continous.
+isSubsequenceOf :: (Streaming t, Monad m, Eq a) => t m a -> t m a -> m Bool
+isSubsequenceOf p m = do
```
harendra-kumar: This is also evaluating elements of the stream more than once.

src/Streamly/Prelude.hs
```diff
@@ -440,6 +612,155 @@ maximum m = go Nothing (toStream m)
...
+        let yield a Nothing =
+                if p a then yld offset Nothing
+                       else stp
+            yield a (Just x) =
+                if p a then yld offset $ Just $ go (offset + 1) x
+                       else runStream (go (offset + 1) x) svr stp yld
+        in runStream m1 Nothing stp yield
+
+-- | Gives the index of the first stream element satisfying the given
+-- preficate.
```
harendra-kumar: typo.

src/Streamly/Prelude.hs
```diff
@@ -440,6 +612,155 @@ maximum m = go Nothing (toStream m)
...
+            case mbUnconsM of
+                Just (headM, m') | headM == headP -> p' `isPrefixOf` m'
+                _ -> return False
+
+-- | Takes two streams and returns true if and only if the second  the second
+-- stream ends in the first.
+-- Both streams must be finite.
+isSuffixOf :: (Streaming t, Monad m, Eq a) => t m a -> t m a -> m Bool
+isSuffixOf p m = do
+    lengthP <- length p
```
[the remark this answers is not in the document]
harendra-kumar: For all the tests you can add a mechanism to automatically verify that we are not evaluating any element of the stream more than once. One way to do that is like in the above example, replace the putStrLn with an append to a list in an IORef. At the end you can check that the IORef does not have duplicate elements.

harendra-kumar:
@Veladus it may be a good idea to move the functions that do not need any more work to a separate PR so that they can be merged.

Veladus:
Ok, I could make that, but first we have to decide which ones are finished. I think the following list of functions should be fine.

- intersperseM
- repeatM
- find
- lookup
- partition
- init
- findIndices
- findIndex
- elemIndices
- elemIndex
- and
- or

I don't know about `foldl1` and `foldr1` concerning #39.

harendra-kumar:
This list is good except `partition`. The names,`foldr1` and `foldl1'` are good as they are consistent with the standard versions and also with the vector library. So you can include these two as well.

Veladus:
Ok, I will create a new PR of these on the weekend.

Issue #21, opened 2018-03-13
Add a dual of StreamT


harendra-kumar:
The `StreamT` type composes first element of the outer loop with all the elements in the inner loop and then goes ahead with the second element of the outer loop. There is a dual possible where we can compose the first element of the inner loop with all the elements of the outer loop and then do the same with the second element of inner loop.

```
x <- [1,2]
y <- [3,4]
return (x,y)
```
StreamT produces `(1,3) (1,4) (2,3) (2,4)` whereas the dual will produce `(1,3) (2,3) (1,4) (2,4)`.

This has the effect of inverting the order of the loops without changing the code. In many cases the efficiency of the program may matter depending on the order of the loops. The dual of StreamT can help in inverting the order by just changing the type just like all other streamly stream types work.

What do we call the type? CoStreamT, InverseT, InverseStreamT? The combinator for this could be `inversely`?

Another possibility could be to have a combinator that just flips the current order?

hussein-aitlahcen:
Looking at #22, `CoSerialT` would be great !

harendra-kumar:
Sounds good.

harendra-kumar:
This is not much useful as discussed in #44. Closing it.

PR #44, opened 2018-03-28
introduce ReverseT

hussein-aitlahcen:
This PR implements #21 !

### DIFF
--- a/benchmark/Main.hs
+++ b/benchmark/Main.hs
@@ -33,6 +33,7 @@ main = do
     -- XXX due to a GHC bug passing bind as an argument causes perf
     -- degradation, so we should keep that in account when comparing.
     let as = streamly_serial
+        ac = streamly_coserial
         ai = streamly_interleaved
         aa = streamly_async
         ap = streamly_parallel
@@ -45,6 +46,13 @@ main = do
                 , bench "fair serial"   $ nfIO (as (A.<=>))
                 , bench "left parallel" $ nfIO (as (A.<|))
                 , bench "fair parallel" $ nfIO (as (A.<|>))
+                ]
+
+            , bgroup "coserial bind"
+                [ bench "serial"        $ nfIO (ac (A.<>))
+                , bench "fair serial"   $ nfIO (ac (A.<=>))
+                , bench "left parallel" $ nfIO (ac (A.<|))
+                , bench "fair parallel" $ nfIO (ac (A.<|>))
                 ]
 
             , bgroup "interleaved bind"
@@ -138,6 +146,12 @@ streamly_serial
     :: (A.StreamT IO Int -> A.StreamT IO Int -> A.StreamT IO Int)
     -> IO Int
 streamly_serial = streamly_basic (A.toList . A.serially)
+
+{-# INLINE streamly_coserial #-}
+streamly_coserial
+    :: (A.ReverseT IO Int -> A.ReverseT IO Int -> A.ReverseT IO Int)
+    -> IO Int
+streamly_coserial = streamly_basic (A.toList . A.reversely)
 
 {-# INLINE streamly_interleaved #-}
 streamly_interleaved
--- a/src/Streamly.hs
+++ b/src/Streamly.hs
@@ -21,6 +21,7 @@ module Streamly
     -- * Product Style Composition
     -- $product
     , StreamT
+    , ReverseT
     , InterleavedT
     , AsyncT
     , ParallelT
@@ -41,6 +42,7 @@ module Streamly
     -- * Stream Type Adapters
     -- $adapters
     , serially
+    , reversely
     , interleaving
     , asyncly
     , parallely
--- a/src/Streamly/Core.hs
+++ b/src/Streamly/Core.hs
@@ -31,6 +31,7 @@ module Streamly.Core
 
     -- * Composition
     , interleave
+    , roundrobin
 
     -- * Concurrent Stream Vars (SVars)
     , SVar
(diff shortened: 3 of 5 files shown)


harendra-kumar:
LGTM. Just a couple of things:

1) let's call it `CoserialT` and `coserially` as you proposed earlier. The reason is that Coserial at least tells you that it is serial and it is some sort of a variant of serial. `Reverse` does not tell you `reverse` of what, and it does not tell you anything about the serial nature.

2) I find the name `roundrobin` rather confusing, but since it is internal I would not fret about it, you can keep it that way if you prefer. My source of confusion is that if you call it round robin then serial is also round robin because they are not different at all, its just a matter of ordering. For example, If you just swap the two lines in your tests you will get serial.

harendra-kumar:
Also, did you run the benchmark, to find out the timing in comparison to serial bind?

harendra-kumar:
One of the reasons why it has limited use, is that reversing the order of the loops can change the logic in most cases in such a way that it may not make sense. My original intention to have this type was for performance reasons where inverting the order of the loops can change perf characteristics without changing the logic. For example, consider this modification to your test case:

```
        it "produce layers reversely" $
          toListReversely (do
            x <- A.each [1, 2]
            y <- A.each [3, 4]
            if (x + y > 5) then mzero else return ()
            z <- A.each [5, 6]
            return (x, y, z))
```

Now, if we run the same code using `serial` then the results will be entirely different, it won't just be ordering change. So it won't be useful. For the use case when we want to just change the order of the loops it may be ok for the programmer to just change it manually. So I  now have a bigger question mark on its utility. Perhaps we should drop the idea altogether?

harendra-kumar:
Also, one more way of doing things will just confuse the programmer about which one to use, serial or coserial?

hussein-aitlahcen:
> let's call it CoserialT and coserially as you proposed earlier. The reason is that Coserial at least tells you that it is serial and it is some sort of a variant of serial. Reverse does not tell you reverse of what, and it does not tell you anything about the serial nature.

I am okay with that, especially because it is closely related to the `Serial` stream.

> I find the name roundrobin rather confusing, but since it is internal I would not fret about it, you can keep it that way if you prefer. My source of confusion is that if you call it round robin then serial is also round robin because they are not different at all, its just a matter of ordering. For example, If you just swap the two lines in your tests you will get serial.

Given the joining nature of the function, why not calling it `cojoin` ? I have found that this function is the only difference betwenn `Serial`/`Coserial` in the Monad instance. And actually, the `Serial` counterpart is `join` (which is not yet implemented in the core of Streamly).

Here is an example:
```haskell
sjoin :: Stream m (Stream m a) -> Stream m a
sjoin m = Stream $ \_ stp yld ->
  let run x = (runStream x) Nothing stp yld
      yield m1 Nothing = run m1
      yield m1 (Just rr) = run $ Stream $ \_ stp1 yld1 ->
        let stop = (runStream $ sjoin rr) Nothing stp1 yld1
            yield1 a Nothing  = yld1 a (Just $ sjoin rr)
            yield1 a (Just r) = yld1 a (Just $ sjoin $ scons r (Just rr))
        in (runStream m1) Nothing stop yield1
  in (runStream m) Nothing stp yield

scojoin :: Stream m (Stream m a)  -> Stream m a
scojoin m = Stream $ \_ stp yld ->
  let run x = (runStream x) Nothing stp yld
      yield m1 Nothing = run m1
      yield m1 (Just rr) = run $ Stream $ \_ stp1 yld1 ->
        let stop = (runStream $ scojoin rr) Nothing stp1 yld1
            yield1 a Nothing  = yld1 a (Just $ scojoin rr)
            yield1 a (Just r) = yld1 a (Just $ scojoin (concat rr $ singleton r))
        in (runStream m1) Nothing stop yield1
  in (runStream m) Nothing stp yield
```

Finally, if you think that this will just make the library harder to understand and use, we could make an experimental module, or just drop it :). That was fun to code tho :D

hussein-aitlahcen:
@harendra-kumar I've just checked the benchmark, we lose something like `30%` of performance in a `CoSerial` execution

harendra-kumar:
@hussein-aitlahcen I am in favor of dropping it at this point. Don't feel disappointed, I drop a lot more code than this quite often, and quite often we learn from it and it helps in doing something better later. I guess you learnt quite a bit about how streamly works with this. You can pick up some other issues.

I tried benchmarking your `sjoin` implementation for `StreamT` but unfortunately it is quite slower than the current implementation:

```haskell
sjoin :: Stream m (Stream m a) -> Stream m a
sjoin m = Stream $ \_ stp yld ->
  let run x = (runStream x) Nothing stp yld
      yield m1 Nothing = run m1
      yield m1 (Just rr) = run $ Stream $ \_ stp1 yld1 ->
        let stop = (runStream $ sjoin rr) Nothing stp1 yld1
            yield1 a Nothing  = yld1 a (Just $ sjoin rr)
            yield1 a (Just r) = yld1 a (Just $ sjoin $ scons r (Just rr))
        in (runStream m1) Nothing stop yield1
  in (runStream m) Nothing stp yield

instance Monad m => Monad (StreamT m) where
    return = pure
    s >>= f = StreamT . sjoin . getStreamT $ fmap (getStreamT . f) s
```

Current implementation:
```
benchmarked streamly/serial bind/serial
mean                 11.11 ms   (10.97 ms .. 11.25 ms)
```

With sjoin:

```
benchmarked streamly/serial bind/serial
mean                 19.10 ms   (18.80 ms .. 19.55 ms)
```

hussein-aitlahcen:
I don't mind dropping it :). Was a good experimentation ! Thanks for your advices, I'll work on differents issues.

harendra-kumar:
Dropping this for reasons discussed above.

PR #54, opened 2018-04-14
Add quickcheck based property tests

### LINKED ISSUE #30: Add property tests for streams
Currently we have unit tests where we make different cases of streams and compositions manually and test them. It may be a good idea to do this automatically using QuickCheck Arbitrary instances instead. It will be less manual labor and more testing.

harendra-kumar:
fixes #30


harendra-kumar:
Merged vi #55

PR #65, opened 2018-04-21
Include singleton as a continuation in stream type

harendra-kumar:


harendra-kumar:
This simplifies the type and the code and also gives some performance bonus.

Issue #74, opened 2018-05-23
Please add `fmapMaybe`


Fresheyeball:
Since streams can be `filter`ed they should have an instance of `Control.Compactable` or at the very least `fmapMaybe :: (IsStream t, Monad m) => (a -> Maybe b) -> t m a -> t m b`
This would eliminate the following anti-pattern appearing in code:
```haskell
foo = fmap fromJust . S.filter isJust
```

harendra-kumar:
Thanks for the suggestion @Fresheyeball . 

I am trying to keep the API as close to lists and the standard prelude as possible. So I am in favor of a `mapMaybe` and the monadic version `mapMaybeM`:

```
mapMaybe :: (IsStream t, Monad m) => (a -> Maybe b) -> t m a -> t m b
mapMaybeM :: (IsStream t, Monad m) => (a -> m (Maybe b)) -> t m a -> t m b
```

Though, in general the monadic version covers the pure case too. I am not sure if we should include the pure version as well.

subbyte:
Hi @harendra-kumar , I am actually using `fmap fromJust . S.filter isJust` in my code. Could you give a little bit more information where can I find the `mapMaybe` function? I will be happy to use it.

harendra-kumar:
@subbyte `mapMaybe` is not yet implemented. It is what I proposed to add above. I will add it in the next release, in case you feel like implementing it please go ahead and send a PR.

harendra-kumar:
This is done, thanks @subbyte !

PR #76, opened 2018-05-26
mapMaybe and mapMaybeM Implemented

subbyte:
For addressing issue #74 

I am new to Haskell. Feel free to guide me towards a better implementation if possible. I have tested my implementation with `test/mapMB.hs` and the logic should be correct.

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -100,12 +100,14 @@ module Streamly.Prelude
     , takeWhile
     , drop
     , dropWhile
+    , mapMaybe
 
     -- * Reordering
     , reverse
 
     -- * Mapping
     , mapM
+    , mapMaybeM
     , sequence
 
     -- * Zipping
@@ -129,6 +131,7 @@ where
 import           Control.Monad (void)
 import           Control.Monad.IO.Class      (MonadIO(..))
 import           Data.Semigroup              (Semigroup(..))
+import           Data.Maybe                  (isJust, fromJust)
 import           Prelude hiding              (filter, drop, dropWhile, take,
                                               takeWhile, zipWith, foldr, foldl,
                                               mapM, mapM_, sequence, all, any,
@@ -570,6 +573,24 @@ dropWhile p m = fromStream $ go (toStream m)
                       | otherwise = yld a r
          in (S.runStream m1) ctx stp single yield
 
+-- | Map + filter: compute a stream of b from a stream of a and filter out
+-- elements that are mapped to Nothing. It is similar to 'mapMaybe' in
+-- 'Data.Maybe' but applies to streams instead of list.
+--
+-- @since 0.2.1
+{-# INLINE mapMaybe #-}
+mapMaybe :: (IsStream t) => (a -> Maybe b) -> t m a -> t m b
+mapMaybe f m = go (toStream m)
+  where
+    go m1 = fromStream $ Stream $ \_ stp sng yld ->
+        let single a = case f a of
+                Just b  -> sng b
+                Nothing -> stp
+            yield a r = case f a of
+                Just b  -> yld b (toStream $ go r)
+                Nothing -> (S.runStream r) Nothing stp single yield
+        in (S.runStream m1) Nothing stp single yield
+
 -- | Determine whether all elements of a stream satisfy a predicate.
 --
 -- @since 0.1.0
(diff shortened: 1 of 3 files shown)


harendra-kumar:
@subbyte thanks for the PR. The implementation looks good to me. 

We can write the tests differently though. Instead of using a separate test suite, we can add it to `test/Prop.hs` in the `tranformOps` function something like this:

```
    let f = if odd x then Just (x + 1) else Nothing
     in prop (desc ++ " mapMaybe") $ transformFromList constr eq (mapMaybe f) $ t . (A.mapMaybe f)
```

It is a property test and will therefore perform much more comprehensive tests. It compares the results of our mapMaybe with the regular mapMaybe. The first mapMaybe in the above code would be `Data.Maybe.mapMaybe`.

Similarly, `mapMaybeM` test can be written, but you cannot use `transformFromList` for that as there is no reference `mapMaybeM` to compare. Instead we can write another function like `transformFromList` in which we can compare the output of `mapMaybeM` with another implementation using `mapM` and `filter`.

In case you find it difficult to go through and understanding the property test stuff, let me know and I can add the tests.

harendra-kumar:
Also, `mapMaybeM` can be made concurrency capable. For that we will have to use `|:` to reconstruct the stream and also use the first argument (`svr`) of `Stream` instead of ignoring it. See the implementation of `mapM` for example. Though this can be done as another change/PR, your current change is good to go in its current form as well.

harendra-kumar:
When you are on it, you can also place these functions next to `mapM` so that similar functions are placed together.

subbyte:
1. `mapMaybeM` revised to reconstruct the stream
2. `mapMaybe` and `mapMaybeM` repositioned in `Prelude.hs`
3. `mapMaybe` test case merged into `test/Prop.hs`
4. `mapMaybeM` test case not implemented using `Hspec` (haven't figured out how to do it)

src/Streamly/Prelude.hs
```diff
@@ -750,6 +770,23 @@ mapM f m = go (toStream m)
...
+mapMaybeM :: (IsStream t, MonadAsync m) => (a -> m (Maybe b)) -> t m a -> t m b
+mapMaybeM f m = go (toStream m)
+  where
+    go m1 = fromStream $ Stream $ \svr stp sng yld ->
+        let single a = f a >>= \mbb -> case mbb of
+                Just b  -> sng b
+                Nothing -> stp
+            reconstructStream x = S.runStream (toStream x) svr stp sng yld
+            yield a r = f a >>= \mbb -> case mbb of
+                Just b  -> reconstructStream $ return b |: go r
```
harendra-kumar: There is a little problem with this, it has limited parallelism. The problem is that we have to first run the action i.e. `f a >>= \mbb -> case mbb of` before we can compose its result in parallel with `go r`. This means we can never run more than one action in parallel.

The root of the problem is because we are doing map and filter in one step. It will have full parallelism if we implement it as `map fromJust . filter isJust . mapM f`. In that case we can run as many actions in parallel as we want (see `mapM`, it runs the actions in parallel) and then filter in one go.

harendra-kumar: 1) When you use `f a` to make a decision it means you need to evaluate the monadic action`f a` to get its result, so the action has already been executed there.  Later, in the `Just` case, we just add the pure value `b`, which is the result of `f a`, to the stream. The `runStream` function now passes this stream constructed with the result of `f a` and the remaining stream `go r` to the continuations.

2) `fmap` is deliberately not concurrent that's why it discards `svr`. fmap is a pure function and therefore not relatively expensive. Only monadic functions are concurrent because monadic actions may be expensive. However, if we want to use a concurrent map for even a pure function we can always use `mapM (return . f)` instead of `fmap f`. In mapMaybeM, the expensive monadic operation is done via concurrent `mapM` but the cheaper `fromJust` is done non-concurrently using `fmap`.

subbyte: Thanks for your patience! Now I understand why my previous versions do not support concurrent execution. I pushed the latest change below using `mapM (return . fromJust)` as you suggested. Now the chain only have `S.mapM` and `S.filter`, which should support concurrent execution.

harendra-kumar:
It is a good idea to squash the changes to keep the commit history clean. I suggest a `git rebase -i 83ffe5c^` and use `fixup` for all the commits except the first one. You can optionally use `edit` on the first commit to change the commit message to something like `add mapMaybe and mapMaybeM`.

harendra-kumar:
Thanks! Looks good. Please add a changelog entry as well.

harendra-kumar:
Thank you @subbyte  !

PR #95, opened 2018-07-23
Approved prelude functions

Veladus:
Finally I got the time and patience to recommit my changes made for #43.

### DIFF
--- a/src/Streamly/Streams/StreamK.hs
+++ b/src/Streamly/Streams/StreamK.hs
@@ -117,6 +117,9 @@ module Streamly.Streams.StreamK
     , mapM
     , sequence
 
+    -- ** Inserting
+    , intersperseM
+
     -- ** Map and Filter
     , mapMaybe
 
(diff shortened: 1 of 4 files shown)


harendra-kumar:
Thanks @Veladus ! I will take a look and hopefully merge it soon.

src/Streamly/Streams/StreamK.hs
```diff
@@ -798,6 +859,22 @@ sequence m = go (toStream m)
...
+{-# INLINE intersperseM #-}
+intersperseM :: (IsStream t, Monad m) => m a -> t m a -> t m a
+intersperseM a m = fromStream $ prependingStart (toStream m)
+    where
+    prependingStart m1 = Stream $ \st stp sng yld ->
+        let yieldk i x = yld i $ go x
+         in unStream m1 st stp sng yieldk
+    go m2 = Stream $ \st stp _ yld ->
+        let single i = a >>= \y -> yld y $ yield i
+            yieldk i x = a >>= \y -> yld y $ cons i $ go x
```
harendra-kumar: `intersperseM` is a monadic function therefore should have concurrency capability. For that you need to make `single` and `yieldk` concurrent, see mapM implementation as an example. Construct a concurrent stream using `|:` and use `unStream` to run it e.g. `unStream (a |:  return i |: go x)`...

harendra-kumar:
Let me check.

src/Streamly/Streams/StreamK.hs
```diff
@@ -798,6 +868,22 @@ sequence m = go (toStream m)
...
+-- Inserting
+-------------------------------------------------------------------------------
+
+{-# INLINE intersperseM #-}
+intersperseM :: (IsStream t, MonadAsync m) => m a -> t m a -> t m a
+intersperseM a m = fromStream $ prependingStart (toStream m)
+    where
+    prependingStart m1 = Stream $ \st stp sng yld ->
+        let yieldk i x = unStream (return i |: go x) st stp sng yld
+         in unStream m1 st stp sng yieldk
```
harendra-kumar: We need to use `rstState st` when running m1 here. The `SVar` in `st` is to be shared only at this level (i.e. in the continuations we composed) and not with `m1`.

src/Streamly/Streams/StreamK.hs
```diff
@@ -798,6 +868,22 @@ sequence m = go (toStream m)
...
+    go m2 = fromStream $ Stream $ \st stp sng yld ->
+        let single i = unStream (a |: yield i) st stp sng yld
+            yieldk i x = unStream (a |: return i |: go x) st stp sng yld
+         in unStream m2 st stp single yieldk
```
harendra-kumar: Same as on line 881.

harendra-kumar: We can add a `intersperseM` test in `transformCombineOpsCommon` to detect the illegal sharing of SVar. I think it will fail if we do not reset the state here.

Veladus:
I now have done the last changes and I did a rebase to clean up the history. How should I publish the new rewritten history? Should I just do a `git push --force`

harendra-kumar:
Yes that’s right.

harendra-kumar:
Thanks for the contribution @Veladus ! Much appreciated.

PR #104, opened 2018-09-21
Linted

wygulmage:
Not sure if you want this, but I incrementally HLinted and benchmarked everything.
There should be no performance degradation and possibly a slight increase in clarity.

I only factored out 1 function, where there were 4 exact duplications. It's set to inline so again there should be no difference in performance.

### DIFF
--- a/src/Streamly/Prelude.hs
+++ b/src/Streamly/Prelude.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE BangPatterns              #-}
+-- {-# LANGUAGE BangPatterns              #-}
 {-# LANGUAGE CPP                       #-}
 {-# LANGUAGE FlexibleContexts          #-}
 {-# LANGUAGE FlexibleInstances         #-}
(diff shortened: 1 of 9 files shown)


wygulmage:
Done. The remaining warnings are matters of opinion (like, should contributors have to know `fromMaybe` when a `case` works?).

You'll probably want to ignore duplicate code warnings for now. When it's only repeated twice, it's not clear what should be factored out.

harendra-kumar:
@wygulmage I am still seeing some commented out code in the diff. Did you forget to push the commit removing it? Let me know if you want me to remove it post merge.

wygulmage:
closing this while looking into how it interacts with testing, etc.

harendra-kumar:
did you find something wrong with it?

wygulmage:
Ah, sorry about that. Everything was working fine, but I noticed that you have a style guide and I wasn't following it. I also need to add a .hlint.yaml. Should I update this pull request or open a new one?

harendra-kumar:
Don't worry about the style guide, its only a guide. As long as the net effect of the PR is to make things better I will accept it. I would have merged this by now had I not been busy with something else. You can reopen this one or create another, either way is fine.

PR #105, opened 2018-09-29
Replace monad-control with unliftio-core

mitchellwrosen:
Hello there, this patch swaps out `monad-control` for `unliftio-core`, in case you want it. Have you heard of `unliftio-core` before? If not, I would be glad to try and dig up some blog posts that I've read that justify why the change is beneficial.

### DIFF
--- a/examples/AcidRain.hs
+++ b/examples/AcidRain.hs
@@ -1,15 +1,39 @@
+{-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 
 -- This example is adapted from Gabriel Gonzalez's pipes-concurrency package.
 -- https://hackage.haskell.org/package/pipes-concurrency-2.0.8/docs/Pipes-Concurrent-Tutorial.html
 
 import Streamly
 import Streamly.Prelude as S
+import Control.Concurrent.STM (TVar, atomically, newTVarIO, readTVar, writeTVar)
 import Control.Monad (when)
+import Control.Monad.Catch (MonadThrow)
 import Control.Monad.IO.Class (MonadIO(liftIO))
-import Control.Monad.State (MonadState, get, modify, runStateT)
+import Control.Monad.IO.Unlift -- (MonadUnliftIO(askUnliftIO))
+import Control.Monad.Reader (ReaderT(..))
+import Control.Monad.State (MonadState, get, modify, runStateT, state)
+import Data.Coerce (coerce)
 
 data Event = Harm Int | Heal Int | Quit deriving (Show)
+
+newtype GameMonad a = GameMonad { unGameMonad :: ReaderT (TVar Int) IO a }
+  deriving (Functor, Applicative, Monad, MonadIO, MonadThrow)
+
+instance MonadState Int GameMonad where
+  state f = GameMonad $ ReaderT $ \healthVar -> do
+    atomically $ do
+      health <- readTVar healthVar
+      let (x, health') = f health
+      writeTVar healthVar health'
+      return x
+
+instance MonadUnliftIO GameMonad where
+  askUnliftIO = GameMonad $ do
+    UnliftIO f <- askUnliftIO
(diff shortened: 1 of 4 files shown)


harendra-kumar:
Hey @mitchellwrosen, thanks for the patch. Yes, I know about `unliftio`, I considered it before. But I am not sure if I want to preclude all stateful monads from use with streamly. I have not gotten used to the idea of using only `ReaderT` as the underlying monad for `streamly`. I would like `ExceptT`, `StateT` and `WriterT` to work with it.

mitchellwrosen:
Fair enough! Feel free to close this out, then.

For anyone else reading this, here is an example of how non-algebraic effects monads need to be handled with care when using `MonadBaseControl`:

```haskell
foobar :: IsStream t => t (StateT Int IO) ()
foobar = modify (+1) `consM` modify (+1) `consM` nil
```

When consumed serially, all state modifications occur to the one, global state...

```haskell
>>> execStateT (runStream (serially foobar)) 0
2
```

But when concurrency is involved, it really depends on how many threads are spawned, and how many cells each thread runs. I think streamly (by design) tries to hide these details from you, so I'm not sure what will be returned. But anyway, when I run it, I get 0.

```haskell
>>> execStateT (runStream (aheadly foobar)) 0
0
```

harendra-kumar:
Hmm, so you have figured this out :-) 👍  Perhaps this was the trigger for this PR. I am aware of this but I was not sure whether to bring this up. Apart from atomic modification of state there is a bug in restoring the monadic state, which is already fixed in my local git repo, will be committed soon.

This is work in progress and proper semantics for fork and join of monadic state are going to come in future. The stock implementation of `StateT` is serial, in all likelihood I will have a concurrent drop in replacement for `StateT` (maybe more than one flavor). 

Similarly, we have to define proper concurrency semantics for `ExceptT` and other stateful monads. And that is one of the reasons to keep `MonadBaseControl` or something equivalent, because it allows us to capture and restore the state which is needed to implement those semantics.

harendra-kumar:
@mitchellwrosen please take a look at #109, let me know if you have any feedback on that. It fixes some issues with monadic state used in the concurrent tasks. Each task inherits the state from the parent. However, current semantics discard any state modifications performed in the forked tasks. It will change in future though. The docs still need to be updated with exact semantics. With that, I am closing this PR.

harendra-kumar:
@mitchellwrosen please take a look at #109, let me know if you have any feedback on that. It fixes some issues with monadic state used in the concurrent tasks. Each task inherits the state from the parent. However, current semantics discard any state modifications performed in the forked tasks. It will change in future though. The docs still need to be updated with exact semantics. With that, I am closing this PR.

mitchellwrosen:
I have not yet used streamly so I'm not sure how much I can say about the PR :)

In general I'm not a fan of deferring some choice of semantics of caller of a computation. For example, I avoid writing `(MonadExcept e m, MonadState s m)` since `StateT` and `ExceptT` don't commute, and I'd rather just pick an ordering for simplicity and predictability.

For streamly the main abstraction is over a choice of IO interleavings, so the least surprising thing (for me) would be to observe different result orderings (and any downstream consequences of that) by picking different monads, but _not_ observe a significant change in semantics. This isn't true of `MonadBaseControl m` in general of course, so the semantics of 

```haskell
f <> do
  modify (+1) 
  snapshot1 <> snapshot2
g
```

are ambiguous in isolation - the choice of `m` determines whether `g` has observed the `(+1)`. It just causes a lot of pain and confusion during refactoring, in my experience, hence my preference for `ReaderT`/`MonadUnliftIO`. Anywho, you already knew all that! I'm sure there's a nice use case for non-algebraic monads, it's just tricky and confusing ;)

harendra-kumar:
I completely agree with that point. It is a discrepancy currently, as the serial stream `SerialT` behaves differently than concurrent streams. `SerialT` would observe the modify in your example whereas `AsyncT` and other concurrent monads wouldn't. I guess this is what you mean? This is _definitely_ not the right thing. All monads must have exactly the same semantics. 

It is a work in progress. It will be fixed in near future, it is the way it is because it is the simplest to implement and no attention has been to paid to it yet, doing it the right way would require some significant work, and some soak time is needed to finalize the APIs. I am thinking of a new type class to define the state joining semantics of various monads. The solution will possibly land in the next major version in some form.

PR #106, opened 2018-10-01
ran src through hlint, made most changes; added .hlint.yaml.

wygulmage:
There are still 30 hints for the src tree. One for unnecessary parentheses around a call to `inline`, and the rest for unused LANGUAGE pragmas, to reduce duplication, or for η-reduction.

I didn't do η-reduction when a function didn't have a type signature and had multiple parameters.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,0 +1,70 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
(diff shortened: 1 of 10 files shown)


harendra-kumar:
Thanks for your contribution @wygulmage ! Any changes that make the code better are valued greatly. Looking forward to more from you in future.

wygulmage:
Great!

PR #112, opened 2018-10-25
Documentation adjustments

mgsloan:
Just some tweaks to the documentation.  Let me know if I should split it into multiple PRs, feel free to only merge a subset of the commits.

### DIFF
--- a/src/Streamly/Tutorial.hs
+++ b/src/Streamly/Tutorial.hs
@@ -9,7 +9,7 @@
 -- Streamly is a general computing framework based on streaming IO. The IO
 -- monad and pure lists are a special case of streamly. On one hand, streamly
 -- extends the lists of pure values to lists of monadic actions, on the other
--- hand it extends the IO monad with concurrrent non-determinism. In simple
+-- hand it extends the IO monad with concurrent non-determinism. In simple
 -- imperative terms we can say that streamly extends the IO monad with @for@
 -- loops and nested @for@ loops with concurrency support. You can understand
 -- this analogy better once you can go through this tutorial.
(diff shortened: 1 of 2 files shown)


src/Streamly/Tutorial.hs
```diff
@@ -244,20 +250,20 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
...
 
 -- $flavors
 --
--- Streams can be combined using semigroup or monoid composition to form a
--- composite stream. Traversal of a composition of streams could be @deep@ or
--- @wide@.  Deep goes depth first i.e.  each stream is traversed fully before
--- we traverse the next stream. Wide goes breadth first i.e. one element from
--- each stream is traversed before coming back to the first stream again.
+-- Streams can be combined using '<>' or 'mappend' to form a composite
+-- stream. Deep streams are combined depth first i.e. each stream is traversed
```
harendra-kumar: It may appear that we are "combining" streams depth first or breadth first. But in fact combining does not matter because we can always interconvert the types from deep to wide or serial to parallel before we consume it. Therefore a better way of expressing the idea would be to explain consumption or traversal rather than combining because that is when it is decided whether we are traversing it depth first or breadth first.

So can we say something like this:

"Composite streams can be interpreted in a depth first or breadth first manner using an appropriate type conversion before consumption. Deep (e.g. 'Serial') stream type variants traverse a composite stream in a depth first manner whereas wide (e.g. 'WSerial') stream types traverse it in a breadth first manner".

harendra-kumar:
Great to see you here Michael! The changes look good. I have one comment posted in the review.

src/Streamly/Tutorial.hs
```diff
@@ -259,11 +259,14 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 
 -- $flavors
 --
--- Streams can be combined using '<>' or 'mappend' to form a composite
--- stream. Deep streams are combined depth first i.e. each stream is traversed
--- fully before we traverse the next stream. Wide streams are combined breadth
--- first i.e. one element from each stream is traversed before coming back to
--- the first stream again.
+-- Composite streams can be interpreted in a depth first or breadth
```
harendra-kumar: Looks good. Just one point, we have not hinted what is a composite stream or how it can be created e.g. using `<>` or `mappend`, that gives a more concrete idea about what we are talking about here.

mgsloan: Good point!  I've added the first sentence back in

harendra-kumar:
Looks good. Ready to merge. Do you want to squash the commits?

harendra-kumar:
Merged. Thanks Michael!

PR #118, opened 2018-10-30
WIP/POC - Add a cleanup continuation for cleanups on exception

harendra-kumar:
This is a more intrusive approach compared to register/releasing cleanup handlers in a global state. Especially because cons/uncons functions too have to be aware of cleanup handlers. Because of that I am tending towards using the register/release approach. This PR is to keep this change until the other approach is tried.


harendra-kumar:
We have already implemented exception handling combinators in Streamly.Prelude. Closing this.

Issue #123, opened 2018-11-02
Implement Foldable and Traversable instances for Streams


Abhiroop:
The core stream type of Streamly as defined in `StreamD.hs` and `StreamK.hs` could both support Foldable as well as Traversable instances. The only catch is it would require the underlying monad to be a Foldable or Traversable respectively.

### MOTIVATION

Having something like `(Foldable m, Monad m) => Foldable (Stream m)` has a number of advantages.

- When the underlying monad is `Identity`, the Streams effectively mirror a plain Haskell list. As mentioned in [1] Streamly attempts to faithfully reproduce the Haskell list API, this change would further allow the Streamly API to be even closer to plain Haskell lists (which are already instances of Foldable and Traversable).

- For any general monad `m`, Foldable streams would support all the utility functions from `Data.Foldable` like `asum`, `sequenceA_` etc

- When the underlying monad is `Sum` or `Max` or `Min` it would allow all the Foldable arithmetic operations across infinite streams of these respective monads.

Having something like `(Traversable m, Monad m) => Traversable (Stream m)` generally follows from the `Foldable` instance. The major benefit in this instance is again replicating the list API.

[1] https://github.com/composewell/streamly/blob/docs/docs/streamly-vs-lists.md

PR #124, opened 2018-11-02
Add Foldable and Traversable instances for StreamD and StreamK

Abhiroop:
Fixes https://github.com/composewell/streamly/issues/123

Need to add documentation about the use cases of these.


harendra-kumar:
1. We need to derive the instances for [newtypes as well](https://github.com/composewell/streamly/blob/c7666a7caa29d9f4e292d7c2c6733f6d5e0d9d55/src/Streamly/Streams/Serial.hs#L124), for all the stream types.
2. You can add some property tests in test/Prop.hs using the instances, so that we are sure that the instances are there and exported. There are tests about folds already, I guess we can just create new tests after those.
3. Adding a benchmark is a bonus, it will help us catch if we missed some inlining

Issue #165, opened 2019-01-04
Support ghcjs build


harendra-kumar:
Compile streamly with ghcjs, we will have to disable multiprocessor primitives like store/load barriers or implement them as no-ops for ghcjs. To make sure that it is not broken by new commits we will need a ghcjs CI build as well. Certain tests that are designed for multi-processor would have to disabled.

See ghcjs/ghcjs#705

harendra-kumar:
fixed by #166 .

PR #166, opened 2019-01-10
Add GHCJS support

harendra-kumar:
Most tests pass. There seems to be a bug in GHCJS ghcjs/ghcjs#635 which often causes the `maxrate` tests to fail and perhaps also the cause of an occasional "thread blocked indefinitely in MVar.." error because of the crash. 

I have also fixed `gauge` so we can also run the benchmarks. All benchmarks can run successfully.

All examples can also be run successfully except the search-query benchmark which needs network package which does not build with ghcjs.

The rate control does seem to be working perfectly due to low resolution timer in javascript. But we may be able to fix it even without a high resolution timer. The AcidRain example currently runs too fast and not at the specified rate.


harendra-kumar:
Fixes #165 .

harendra-kumar:
The high rate benchmarks improved significantly:

```
Generating reports for linear-rate...
(time)
Benchmark                                       default(0)(ms)(base) default(1)(%)(-base)
----------------------------------------------- -------------------- --------------------
avgrate/aheadly/unfoldrM/1,000,000                            100.09                -0.06
avgrate/asyncly/unfoldrM/Nothing                               34.24                -0.85
avgrate/asyncly/unfoldrM/10,000,000/maxThreads1                39.53                -1.46
avgrate/asyncly/unfoldrM/1,000,000                            102.61                -2.12
avgrate/asyncly/unfoldrM                                       30.91                -2.24
avgrate/asyncly/unfoldrM/3,000,000                             48.34               -21.32
avgrate/asyncly/unfoldrM/10,000,000                            62.30               -29.85
avgrate/asyncly/unfoldrM/20,000,000                            97.91               -49.65
```

This might be because of rate/control/clock related changes.

harendra-kumar:
fixes #165

Issue #160, opened 2018-12-22
unwanted consequences of restricting concurrency using maxYields


harendra-kumar:
For APIs that use only a few results from the stream we restrict concurrency to a maximum of those many threads (using `maxYields` internally). However, this may not be correct always, because the number of threads may not directly translate into number of results. For example, the following code effectively becomes serial, because `take 1` restricts the number of threads to 1:

```haskell
constraint p = if p then return () else S.nil

interleave :: WAsyncT IO (Word, Word)
interleave = do
    x <- S.enumerateFromTo 0 20
    y <- S.enumerateFromTo 0 20
    tid <- S.yieldM myThreadId
    S.yieldM $ putStrLn $ show tid ++ " " ++ show (x,y)
    constraint $ x == 20
    return (x, y)

main :: IO ()
main = do
    xs <- S.toList $ wAsyncly $ S.take 1 interleave
    print xs
```

Instead of restricting concurrency implicitly via operations like `take` we should give the control in the hands of the user, via maxBuffer etc. Maybe we can provide a configuration combinator to enable this behavior of `take` et al.

PR #182, opened 2019-03-20
Only dispatch a new worker if we haven't reached the limit.

### LINKED ISSUE #181: `maxThreads` ignored unless `Rate` is also specified.
```
λ S.runStream . S.maxThreads 1 . S.asyncly $ S.replicateM 5 (threadDelay 1000000)
(1.01 secs, 722,616 bytes)
λ S.runStream . S.avgRate 10000 . S.maxThreads 1 . S.asyncly $ S.replicateM 5 (threadDelay 1000000)
(5.02 secs, 736,720 bytes)
```

I would expect both of these to take 5 seconds due to `maxThreads 1` in both of them.

luke-clifton:
Fixes #181

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,16 +1,12 @@
-resolver: lts-12.11
+resolver: lts-13.13
 packages:
 - '.'
 extra-deps:
     - SDL-0.6.6.0
-    - gauge-0.2.4
     - Chart-1.9
     - Chart-diagrams-1.9
     - SVGFonts-1.6.0.3
     - bench-show-0.2.2
-    - statistics-0.15.0.0
-    - dense-linear-algebra-0.1.0.0
-    - math-functions-0.3.0.2
 
 flags: {}
 extra-package-dbs: []


harendra-kumar:
@luke-clifton thanks for the PR, I appreciate your time finding and fixing this. 

This fix breaks the previous fix targeted by commit cca869f7ed98fbe513cff1e67635917f9e16e113. You can see that by running these commands (please sync and rebase with master for the `no-charts` flag):

```
$ cabal new-build --flag dev --flag no-charts --disable-benchmarks maxrate
$ dist-newstyle/build/x86_64-osx/ghc-8.6.2/streamly-0.6.0/build/maxrate/maxrate
```

Compare the results of this test with and without the fix. You will notice that several tests now have the timings too far off the desired range. Note that if your machine is not powerful enough the very high rate tests might fail anyway in both cases.

So we can add something like the following, so that both with and without rate-control work well:

```
@@ -1426,10 +1426,13 @@ dispatchWorker yieldCount sv = do
                 Nothing -> return workerLimit
                 Just ref -> do
                     n <- liftIO $ readIORef ref
-                    return $
-                        case workerLimit of
-                            Unlimited -> Limited (fromIntegral n)
-                            Limited lim -> Limited $ min lim (fromIntegral n)
+                    case yieldRateInfo sv of
+                        Just _ -> return workerLimit
+                        Nothing ->
+                            return $
+                                case workerLimit of
+                                    Unlimited -> Limited (fromIntegral n)
+                                    Limited lim -> Limited $ min lim (fromIntegral n)
```

harendra-kumar:
Note that the above mentioned problem is due to the `maxYields` mechanism which tries to restrict the number of yields. The `remainingWork` in the code above is related to `maxYields` internal combinator. We use `maxYields` to restrict the yields to, for example, the number of elements specified by a `take` on the stream. The fix above is just disabling that mechanism in case rate-control is enabled. 

The `maxYields` mechanism is hacky and has other problems too, see #160. We should perhaps consider removing it.

luke-clifton:
That patch you posted doesn't work either. I don't currently have the time to dig into this much further unfortunately.

I guess I don't understand why limiting it they way I was breaks the tests you posted. I also don't see HOW that can be wrong. If `maxThreads` is set, I really don't want more than that being spawned ever. Thus imposing a hard limit seems like the right thing to do.

I'll see if I get some more time to take a look, but I'm pretty swamped at the moment.

harendra-kumar:
@luke-clifton I may not have been clear enough. The patch that I posted is in addition to your previous change, it has to be applied on top of that. So the hard limit stays, I did not intend to remove it. What that patch does is fix the `maxYields` problem that I pointed out in previous update, this is applicable only when rate-control is used.

PR #185, opened 2019-03-27
Fix maxThreads being ignored

### LINKED ISSUE #181: `maxThreads` ignored unless `Rate` is also specified.

luke-clifton:
Fixes #181

### DIFF
--- a/src/Streamly/SVar.hs
+++ b/src/Streamly/SVar.hs
@@ -1426,10 +1426,13 @@ dispatchWorker yieldCount sv = do
                 Nothing -> return workerLimit
                 Just ref -> do
                     n <- liftIO $ readIORef ref
-                    return $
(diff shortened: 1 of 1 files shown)


harendra-kumar:
Thanks @luke-clifton ! I merged the PR, added a test and updated the changelog as well.

PR #192, opened 2019-05-16
Implementation added for grouping, spanning Fold functions

aravindgopall:
Added grouped, span functions implementations which are undefined
before.
Might not follow some inline rules. Please give suggestions

### DIFF
--- a/src/Streamly/Fold.hs
+++ b/src/Streamly/Fold.hs
@@ -1220,13 +1220,20 @@ toArrayN limit = Fold step begin done
 -- prefix of length @n@ and the second group consisting of the rest of the
 -- stream.
 --
+{-# INLINE splitAt #-}
 splitAt
     :: Monad m
     => Int
     -> Fold m a b
     -> Fold m a c
     -> Fold m a (b, c)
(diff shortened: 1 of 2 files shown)


harendra-kumar:
@aravindgopall thanks a lot for the changes. The `next` branch is defunct now, all the changes from `next` have now been brought into master. Can you port your changes to a branch off master?

Also, you can take a look at `test/prop.hs` for existing tests and add property tests for these new ops as well. See the `groupSplitOps` function for example. For operations that exist for lists e.g. `groupBy`, we can straightforward compare the results of these with the corresponding list operations.  See `transformCombineOpsCommon` for example, we compare several stream ops with list ops in that, e.g. see the `reverse` test.

harendra-kumar:
Thanks for your contribution @aravindgopall . I merged the PR. I fixed the comparison order in `groupsRollingBy` via commit 2f2d2268c26a544db2c343130acfe4494e32478f and updated docs via commit 57987870113a8885f574f79c40df02b9785de7f6. Looking forward to more contributions from you in future.

PR #193, opened 2019-05-17
Implementation added for grouping, spanning Fold functions

aravindgopall:
Added grouped, span functions implementations which are undefined
before. 
-- Haven't added Prop tests, working on it

### DIFF
--- a/src/Streamly/Fold.hs
+++ b/src/Streamly/Fold.hs
@@ -172,13 +172,12 @@ module Streamly.Fold
     -}
 
     -- * Stream Splitting Operations
-    {-
     -- ** Spanning
     -- | Spanning splits the input into two groups and applies two different
     -- folds on each group.
 
     -- Element unaware spanning
-    -- , splitAt
+    , splitAt
 
     -- Element aware spanning
     , span
@@ -186,7 +185,6 @@ module Streamly.Fold
     , spanBy
     , spanRollingBy
     , spanned
-    -}
 
     -- ** Grouping
     -- | Grouping splits the stream into groups and applies the supplied fold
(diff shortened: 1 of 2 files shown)


src/Streamly/Fold.hs
```diff
@@ -952,14 +950,21 @@ ltakeWhile predicate (Fold step initial done) = Fold step' initial' done'
...
     => Int
     -> Fold m a b
     -> Fold m a c
     -> Fold m a (b, c)
-splitAt n = undefined
--}
+splitAt n (Fold stepL initL doneL) (Fold stepR initR doneR) = Fold step init done
+  where
+    step (ind,v1,v2) input =
+            if ind > 0 then stepL v1 input >>= (\a -> return (ind-1,a,v2))
```
harendra-kumar: We use the following style for if-then-else:

```
if x
then y
else x
```

Also, always put spaces around operators and after ",", e.g.

```
(ind - 1, a, v2)
```

Use `xL` and `xR` instead of `v1` and `v2`. That provides a hint about which stream they belong to.

src/Streamly/Fold.hs
```diff
@@ -952,14 +950,21 @@ ltakeWhile predicate (Fold step initial done) = Fold step' initial' done'
...
+                       else stepR v2 input >>= (\b -> return (ind,v1,b))
+    init  = (,,) <$> return n <*> initL <*> initR
```
harendra-kumar: 1) Move the `init` function before `step` function, that's the logical order to read the code.

2) Use `Pair3'` from `Streamly.Strict` instead of using `(,,)`. Always use a strict data type for the left fold accumulator.

PR #194, opened 2019-05-21
Fixes for iterateM, uniq, fromIndicesM

pranaysashank:


src/Streamly/Prelude.hs
```diff
@@ -677,8 +676,11 @@ fromIndices = fromStreamD . D.fromIndices
 --
 -- @since 0.6.0
 {-# INLINE fromIndicesM #-}
-fromIndicesM :: (IsStream t, Monad m) => (Int -> m a) -> t m a
-fromIndicesM = fromStreamD . D.fromIndicesM
+fromIndicesM :: (IsStream t, MonadAsync m) => (Int -> m a) -> t m a
```
harendra-kumar: 1) This will have bad performance in the SerialT case. We need to use rewrite rules to rewrite this to `D.fromIndicesM` in serial case. See the implementation of `unfoldrM` for an example.

2) Move this implementation to StreamK module and use `K.fromIndicesM` here.

harendra-kumar:
@pranaysashank can you please add a changelog entry under bug fixes for `uniq` and under breaking changes for `iterateM`? Also, make sure to once check the serial benchmark results for fromIndices/fromIndicesM.

pranaysashank:
### Serial Benchmarks for `fromIndices` and `fromIndicesM`
```
Benchmark      StreamD(μs)         StreamK(ms)
----------- ---------------- ------------------
fromIndices        37.90                 6.571
fromIndicesM       38.07                 4.691
```

harendra-kumar:
The `StreamD` and `StreamK` benchmarks are low level benchmarks. Rewrite rules are not in the picture in these benchmarks. To see whether the rewrite rules are helping we need to see if the performance of serial stream version of fromIndices in the `linear` benchmark is similar to what we see in `StreamD`.

pranaysashank:
Yes, I realise I haven’t been clear above but I tested that and the rewrite rules are being fired.

PR #202, opened 2019-05-30
Exposing the Fold constructor

paulvictor:


harendra-kumar:
@paulvictor I exposed it via the `Streamly.Internal` module.

PR #204, opened 2019-06-12
A few String functions

pranaysashank:


src/Streamly/FileSystem/Handle.hs
```diff
@@ -174,10 +175,10 @@ writeArray h Array{..} = withForeignPtr aStart $ \p -> hPutBuf h p aLen
 -- | @readArraysUpto size h@ reads a stream of arrays from file handle @h@.
```
harendra-kumar: Run cabal new-haddock, to check if the documentation is correct. This doc needs to be move to the enabled definition of readArraysUpto.

PR #215, opened 2019-07-24
Add documentation for string functions.

pranaysashank:


src/Streamly/String.hs
```diff
@@ -177,18 +194,56 @@ isSpace c
...
 {-# INLINE words #-}
 words :: (MonadIO m, IsStream t) => t m Char -> t m (Array Char)
 words = FL.wordsBy isSpace toArray
 
+-- | Flattens the stream of @Array Char@, after appending a terminating
+-- newline to each string.
+--
+-- 'unlines' is an inverse operation to 'lines'.
+--
+-- >>> S.toList $ unlines $ S.fromList ["lines", "this", "string"]
```
harendra-kumar: We may need to mention `OverloadedStrings` extension wherever we are relying on it. Novice users may not be able to figure that out easily and they may assume that the example just fails.

PR #218, opened 2019-07-25
Array ops

pranaysashank:


src/Streamly/Mem/Array.hs
```diff
@@ -271,9 +327,17 @@ splice arr1 arr2 = undefined
 -- first half retains values where the predicate is 'False' and the second half
 -- retains values where the predicate is 'True'.
 {-# INLINE partitionBy #-}
-partitionBy :: (a -> Bool) -> Array a -> (Array a, Array a)
-partitionBy f arr = undefined
+partitionBy :: (MonadIO m, Storable a) => (a -> Bool) -> Array a -> m (Array a, Array a)
```
harendra-kumar: This should be an in-place mutation primitive. We should just swap the elements in the existing array and then slice the array into two arrays by sharing the same underlying array just different start/end indices.

src/Streamly/Mem/Array.hs
```diff
@@ -321,27 +385,57 @@ readIndex arr i =
...
+
 -- | @readSlice arr i count@ streams a slice of the array @arr@ starting
 -- at index @i@ and reading up to @count@ elements in the forward direction
 -- ending at the index @i + count - 1@.
 --
+-- >>> S.toList $ readSlice (fromList [1..10]) 3 4
+-- > [4,5,6,7]
+--
 -- @since 0.7.0
 {-# INLINE readSlice #-}
```
harendra-kumar: Make it `INLINE_NORMAL` so that it can be used efficiently in higher order functions.

src/Streamly/Mem/Array.hs
```diff
@@ -321,27 +385,57 @@ readIndex arr i =
...
-readSlice :: (IsStream t, Monad m, Storable a)
+readSlice :: (IsStream t, MonadIO m, Storable a)
     => Array a -> Int -> Int -> t m a
-readSlice arr i len = undefined
+readSlice arr i len = D.fromStreamD (D.Stream step i)
+  where
+    step _ j
```
harendra-kumar: Add `INLINE_LATE` as usual.

src/Streamly/Mem/Array.hs
```diff
@@ -321,27 +385,57 @@ readIndex arr i =
...
+        | j < i + len = do
+            r <- liftIO $ readIndex arr j
```
harendra-kumar: `readIndex` performs a bounds check every time and then we are checking for the return value for Nothing or Just. We can avoid that by reading directly instead of using `readIndex`, that will be more efficient.

src/Streamly/Mem/Array.hs
```diff
@@ -218,28 +222,81 @@ _last arr = readIndex arr (length arr - 1)
...
-bsearch :: a -> Array a -> Maybe Bool
-bsearch = undefined
+{-# INLINE bsearch #-}
+bsearch :: (MonadIO m, Ord a , Storable a) => a -> Array a -> m Bool
+bsearch ele arr = go 0 (length arr - 1)
+  where
+    go low high
+        | high < low = return False
+        | otherwise = do
+            r <- readIndex arr mid
```
harendra-kumar: `readIndex` is not an efficient way of reading because it performs bounds checks every time. See my comments in `readSlice` for the same.

src/Streamly/Mem/Array.hs
```diff
@@ -359,27 +453,55 @@ writeIndex arr i a = do
...
-{-
+
 -- | @writeSlice arr i count stream@ writes a stream to the array @arr@
 -- starting at index @i@ and writing up to @count@ elements in the forward
 -- direction ending at the index @i + count - 1@.
 --
 -- @since 0.7.0
 {-# INLINE writeSlice #-}
-writeSlice :: (IsStream t, Monad m, Storable a)
+writeSlice :: (IsStream t, MonadIO m, Storable a)
```
harendra-kumar: Comments in readSlice apply here as well. Avoid `writeIndex` as performs bounds checks every time.

PR #224, opened 2019-07-28
Add credits and copyrights

harendra-kumar:
I have added the names of the contributors to the library and licenses for the third party source code. 

I was not sure if just including the one line copyright notice is enough to meet the license conditions, so I included the full licenses from the original packages. The copyright notices are also mentioned in the respective files as well wherever we have included the code.

Any suggestions are welcome!

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -40,6 +40,10 @@
 
     * Change the `Unreleased` section at the top of changelog file to the new
       release version number.
+    * Make sure contributors to the release are listed in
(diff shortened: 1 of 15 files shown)


pranaysashank:
Some primitives for encode/decodeUtf8 are lifted from base, and transitively text. I think we also have to include the authors mentioned in the copyright at http://hackage.haskell.org/package/base-4.12.0.0/docs/src/GHC.IO.Encoding.UTF8.html

PR #228, opened 2019-07-30
Hide modules whose apis have not yet been finalised for 0.7.0

pranaysashank:
Updated Changelog, the benchmarking ops, and the File IO example in README as well

### DIFF
--- a/examples/FileIOExamples.hs
+++ b/examples/FileIOExamples.hs
@@ -1,7 +1,7 @@
 import qualified Streamly.Prelude as S
 import qualified Streamly.Fold as FL
 import qualified Streamly.Mem.Array as A
-import qualified Streamly.FileSystem.Handle as Handle
+import qualified Streamly.FileSystem.File as File
 
 import Data.Char (ord)
 import System.Environment (getArgs)
--- a/examples/FileIOExamples.hs
+++ b/examples/FileIOExamples.hs
@@ -1,7 +1,7 @@
 import qualified Streamly.Prelude as S
 import qualified Streamly.Fold as FL
 import qualified Streamly.Mem.Array as A
-import qualified Streamly.FileSystem.File as File
+import qualified Streamly.FileSystem.Handle as Handle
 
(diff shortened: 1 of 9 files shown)


examples/FileIOExamples.hs
```diff
@@ -1,7 +1,7 @@
 import qualified Streamly.Prelude as S
 import qualified Streamly.Fold as FL
 import qualified Streamly.Mem.Array as A
-import qualified Streamly.FileSystem.File as File
+import qualified Streamly.FileSystem.Handle as Handle
```
harendra-kumar: Why this change? Is this accidental?

pranaysashank: Yes, this is accidental.

examples/FileSinkServer.hs
```diff
@@ -9,21 +9,24 @@ import System.Environment (getArgs)
...
 import qualified Streamly.Network.Server as NS
 import qualified Streamly.Prelude as S
 
+import System.IO (withFile, IOMode(..))
+
 main :: IO ()
 main = do
     file <- fmap head getArgs
-    File.append file
+    withFile file AppendMode
```
harendra-kumar: Where is `withFile` coming from?

pranaysashank: From System.IO

harendra-kumar: Now I see, its right there, my brain just filtered that out because I was expecting it to come from some Streamly module.

harendra-kumar:
This change will render some APIs unusable but visible, because types like `AbsTime` are not available. Should we hide them instead, users will think that they are there but they can't use them?

pranaysashank:
Oh yeah, I forgot about the apis that depend on time. Those apis are cool, I guess we can either use a `Num a` instead of AbsTime (which would make the change to AbsTime a breaking change later) or hide them and ship them later

harendra-kumar:
Let's hide them for now, they can be added later in a minor release as well.

PR #233, opened 2019-08-01
Fix a bug in zipWithM rule that causes effects to run only once.

pranaysashank:
I looked through other rules and as far as I can tell they are safe, these was the only one that needed fixing to run only in Identity.


harendra-kumar:
We can disable the rule entirely for GHC 7.10.3.

pranaysashank:
The Maintaining guide mentions all bug fixes should have a target issue, do I need to create an issue for this?

harendra-kumar:
Ideally, yes. But depends on if you have time to do that.

PR #234, opened 2019-08-01
Documentation, API name/signature changes for 0.7.0

harendra-kumar:


src/Streamly/Fold.hs
```diff
@@ -1665,16 +1694,18 @@ splitSuffixBy' predicate f m =
...
+-- sequence, that way we can avoid the Storable constraint. If we still need
+-- Storable Array for performance, we can use a separate splitOnArray API for
+-- that. We can also have an API where the sequence itself is a lazy stream, so
+-- that we can search files in files for example.
+{-# INLINE splitOnSeq #-}
+splitOnSeq
     :: (IsStream t, MonadIO m, Storable a, Enum a, Eq a)
     => Array a -> Fold m a b -> t m a -> t m b
-splitOn patt f m = D.fromStreamD $ D.splitOn patt f (D.toStreamD m)
+splitOnSeq patt f m = D.fromStreamD $ D.splitOn patt f (D.toStreamD m)
```
pranaysashank: Don't think this is a good name, how about using full sequence instead, `splitOnSequence`

harendra-kumar: It becomes a little longer. The problem may not be that this becomes longer, but the problem is that if we introduce its cousins like `splitOnSuffixSequence` later they become even longer. I am still willing to consider it if the `Seq` prefix sounds less intuitive/confusing, its just trade off with longer name vs more explicit.

src/Streamly/Prelude.hs
```diff
@@ -872,7 +874,7 @@ each = K.fromFoldable
 -- | Read lines from an IO Handle into a stream of Strings.
 --
 -- @since 0.1.0
-{-# DEPRECATED fromHandle "Please use Streamly.FileIO instead." #-}
+{-# DEPRECATED fromHandle "Please use Streamly.FileSystem.Handle instead." #-}
```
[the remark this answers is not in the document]
harendra-kumar: We have to use `decodeChar8` or `decodeChar8Unchecked` and `FL.splitSuffixBy (== '\n') toList` for equivalent functionality. We can mention that in the deprecation message.

src/Streamly/Fold.hs
pranaysashank: Does using a `Splitter` record like the split package, as mentioned in one comment, do away with the long names?

src/Streamly/Mem/Array.hs
```diff
@@ -456,6 +463,10 @@ toArray = toArrayMinChunk (bytesToCount (undefined :: a) (A.mkChunkSize 1024))
 
 -- | Convert a stream of arrays into a stream of their elements.
 --
+-- Same as the following but more efficient:
```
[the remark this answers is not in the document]
harendra-kumar: Can't figure out the difference between the two. It has the "same" effect.

src/Streamly/Fold.hs
harendra-kumar: I am not sure if we will ever implement something like that, it may not be worth the complexity. Just using prefix, infix, suffix style splitting covers a lot of ground, and we may not even want these as we will have other ways, like parser combinators, to achieve custom splitting.

PR #250, opened 2019-08-06
Minor updates to Streamly.Tutorial.

pranaysashank:


src/Streamly/Tutorial.hs
```diff
@@ -170,7 +170,7 @@ module Streamly.Tutorial
     )
 where
 
-import Streamly
+import Streamly hiding (foldWith, foldMapWith, forEachWith)
```
harendra-kumar: Do we still need to hide these if we import Streamly.Prelude qualified?

pranaysashank: Way down at lines 920, 922, 924. We refer these functions in the documentation, where haddock creates links to the ones in Streamly module which are deprecated, however hiding these functions here makes haddock create links to the ones in Prelude. 

Haddock is probably creating links to the first module where it can find a particular function. So I guessed hiding here forces haddock to look in the next module, Streamly.Prelude.

harendra-kumar: Ok.

PR #262, opened 2019-08-16
Add decodeUtf8Lenient to String module.

pranaysashank:
This commit doesn't affect current performance.

### DIFF
--- a/src/Streamly/Streams/StreamD.hs
+++ b/src/Streamly/Streams/StreamD.hs
@@ -250,6 +250,7 @@ module Streamly.Streams.StreamD
     -- * UTF8 Encoding / Decoding transformations.
     , decodeUtf8
     , encodeUtf8
+    , decodeUtf8Lenient
     )
 where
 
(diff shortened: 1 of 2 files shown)


src/Streamly/Streams/StreamD.hs
```diff
@@ -2558,25 +2559,43 @@ data CodePoint = FreshPoint
...
+    = TransliterateCodingFailure
+    | ErrorOnCodingFailure
+    deriving (Show)
+
+{-# INLINE_NORMAL decodeUtf8With #-}
+decodeUtf8With :: Monad m => CodingFailureMode -> Stream m Word8 -> Stream m Char
+decodeUtf8With cfm (Stream step state) = Stream step' (state, FreshPoint)
   where
+    {-# INLINE transliterateOrError #-}
+    {-# INLINE inputUnderflow #-}
```
harendra-kumar: Better to keep the pragma along with definition.

src/Streamly/Streams/StreamD.hs
```diff
@@ -2558,25 +2559,43 @@ data CodePoint = FreshPoint
...
+    transliterateOrError e s =
+        case cfm of
+            ErrorOnCodingFailure -> error e
+            TransliterateCodingFailure -> Skip (s, YieldAndContinue '\xFFFD')
+    inputUnderflow s = case cfm of
+       ErrorOnCodingFailure -> error "Input Underflow"
```
harendra-kumar: Should we use something like "decodeUtf8With: Not enough input to decode."? Ideally, we should use "Streamly.String.decodeUtf8" or "Streamly.String.decodeUtf8Lenient" as identification but for that we will have to pass the identification string here.

src/Streamly/Streams/StreamD.hs
```diff
@@ -2558,25 +2559,43 @@ data CodePoint = FreshPoint
...
         r <- step (adaptState gst) st
         return $
             case r of
                 Yield c0 s
                     | c0 <= 0x7f ->
-                        Skip (s, YieldPoint $ unsafeChr (fromIntegral c0))
-                    | c0 >= 0xc0 && c0 <= 0xc1 -> error "Invalid Overlong forms"
+                        Skip (s, YieldAndContinue $ unsafeChr (fromIntegral c0))
+                    | c0 >= 0xc0 && c0 <= 0xc1 ->
+                        transliterateOrError "Invalid Overlong forms" s
```
harendra-kumar: All errors should have some identification to tell the programmer/user where they are coming from. Ideally we should also print some context and what exactly is invalid so that the user can debug.

src/Streamly/Streams/StreamD.hs
```diff
@@ -2587,10 +2606,11 @@ decodeUtf8 (Stream step state) = Stream step' (state, FreshPoint)
...
             case r of
                 Yield c1 s
-                    | c1 < 0x80 || c1 >= 0xc0 -> error "Invalid Sequence"
-                    | otherwise -> Skip (s, YieldPoint $ chr2 c0 c1)
+                    | c1 < 0x80 || c1 >= 0xc0 ->
+                        transliterateOrError "Invalid Sequence" s
+                    | otherwise -> Skip (s, YieldAndContinue $ chr2 c0 c1)
                 Skip s -> Skip (s, Two c0)
-                Stop -> error "Input Underflow"
+                Stop -> inputUnderflow st
```
harendra-kumar: Here we can say something like "need one more codepoint after 0xc0" to know the context?

harendra-kumar:
We can make the errors more informative to help debug issues in decoding.

Issue #242, opened 2019-08-06
Core types for full composability


harendra-kumar:
Currently we have two fundamental (and dual) types, `Stream` for composing stream sources and `Fold` for composing stream consumers. For full composablity we need a fe more types, which includes:

* An `Unfold` type to compose sources with a source state, the `Stream` type is a special case of unfold where there is no source seed.

* A `Pipe` type to represent composable transformations. The transformations in `Streamly.Prelude` should be representable using the `Pipe` type. Also, the contravariant trasnformations of folds should be representable by pipes. A pipe can be attached covariantly to a source and contravariantly to a fold to extend it. A pipe would be a combination of fold and unfold. Pipes would enable splitting a stream into many streams and merging them back into a single stream.

* The `Fold` type needs to use a generalized state type like the `Step` in StreamD. That will enable termination of folds and representation of backtracking parsers using folds.

harendra-kumar:
See #2684.

PR #264, opened 2019-08-17
Use Unfolds for better concatMap fusion

harendra-kumar:


harendra-kumar:
See issue #242 as well.

PR #271, opened 2019-09-15
Implement the combinators in ArrayStream module generically

harendra-kumar:
To finally get rid of these from the Array module and be able to implement such combinators generically for a stream of any kind of containers.

@pranaysashank take a look, if I screwed up something.


src/Streamly/Data/String.hs
```diff
@@ -242,6 +247,15 @@ lines = foldLines A.write
 words :: (MonadIO m, IsStream t) => t m Char -> t m (Array Char)
 words = foldWords A.write
 
+-- | Unfold a stream to character streams using the supplied 'Unfold'
+-- and concat the results suffixing a newline character @\\n@ to each stream.
+--
+-- > unfoldLines = S.intercalateSuffix UF.singleton '\n'
+--
```
pranaysashank: An example here would be nice

src/Streamly/Memory/Array/Types.hs
```diff
@@ -893,6 +900,40 @@ read = D.fromStreamD . toStreamD
...
+    -- ccall?
+    liftIO $ do
+        memcpy (castPtr dst) (castPtr src1) len1
+        touchForeignPtr (aStart arr1)
+        memcpy (castPtr (dst `plusPtr` len1)) (castPtr src2) len2
+        touchForeignPtr (aStart arr2)
+    return arr { aEnd = dst `plusPtr` (len1 + len2) }
+
+instance Storable a => Semigroup (Array a) where
+    arr1 <> arr2 = unsafeInlineIO $ spliceTwo arr1 arr2
```
pranaysashank: From my understanding, `unsafeInlineIO` should not be used if we are allocating memory. There is a chance that multiple calls to malloc are shared, resulting in different arrays sharing the same memory address. I don't know if this is not true in newer releases.

src/Streamly/Memory/Array/Types.hs
```diff
@@ -893,6 +900,40 @@ read = D.fromStreamD . toStreamD
...
+
+nullForeignPtr :: ForeignPtr a
+nullForeignPtr = ForeignPtr nullAddr# (error "nullForeignPtr")
```
pranaysashank: Why not use `newForeignPtr_ nullPtr`, it would deal with the finalizer.

src/Streamly/Benchmark/FileIO/Array.hs
```diff
@@ -124,10 +138,25 @@ linesUnlinesCopy :: Handle -> Handle -> IO ()
 linesUnlinesCopy inh outh =
     S.runFold (FH.writeArraysInChunksOf (1024*1024) outh)
         $ Internal.insertAfterEach (return $ A.fromList [10])
-        $ A.splitOn 10
+        $ A.splitOnSuffix 10
         $ FH.readArraysOf (1024*1024) inh
 
 #ifdef INSPECTION
--- inspect $ hasNoTypeClasses 'linesUnlinesCopy
+inspect $ hasNoTypeClassesExcept 'linesUnlinesCopy [''Storable]
```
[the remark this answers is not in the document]
harendra-kumar: This is for `realloc`, realloc is anyway not in the fast path so we do not care.  But this exception here will also mask anyone else using Storable dictionary.

src/Streamly/Data/String.hs
harendra-kumar: I know, but too many things to do, did not have time for that :-)

src/Streamly/Memory/Array/Types.hs
harendra-kumar: `newForeignPtr_` uses `newIORef`, something expensive for creating just an empty Array.

src/Streamly/Memory/Array/Types.hs
harendra-kumar: Yes, theoretically GHC can potentially optimize/float multiple mallocs into one thinking they are perfectly pure computations, though in this case it should not be easy for the compiler to do it. However, to be sure we should definitely change it to `unsafeDupablePerformIO`.

PR #274, opened 2019-09-18
Improve demuxing

treeowl:
* Use `alterF` to speed up `Map` modification.

* Make `demuxWith` and `demuxWith_` just powerful enough to implement
`demux` and `demux_` without extra mapping. We could actually go
further if we wanted.

### DIFF
--- a/src/Streamly/Fold.hs
+++ b/src/Streamly/Fold.hs
@@ -925,24 +925,24 @@ partition = partitionBy id
 -- @since 0.7.0
 {-# INLINE demuxWith #-}
 demuxWith :: (Monad m, Ord k)
-    => (a -> k) -> Map k (Fold m a b) -> Fold m a (Map k b)
+    => (a -> (k, a')) -> Map k (Fold m a' b) -> Fold m a (Map k b)
 demuxWith f kv = Fold step initial extract
 
     where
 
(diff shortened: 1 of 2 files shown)


harendra-kumar:
@treeowl thanks for the PR. The changes looks good to me. I was not aware of the `alterF` operation, I needed this elsewhere for keeping mutable cells as values in the Map. In fact, I forked `containers` and implemented alterF for that experiment. I am feeling stupid now, I did not look carefully in the library.

BTW, the demux fold is not very efficient. It mutates the `Map` on every step of the fold, generating a lot of work for GC. We should perhaps use an API that forces a mutable value (e.g. an IORef or a Vector/MArray) in the map to avoid that problem. I got more than 30% improvement in the word-classifier example by using a mutable value in the hashmap.

treeowl:
It would be worth checking whose `alterF` performs better with typical streams. I wanted mine to be good for potentially expensive functors, which are likely unusual here.

harendra-kumar:
Both implementations are more or less the same with a marginal difference. The perf difference comes from the fact that we use a mutable IORef to keep the map immutable on updates, which is irrespective of the implementation of `alterF`.

harendra-kumar:
Thanks!

PR #275, opened 2019-09-19
Simplify and optimize utf-8 decoding

harendra-kumar:
* Simplify the state machine to carry minimal state
* reduce branches taken and improve branch prediction

With this change decode-utf8 has almost the same performance as decodeChar8 for
the ascii case. For the mixed unicode chars case as well the performance is
pretty good and not too far away from the ascii case.

Stream and array versions both fuse to similar core which pretty small and both
cases show exactly the same performance.

### DIFF
--- a/src/Streamly/Streams/StreamD.hs
+++ b/src/Streamly/Streams/StreamD.hs
@@ -3049,27 +3049,41 @@ utf8d = unsafePerformIO $ runFold (A.writeAligned 64) $ fromList [
 unsafePeekElemOff :: forall a. Storable a => Ptr a -> Int -> a
 unsafePeekElemOff p i = let !x = A.unsafeInlineIO $ peekElemOff p i in x
 
-{-# INLINE decode #-}
-decode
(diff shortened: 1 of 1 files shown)


harendra-kumar:
Benchmarks:

```
readArray/sum                            time                 497.9 ms 
readArray/read-utf8                      time                 413.2 ms
readStream/sum                           time                 330.9 ms 
readStream/read-word8                    time                 316.8 ms  
readStream/read-char8                    time                 323.3 ms  
readStream/read-utf8                     time                 409.2 ms  
decode-encode/char8                      time                 1.242 s   
decode-encode/utf8-arrays                time                 12.09 s   
decode-encode/utf8                       time                 14.34 s   
```

harendra-kumar:
Performance of streamly compared to `text/bytestring` (lazy) packages, utf-8 decoding and computing length of a 500MB text file:

```
streamly (ascii chars): 0.414s
text (ascii chars): 0.653s

streamly (all unicode chars): 0.890s
text (all unicode chars): 1.652s
```

PR #277, opened 2019-09-22
Implementation of Doubly Linked List (Pinned Memory)

adithyaov:
The implementation is similar to the `List` module in the `linkedlist` branch.
Still in the initial phase of development.

Potential implementation changes: Eliminate `Null`

Still REPL testing. Will change the implementation accordingly.

**TODO:** Add more helper functions.
**TODO:** Add tests.
**TODO:** Improve docs.

**Note:** Some questions I ask may seem silly as I've not played around with FFI a lot. Apologies in advance.


src/Streamly/Memory/DList.hs
```diff
@@ -0,0 +1,117 @@
...
+import Foreign.ForeignPtr (ForeignPtr, withForeignPtr)
+import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
+import Foreign.Ptr (Ptr, nullPtr)
+import Foreign.Storable (Storable(..))
+import GHC.ForeignPtr (mallocPlainForeignPtr)
+
+-- Either listHead or listTail could be Ptr. We need only
+-- one of them to be ForeignPtr. But both of them here
+-- are described as ForeignPtr for consistency.
+data DList a = DList
```
adithyaov: What if I just use `Ptr` and eliminate `ForeignPtr` completely? If so, I can eliminate `Null` by replacing it with `DList nullPtr nullPrt 0`.

pranaysashank: A `ForeignPtr` is defined as `data ForeignPtr a = ForeignPtr Addr# ForeignPtrContents`.
 You can just check whether the address equals `nullAddr#` to see if it is null. Look at `nullForeignPtr` defined in `Streamly.Memory.Array.Types`

adithyaov: Will check it out. But why use ForeignPtr in the first place though? Is this related to finalizers in any way? The code in `Memory.List` uses `mallocPlainForeignPtr` which I believe does not associate the `ForeignPtr` with a finalizer. I might be missing something here.

pranaysashank: Using just a `Ptr` ties us too much to the current GHC implementation. Using a `ForeignPtr` also allows us to swap out the `mallocPlainForeignPtr` to a different implementation, maybe to the system's malloc which allocates memory that resides outside the RTS storage manager which needs a finalizer to free the memory.
I think it's more to do with best practices and future proofing the code than anything else.

Also, I haven't tested `Memory.List`. There might be a bug in it somewhere that doesn't create new lists properly / share it with the old list.

adithyaov: Ah, I see, Indeed, that makes sense.

>Also, I haven't tested Memory.List. There might be a bug in it somewhere that doesn't create new lists properly / share it with the old list.

I'll let you know if I find something.

pranaysashank:
@adithyaov The preferred to way sync up your local branch with master is to rebase your branch on top of master instead of merging. We only merge when pulling another branch into master. This helps in keeping a clean commit history.

adithyaov:
I'll work on this by creating another branch and cleaning things up.

Issue #256, opened 2019-08-12
Add mmorph-style operations


treeowl:
I can't guarantee that these are right, but I think they probably make sense.

```haskell
hoistSerialT :: (Monad m, Monad n) => (forall x. m x -> n x) -> SerialT m a -> SerialT n a
hoistSerialT f = embedSerialT (lift . f)

embedSerialT
  :: (Monad m, Monad n)
  => (forall x. m x -> SerialT n x)
  -> SerialT m a -> SerialT n a
embedSerialT f xs =
  (f $ SP.uncons xs) >>= \case
    Nothing -> nil
    Just (a, as) -> yield a <> embedSerialT f as
```

I'm not sure whether access to the internals would allow the looser constraints needed for instances of `MFunctor` and `MMonad`.

pranaysashank:
I don't think it is possible to write an instance for MFunctor, we need a Monad constraint on the transformed stream as well. However we can provide similar functions to hoist, implemented internally, from Streamly.Prelude in future releases. #79

harendra-kumar:
@treeowl your hoist code looks good to me. I tried putting your code in StreamK.hs and make it polymorphic, like so:

```
embed
  :: (IsStream t, Monad m, Monad n, Monad (t n))
  => (forall x. m x -> t n x) -> t m a -> t n a
embed f = go
    where
    go xs = do
        r <- f $ uncons xs
        case r of
            Nothing -> nil
            Just (a, as) -> a `cons` embed f as

hoist :: (IsStream t, Monad m, Monad n, MonadTrans t, Monad (t n))
    => (forall x. m x -> n x) -> t m a -> t n a
hoist f = embed (lift . f)
```

We can export it through the `Streamly.Internal` module for now, that's where we keep experimental APIs which can potentially be included in future releases. We will be happy to accept a PR if you feel like raising one.

We can also add a benchmark to see how it performs. We can add a simple benchmark hoisting to Identity in `benchmark/LinearOps.hs/Linear.hs`. Like we have for `foldrT` which switches only the transformer monad.

treeowl:
I don't know much about `streamly` yet. Vague intuition suggests that for `SerialT` these will work for any natural transformation, but that for at least some of the others they only make sense for monad morphisms. Is that right? If so, I'll need to document that.

pranaysashank:
[...]

PR #302, opened 2019-11-09
Hoist and generalizing routines

harendra-kumar:
Added `hoist` and `generally`. The latter is a special case of `hoist` to generalize a pure stream to a general monad.

The StreamK code is based on #256 . The original code based on `uncons` required `MonaTrans t` and `Monad (t m)` constraints and was 2x slower (9ms vs 4ms for `length`) compared to the improved version by @pranaysashank .

However, the StreamK code is not being used in the exposed combinators. We are using the `StreamD` version which fuses well and is much faster (33 us for `length`), adding the `hoist` combinators makes no difference at all to `length`.

### DIFF
--- a/benchmark/Linear.hs
+++ b/benchmark/Linear.hs
@@ -48,6 +48,13 @@ benchIOSink
     => String -> (t IO Int -> IO b) -> Benchmark
 benchIOSink name f = bench name $ nfIO $ randomRIO (1,1) >>= f . Ops.source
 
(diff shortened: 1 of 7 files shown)


harendra-kumar:
I was tight on time so I implemented only more useful ones. Yes, we should be able to easily implement WriterT as well. Yeah, where order is not important (commutative monoids) all types of concurrent streams should work.

PR #308, opened 2019-11-14
Put utf8 design from Björn Höhrmann's website under design/.

pranaysashank:
This is compatible and allowed with his license. We can update the doc
as we make any changes in the future.

#


harendra-kumar:
We need to add it to the cabal file as well so that it is shipped with the package.

Issue #309, opened 2019-11-17
Move the Benchmark.* modules to Internal


harendra-kumar:
These modules are showing up as not "Internal" in the haddock docs on Hackage.

```
Streamly
  Benchmark
    FileIO
      Streamly.Benchmark.FileIO.Array
      Streamly.Benchmark.FileIO.Stream
   Streamly.Benchmark.Prelude
```

adithyaov:
I can take this up?

PR #310, opened 2019-11-19
Move the Benchmark.* modules to Internal

adithyaov:
Solves #309


pranaysashank:
These benchmarks should ideally be moved to a different component, so that when we use the plugin, the inspection testing doubles as a check for Plugin and vice-versa

harendra-kumar:
Let's commit the current change and then we can raise another issue for further changes.

PR #312, opened 2019-11-20
Fix bug in findIndices and add a regression test. Fixes #311.

### LINKED ISSUE #311: findIndices returns wrong indices.
```Haskell

main :: IO ()
main = S.mapM_ print $ S.findIndices (< 6) $ S.filter (> 3) $ S.fromList [1..10]
```
Compile with -O2

Results:
```
3
4
```

Expected Result:
```
0
1
```

pranaysashank:
- Update Changelog.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+* Fix a bug that caused `findIndices` to increment the index during
+  a Skip condition.
+
 ## 0.7.0
 
 ### Breaking changes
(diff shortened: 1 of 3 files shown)


Changelog.md
```diff
@@ -1,3 +1,10 @@
+## Unreleased
+
+### Bug Fixes
+
+* Fix a bug that caused `findIndices` to increment the index during
+  a Skip condition.
```
harendra-kumar: Just write user facing behavior in the changelog, not internal details like `Skip`. The goal is to explain in sucha way that the user of the library can understand the bug behavior and identify it when they hit it without having to understand the design details of the library.

PR #322, opened 2019-11-30
Implementation of unpinned arrays

pranaysashank:
Only a few api's are currently exposed. Will add others as needed if the base implementation looks correct.

- fromStreamN
- fromStream
- fromListN
- fromList
- writeN
- write
- toStream
- toStreamRev
- read
- fold
- streamFold

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/ArrayOps.hs
+++ b/benchmark/Streamly/Benchmark/Data/ArrayOps.hs
@@ -16,7 +16,7 @@
 module Streamly.Benchmark.Data.ArrayOps where
 
 import Control.Monad.IO.Class (MonadIO)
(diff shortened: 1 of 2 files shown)


benchmark/Streamly/Benchmark/Data/ArrayOps.hs
```diff
@@ -0,0 +1,153 @@
...
+#ifdef DEVBUILD
+import qualified Data.Foldable as F
+#endif
+
+import qualified Streamly           as S hiding (foldMapWith, runStream)
+import qualified Streamly.Data.Array as A
+import qualified Streamly.Prelude   as S
+
+value, maxValue :: Int
+#ifdef LINEAR_ASYNC
```
harendra-kumar: we can remove this.

PR #324, opened 2019-12-05
Simple combinators

adithyaov:
Simple one-line combinators.
`iterate`
`iterateM`
`drainN`
`drainWhile`
`rollingHashFirstN`
The changes were small, hence used a single PR.

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -56,7 +56,7 @@ module Streamly.Internal.Data.Fold
     , stdDev
     , rollingHash
     , rollingHashWithSalt
-    -- , rollingHashFirstN
+    , rollingHashFirstN
     -- , rollingHashLastN
 
     -- ** Full Folds (Monoidal)
@@ -70,8 +70,8 @@ module Streamly.Internal.Data.Fold
     , toListRevF  -- experimental
 
     -- ** Partial Folds
-    -- , drainN
-    -- , drainWhile
+    , drainN
+    , drainWhile
     -- , lastN
     -- , (!!)
     -- , genericIndex
@@ -563,6 +563,14 @@ defaultSalt = 0x087fc72c
 rollingHash :: (Monad m, Enum a) => Fold m a Int
 rollingHash = rollingHashWithSalt defaultSalt
 
+-- | Compute an 'Int' sized polynomial rolling hash of the first n elements of
+-- a stream.
+--
+-- > rollingHashFirstN = ltake n rollingHash
+{-# INLINABLE rollingHashFirstN #-}
+rollingHashFirstN :: (Monad m, Enum a) => Int -> Fold m a Int
+rollingHashFirstN n = ltake n rollingHash
+
 ------------------------------------------------------------------------------
 -- Monoidal left folds
 ------------------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -563,6 +563,11 @@ defaultSalt = 0x087fc72c
 rollingHash :: (Monad m, Enum a) => Fold m a Int
 rollingHash = rollingHashWithSalt defaultSalt
 
+
+{-# INLINABLE rollingHashFirstN #-}
+rollingHashFirstN :: (Monad m, Enum a) => Int -> Fold m a Int
```
harendra-kumar: It will be nice to have brief haddock documentation about these combinators. It will also be nice to have some simple tests for these.

src/Streamly/Internal/Prelude.hs
```diff
@@ -751,10 +751,8 @@ repeatMSerial = fromStreamS . S.repeatM
 -- @
 --
 -- @since 0.1.2
-iterate :: IsStream t => (a -> a) -> a -> t m a
-iterate step = K.fromStream . go
-    where
-    go s = K.cons s (go (step s))
+iterate :: (Monad m, IsStream t) => (a -> a) -> a -> t m a
```
harendra-kumar: Instead of removing the old implementation of `iterate` and `iterateM` move it to the StreamK module. Here use `fromStreamS . S.iterate` so that appropriate implementation (StreamD or StreamK) can be picked based on the `USE_STREAMK_ONLY` cpp macro.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -631,6 +639,14 @@ toList = Fold (\f x -> return $ f . (x :))
 -- Partial Folds
 ------------------------------------------------------------------------------
 
+{-# INLINABLE drainN #-}
+drainN :: Monad m => Int -> Fold m a () 
+drainN n = ltake n drain
+
+{-# INLINABLE drainWhile #-}
```
adithyaov: Add documentation!

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -631,6 +639,14 @@ toList = Fold (\f x -> return $ f . (x :))
 -- Partial Folds
 ------------------------------------------------------------------------------
 
+{-# INLINABLE drainN #-}
```
adithyaov: Add documentation!

src/Streamly/Internal/Prelude.hs
adithyaov: Add documentation!

adithyaov:
@harendra-kumar I think this can be merged. The only problem is the GHCJS build.

src/Streamly/Internal/Prelude.hs
```diff
@@ -783,12 +782,9 @@ iterate step = K.fromStream . go
...
 --
 -- /Since: 0.1.2/
+{-# INLINE_EARLY iterateM #-}
 iterateM :: (IsStream t, MonadAsync m) => (a -> m a) -> m a -> t m a
-iterateM step = go
-    where
-    go s = K.mkStream $ \st stp sng yld -> do
-        next <- s
-        K.foldStreamShared st stp sng yld (return next |: go (step next))
+iterateM step = fromStreamS . S.iterateM step
```
harendra-kumar: This is not concurrent any more. It will pick up only StreamD implementation which is not concurrent. We need to use the StreamK implementation and use a rewrite rule for the serial case, see repeatM implementation above.

src/Streamly/Internal/Prelude.hs
```diff
@@ -751,10 +751,9 @@ repeatMSerial = fromStreamS . S.repeatM
 -- @
 --
 -- @since 0.1.2
-iterate :: IsStream t => (a -> a) -> a -> t m a
-iterate step = K.fromStream . go
-    where
-    go s = K.cons s (go (step s))
+{-# INLINE_EARLY iterate #-}
```
harendra-kumar: A simple "INLINE" should be enough here, we do not need INLINE_EARLY as we do not have a rewrite rule here.

src/Streamly/Internal/Prelude.hs
harendra-kumar: We should perhaps have a test to check for concurrency so that we do not miss something like this in future.

PR #325, opened 2019-12-06
LastN combinator

adithyaov:
The `lastN` combinator.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -305,6 +305,10 @@ module Streamly.Internal.Data.Stream.StreamD
     , mkParallel
     , applyParallel
     , foldParallel
+
+    -- XXX Move this later
+    -- XXX Exported from Array again
+    , lastN
     )
 where
 
@@ -346,6 +350,7 @@ import Streamly.Internal.Data.Fold.Types (Fold(..))
 import Streamly.Internal.Data.Pipe.Types (Pipe(..), PipeState(..))
 import Streamly.Internal.Data.Time.Clock (Clock(Monotonic), getTime)
 import Streamly.Internal.Data.Unfold.Types (Unfold(..))
+import Streamly.Internal.Data.Strict (Tuple3'(..))
 
 import Streamly.Internal.Data.Stream.StreamD.Type
 import Streamly.Internal.Data.SVar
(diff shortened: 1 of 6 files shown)


src/Streamly/Streams/StreamD.hs
```diff
@@ -4237,3 +4241,22 @@ foldParallel f m = do
     sv <- newParallelVar StopNone defState
     toSVarParallel defState sv (toStreamD m)
     f $ fromStreamD $ fromSVar sv
+
+-- XXX This is taking too much time
+-- XXX This seems to fuse as far as I've checked
+-- | Take last 'n' elements from the stream and discard the rest.
+{-# INLINABLE lastN #-}
+lastN :: (Storable a, MonadIO m) => Int -> Fold m a (Array a)
```
harendra-kumar: 1) It will be nice to have a test for this combinator.
2) Please use 4 space indentation consistent with the rest of the library.

adithyaov:
@harendra-kumar I think this can be merged. The only problem is the GHCJS build.

test/Streamly/Test/Array.hs
```diff
@@ -132,26 +132,45 @@ testArraysOf =
                     $ IP.arraysOf 240
                     $ S.fromList list
                 assert (xs == list)
+
+lastN :: Int -> [a] -> [a]
+lastN n l | length l < n = l
```
harendra-kumar: This condition is unnecessary, just `drop (length l - n)` should be enough.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3844,3 +3849,21 @@ tapAsync f (Stream step1 state1) = Stream step TapInit
...
+lastN n = Fold step initial done
+    where
+        step (Tuple3' rb rh i) a = do
+            rh1 <- liftIO $ RB.unsafeInsert rb rh a
+            return $ Tuple3' rb rh1 (i + 1)
+        initial = fmap (\(a, b) -> Tuple3' a b 0) $ liftIO $ RB.new n 
+        done (Tuple3' rb rh i) = do
+            arr <- liftIO $ A.newArray n
+            foldFunc i rh snoc' arr rb
+        snoc' b a = liftIO $ A.unsafeSnoc b a
```
harendra-kumar: unsafeSnoc performs an array bound check, we do not need that here as we have already allocated an array of correct size, maybe that can help improve perf a bit.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3844,3 +3849,21 @@ tapAsync f (Stream step1 state1) = Stream step TapInit
             Yield a s -> Yield a (TapDone s)
             Skip s    -> Skip (TapDone s)
             Stop      -> Stop
+
+-- XXX Is there room for improvement?
```
harendra-kumar: If we make the ring buffer a first class data structure, we can just return the ring buffer instead of converting it into an array first. Consumers can directly consume the ring buffer.

harendra-kumar:
We should check `splitting/predicate` benchmarks in the `fileio` benchmark suite to  check for regressions because of the ring buffer touch ptr changes.

PR #331, opened 2019-12-11
Time related combinators

adithyaov:
The PR consists of `timelyTake` and `timelyDrop`.

TODO: Add tests and benchmarks

### DIFF
--- a/benchmark/Linear.hs
+++ b/benchmark/Linear.hs
@@ -356,19 +356,22 @@ main = do
         [ benchIOSink value "filter-even"     (Ops.filterEven 1)
         , benchIOSink value "filter-all-out"  (Ops.filterAllOut value 1)
         , benchIOSink value "filter-all-in"   (Ops.filterAllIn value 1)
+
         , benchIOSink value "take-all"        (Ops.takeAll value 1)
-        
-        -- XXX Compare this to takeAll
-        , benchIOSink value "takeByTime: All" (Ops.takeByTime (NanoSecond64 maxBound) 1)
-        , benchIOSink value "dropByTime: All" (Ops.dropByTime (NanoSecond64 maxBound) 1)
-        
+        , benchIOSink value "takeByTime-all"
+            (Ops.takeByTime (NanoSecond64 maxBound) 1)
         , benchIOSink value "takeWhile-true"  (Ops.takeWhileTrue value 1)
         --, benchIOSink value "takeWhileM-true" (Ops.takeWhileMTrue 1)
+
+        -- "drop-one" is dual to "last"
         , benchIOSink value "drop-one"        (Ops.dropOne 1)
         , benchIOSink value "drop-all"        (Ops.dropAll value 1)
+        , benchIOSink value "dropByTime-all"
+            (Ops.dropByTime (NanoSecond64 maxBound) 1)
         , benchIOSink value "dropWhile-true"  (Ops.dropWhileTrue value 1)
         --, benchIOSink value "dropWhileM-true" (Ops.dropWhileMTrue 1)
         , benchIOSink value "dropWhile-false" (Ops.dropWhileFalse value 1)
+
         , benchIOSink value "deleteBy" (Ops.deleteBy value 1)
         , benchIOSink value "intersperse" (Ops.intersperse value 1)
         , benchIOSink value "insertBy" (Ops.insertBy value 1)
--- a/benchmark/Linear.hs
+++ b/benchmark/Linear.hs
@@ -26,6 +26,7 @@ import qualified Streamly.Memory.Array as A
 import qualified Streamly.Prelude as S
 import qualified Streamly.Internal.Data.Sink as Sink
 
+import Streamly.Internal.Data.Time.Units
 import qualified Streamly.Internal.Memory.Array as IA
 import qualified Streamly.Internal.Data.Fold as IFL
 import qualified Streamly.Internal.Prelude as IP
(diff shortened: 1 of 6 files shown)


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3844,3 +3852,64 @@ tapAsync f (Stream step1 state1) = Stream step TapInit
...
+        (x, _) = properFraction (b * tenPow8)
+
+{-# INLINE_NORMAL timelyTake #-}
+timelyTake :: MonadIO m => Double -> Stream m a -> Stream m a
+timelyTake lim' (Stream step state) = Stream step' (Tuple' state Nothing)
+    where
+        lim = doubleToTimeSpec lim'
+        {-# INLINE_LATE step' #-}
+        step' gst (Tuple' s Nothing) = do
+            t <- liftIO $ getTime Monotonic
```
harendra-kumar: refactor as discussed.

test/Streamly/Test/Internal/Prelude.hs
```diff
@@ -0,0 +1,60 @@
...
+    where
+        tenPow8 = 10^(8 :: Int)
+        (a, b) = properFraction d
+        (x, _) = properFraction (b * tenPow8)
+
+testTimelyTake :: Property
+testTimelyTake =
+    forAll (choose (0, 20)) $ \(n :: Int) -> monadicIO $ do
+        let lim = 0.1 * (fromIntegral n :: Double)
+        let graceTime = 0.01 :: Double
```
harendra-kumar: Let's raise the grace time to 0.1 second, 0.01 second is too small.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3844,3 +3852,64 @@ tapAsync f (Stream step1 state1) = Stream step TapInit
...
+{-# INLINE doubleToTimeSpec #-}
+doubleToTimeSpec :: Double -> TimeSpec
+doubleToTimeSpec d = TimeSpec (a :: Int64) (x :: Int64)  
+    where
+        tenPow8 = 10^(8 :: Int)
+        (a, b) = properFraction d
+        (x, _) = properFraction (b * tenPow8)
+
+{-# INLINE_NORMAL timelyTake #-}
+timelyTake :: MonadIO m => Double -> Stream m a -> Stream m a
```
adithyaov: Rename to `takeByTime`

benchmark/Linear.hs
```diff
@@ -349,6 +349,8 @@ main =
         , benchIOSink "filter-all-out"  (Ops.filterAllOut 1)
         , benchIOSink "filter-all-in"   (Ops.filterAllIn 1)
         , benchIOSink "take-all"        (Ops.takeAll 1)
+        , benchIOSink "takeByTime: 10 sec" (Ops.takeByTime 10.0 1)
```
harendra-kumar: 1) 10 sec is very long, we keep benchmark times such that they complete quickly, this will hold up all benchmarks for too long. Usually more than 500 ms is too long because each benchmark also runs many samples and many iterations per sample. If the benchmark has to be long then we can put it under DEVBUILD or some such conditional.

2) Is this a benchmark? It does not seem to be measuring performance, it sounds more like a test rather than benchmark.

harendra-kumar: The benchmark should perhaps measure how many elements it can take in the given amount of time.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -4034,3 +4041,65 @@ lastN n = Fold step initial done
...
+    where
+        tenPow9 = 10^(9 :: Int)
+        (a, b) = properFraction d
+        x = round $ b * tenPow9
+
+-- XXX This does not fuse. Removing the if function results in fusion though.
+-- XXX This function will fuse with the use of fusion-plugin.
+-- XXX This function also fuses if compiled with -funfolding-use-threshold=500
+{-# INLINE_NORMAL takeByTime #-}
+takeByTime :: MonadIO m => Double -> Stream m a -> Stream m a
```
harendra-kumar: A `Double` seconds requires `TimeSpec` to hold it. Arithmetic on `TimeSpec` may be slow, this could be expensive since we have to do it in each iteration of the loop, for faster operations we may want to restrict the time argument to `TimeUnit64` which can do faster arithmetic.

However, currently all the cost is masked by the `getTime` call, so using efficient time arithmetic may not have any significant impact. But if we use a less costly time acquisition function (coarse granularity) then it may start to matter.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -4034,3 +4041,65 @@ lastN n = Fold step initial done
...
+takeByTime lim' (Stream step state) = Stream step' (state, Nothing)
+  where
+    lim = secondsToTimeSpec lim'
```
harendra-kumar: We can convert lim' to nanoseconds and then use `fromInteger` to convert nanoseconds to TimeSpec.

test/Streamly/Test/Internal/Prelude.hs
```diff
@@ -0,0 +1,60 @@
...
+import qualified Streamly.Internal.Prelude as SI
+
+import Streamly.Internal.Data.Time.Clock (Clock(Monotonic), getTime)
+import Streamly.Internal.Data.Time.Units
+import Data.Int (Int64)
+
+doubleToTimeSpec :: Double -> TimeSpec
+doubleToTimeSpec d = TimeSpec (a :: Int64) (x :: Int64)  
+    where
+        tenPow8 = 10^(8 :: Int)
```
harendra-kumar: 10^9

src/Streamly/Internal/Prelude.hs
```diff
@@ -4192,3 +4196,16 @@ usingStateT s f xs = evalStateT s $ f $ liftInner xs
 {-# INLINE runStateT #-}
 runStateT :: Monad m => s -> SerialT (StateT s m) a -> SerialT m (s, a)
 runStateT s xs = fromStreamD $ D.runStateT s (toStreamD xs)
+
+------------------------------------------------------------------------------
+-- Time related
+------------------------------------------------------------------------------
+
+{-# INLINE takeByTime #-}
+takeByTime :: (IsStream t, MonadIO m) => Double -> t m a -> t m a
```
harendra-kumar: Some documentation would be nice. We need to mention that the time check is done before generating the stream element, therefore, the total actual time may differ depending on how long it takes to generate the stream element.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: I think we make the signature `MonadIO m, TimeUnit64 t => ...` instead of using `Double`.
This is more clear and simple than converting time from `Double` (seconds). We can export it once we decide on `Units` for time.

PR #332, opened 2019-12-11
Move Benchmark* from src/ to benchmark/

adithyaov:
The modules have been specified to common benchmark properties in the cabal file. They could be moved to individual benchmarks though.

### DIFF
--- a/streamly.cabal
+++ b/streamly.cabal
@@ -557,11 +557,10 @@ test-suite parallel-loops
 
 benchmark linear
   import: bench-options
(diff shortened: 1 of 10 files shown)


streamly.cabal
```diff
@@ -257,6 +257,11 @@ common exe-options
 common bench-options
   import: compile-options, optimization-options
   ghc-options: -with-rtsopts "-T"
+  if flag(benchmark)
```
harendra-kumar: This flag can now be removed.

PR #334, opened 2019-12-12
Implement a stream sampling combinator

harendra-kumar:
Whenever the stream is evaluated it provides the newest available sample from the input stream which is strictly evaluated.


harendra-kumar:
Only the last commit belongs to this PR. the previous two commits are part of another yet unmerged PR.

pranaysashank:
@harendra-kumar What is the status of this? Should someone else pick this up?

PR #343, opened 2019-12-23
Implement <*> and >> for SerialT via StreamD.

pranaysashank:
This PR
- implements custom definitions for SerialT's `<*>`, `*>`,`>>` .
- Use a weaker constraint on many instance definitions in StreamD.

Url for files changed because github won't recompute the diff.
https://github.com/composewell/streamly/compare/master...ap

### DIFF
--- a/src/Streamly/Internal/Data/Stream/Serial.hs
+++ b/src/Streamly/Internal/Data/Stream/Serial.hs
@@ -76,7 +76,6 @@ import Streamly.Internal.Data.Stream.StreamK (IsStream(..), adapt, Stream, mkStr
 import qualified Streamly.Internal.Data.Stream.Prelude as P
 import qualified Streamly.Internal.Data.Stream.StreamK as K
 import qualified Streamly.Internal.Data.Stream.StreamD as D
-import qualified Streamly.Internal.Data.Unfold as UF
 
 #include "Instances.hs"
 #include "inline.hs"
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Data/Stream/Serial.hs
```diff
@@ -213,15 +214,19 @@ map f = mapM (return . f)
 
 {-# INLINE apSerial #-}
 apSerial :: Monad m => SerialT m (a -> b) -> SerialT m a -> SerialT m b
-apSerial (SerialT m1) (SerialT m2) =
-    let f x1 = D.concatMapU (UF.singleton (pure . x1)) (D.toStreamD m2)
-    in D.fromStreamD $ D.concatMap f (D.toStreamD m1)
+apSerial (SerialT m1) (SerialT m2) = D.fromStreamD $ D.toStreamD m1 <*> D.toStreamD m2
```
harendra-kumar: Use `D.concatAp` and `D.apSequence` directly instead of depending on the `Applicative` type class instance resolution?

src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -231,23 +232,61 @@ concatMap f = concatMapM (return . f)
...
+yield x = Stream (\_ s -> pure $ step undefined s) True
   where
     {-# INLINE_LATE step #-}
     step _ True  = Yield x False
     step _ False = Stop
 
-instance Monad m => Applicative (Stream m) where
+{-# INLINE_NORMAL concatAp #-}
+concatAp :: Functor f => Stream f (a -> b) -> Stream f a -> Stream f b
+concatAp (Stream stepa statea) (Stream stepb stateb) = Stream step' (Left statea)
```
harendra-kumar: Instead of stepa/statea I generally use step1/state1, that way it is more readable.

PR #345, opened 2019-12-25
Reorganize prelude benchmarks

harendra-kumar:
This is purely a reorganization of code by moving related code together and adding comments for subsections of code. There is no functionality change at all.

### DIFF
--- a/benchmark/LinearAsync.hs
+++ b/benchmark/LinearAsync.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 -- |
 -- Module      : Main
 -- Copyright   : (c) 2018 Harendra Kumar
@@ -31,6 +32,10 @@ benchSrcIO
 benchSrcIO t name f
     = bench name $ nfIO $ randomRIO (1,1) >>= Ops.toNull t . f
 
+{-# INLINE benchMonadicSrcIO #-}
+benchMonadicSrcIO :: String -> (Int -> IO ()) -> Benchmark
+benchMonadicSrcIO name f = bench name $ nfIO $ randomRIO (1,1) >>= f
+
 {-
 _benchId :: NFData b => String -> (Ops.Stream m Int -> Identity b) -> Benchmark
 _benchId name f = bench name $ nf (runIdentity . f) (Ops.source 10)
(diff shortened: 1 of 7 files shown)


benchmark/LinearAsync.hs
```diff
@@ -104,5 +127,9 @@ main =
         -- Parallel stages in a pipeline
         , benchIO "parAppMap" Ops.parAppMap
         , benchIO "parAppSum" Ops.parAppSum
+        , benchMonadicSrcIO "concatMapWith (2,x/2)"
+            (Ops.concatStreamsWith parallel 2 Ops.value)
```
adithyaov: It should be `Ops.value 'div' 2` acccording to the comment.

benchmark/NestedConcurrent.hs
```diff
@@ -0,0 +1,77 @@
...
+      , benchIO "filterAllOut"   $ Ops.filterAllOut   parallely
+      , benchIO "filterAllIn"    $ Ops.filterAllIn    parallely
+      , benchIO "filterSome"     $ Ops.filterSome     parallely
+      , benchIO "breakAfterSome" $ Ops.breakAfterSome parallely
+      ]
+
+    , bgroup "zipAsyncly"
+      [ benchIO "toNullAp"       $ Ops.toNullAp       zipAsyncly
+      ]
+    ]
```
adithyaov: I just glanced over this file as it seems to be a direct copy and paste. As long as it's compiling it should be fine.

streamly.cabal
```diff
@@ -249,9 +249,17 @@ common exe-options
   import: compile-options, optimization-options, threading-options
 
 -- Some benchmarks are threaded some are not
+-- XXX dependencies should be separated under bench-depends
 common bench-options
   import: compile-options, optimization-options
-  ghc-options: -with-rtsopts "-T"
+  ghc-options: -with-rtsopts "-T -M256M"
```
adithyaov: Although not required, it would me nice to document the problem that lead to the use of limited memory and mention it here or a small writeup would be sufficient.
Or just leave it be, I'll document it later.

adithyaov: Never mind this. I just noticed that you've written some of it in the commit message.

benchmark/LinearAsync.hs
harendra-kumar: Yes, changed, thanks!

PR #350, opened 2019-12-26
Add mkAsync_ to make a fold concurrent

harendra-kumar:


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -847,6 +854,43 @@ distribute :: Monad m => [Fold m a b] -> Fold m a [b]
...
+-- separate thread. The step function of the resulting fold just sends the
+-- input to the thread. The output of the fold is discarded. If an exception
+-- occurs in the original fold, the exception is thrown in the driving thread.
+-- Note that exception checking is poll based, we check if the fold has
+-- thrown any exceptions only at the point when we push an element to the fold.
+--
+-- /Internal/
+--
+{-# INLINE mkAsync_ #-}
+mkAsync_ :: MonadAsync m => Fold m a () -> m (Fold m a ())
```
pranaysashank: Change this signature to `Fold m a () -> Fold m a ()` maybe? By moving the outside monadic action to the initial step.

harendra-kumar: Yes, I think it can be done. It will be more convenient to use that way.

harendra-kumar:
This is replaced by `parEval` concurrent fold combinator in 0.9.0 .

PR #352, opened 2019-12-27
Add a mutable variable for Prim types

harendra-kumar:


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3230,36 +3240,36 @@ tapRate ::
...
                      throwM (MC.toException e))
-        loop countRef i
+        loop countVar i
 
     {-# INLINE_LATE step' #-}
     step' _ Nothing = do
-        countRef <- liftIO $ newIORef 0
-        tid <- fork $ loop countRef 0
+        countVar <- liftIO $ newVar 0
+        tid <- fork $ loop countVar 0
```
pranaysashank: Use `forkManaged` from the other PR?

harendra-kumar: we can remove `tapRate` itself later, it can be expressed in terms of those new combinators.

Issue #370, opened 2020-01-15
Finalization when the stream is not fully evaluated


harendra-kumar:
`after`, `finally`, and `bracket` combinators attach an action with a stream which is run if the stream stops normally or due to an exception. However, if the stream is lazily evaluated using a right fold and the fold terminates without fully evaluating the stream, the finalization action may never be invoked. This can cause confusing behavior.

One possible solution could be to bracket/finalize folds rather than bracketing/finalizing streams and unfolds.

harendra-kumar:
Thanks @hasufell for reporting this issue.

harendra-kumar:
Fixed by #372

pranaysashank:
Reopening this issue as these are only exposed from Internal modules in 0.7.1

harendra-kumar:
This is fixed in released code now.

PR #372, opened 2020-01-15
Add exception handling combinators with guaranteed cleanup

harendra-kumar:
The "after", "finally" and "bracket" combinators did not run the "cleanup"
handler in case the stream is lazily partially evaluated e.g. using the lazy
right fold (foldrM), "Streamly.Prelude.head" is an example of such a fold.

Since we run the cleanup action when the stream Stops, the action won't be run
if the stream is not fully drained.

In the new implementation, we use a GC hook to run the cleanup action in case
the stream got garbage collected even before finishing. This will take care of
the lazy right fold cases mentioned above.

### DIFF
--- a/benchmark/FileIO.hs
+++ b/benchmark/FileIO.hs
@@ -91,9 +91,15 @@ main = do
            , mkBench "catBracket" href $ do
                Handles inh _ <- readIORef href
                BFA.catBracket devNull inh
+           , mkBench "catBracketIO" href $ do
+               Handles inh _ <- readIORef href
+               BFA.catBracketIO devNull inh
            , mkBench "catBracketStream" href $ do
                Handles inh _ <- readIORef href
                BFA.catBracketStream devNull inh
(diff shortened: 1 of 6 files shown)


harendra-kumar:
See #370

shlok:
@harendra-kumar Did this fix also bring about proper cleanup in the case of async exceptions?

My rough guess/thoughts as an exceptions non-expert: Yes it did, unless the async exception occurs after the "before" action but before the `IORef` creation. (Could it be worth masking async exceptions at that precise point?)

shlok:
On the second thought, it doesn't seem so simple. An async exception could still kill the thread at other future points where the cleanup action would not be called.

I'm beginning to wonder if the most robust resource allocation methodology is to always do it completely outside of the stream. (Such seems seems to be the design of `Streamly.FileSystem.Handle`; where the `Handle` comes from and how it gets cleaned up is not its concern.)

PR #373, opened 2020-01-15
large stream size benchmarks

harendra-kumar:
* Fix `--stream-size` passing to child process for gauge `--measure-with` option to work correctly
* Support command line stream-size option for nested-unfold benchmark as well
* Support `--long` option in `bench.sh` to conveniently run large stream size (10 million elements) benchmarks before releasing.

### DIFF
--- a/benchmark/Chart.hs
+++ b/benchmark/Chart.hs
@@ -353,6 +353,11 @@ main = do
                             makeNestedGraphs
                             "charts/nested/results.csv"
                             "charts/nested"
+                NestedConcurrent -> benchShow opts cfg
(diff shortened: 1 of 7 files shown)


benchmark/Chart.hs
```diff
@@ -351,11 +353,11 @@ main = do
                             makeNestedGraphs
                             "charts/nested/results.csv"
                             "charts/nested"
-                NestedConcurrent -> benchShow opts cfg
-                            { title = Just "Nested concurrent loops 316 x 316 elems" }
+                NestedUnfold -> benchShow opts cfg
```
adithyaov: `NestedConcurrent` should not be removed.

PR #376, opened 2020-01-16
Group benchmarks based on memory and cpu costs

harendra-kumar:
So that we can run different benchmark groups with different heap size restrictions. Especially we want to enforce that streaming benchmarks run in constant memory.

See the commit messages for more details.

### DIFF
--- a/benchmark/Linear.hs
+++ b/benchmark/Linear.hs
@@ -143,17 +143,9 @@ main = do
         -- length is used to check for foldr/build fusion
         , benchPureSink value "length . IsList.toList" (length . GHC.toList)
         , benchPureSrc "IsString.fromString" (Ops.sourceIsString value)
-        , mkString value `deepseq` (bench "readsPrec pure streams" $
-                                nf Ops.readInstance (mkString value))
-        , mkString value `deepseq` (bench "readsPrec Haskell lists" $
-                                nf Ops.readInstanceList (mkListString value))
         , benchPureSink value "showsPrec pure streams" Ops.showInstance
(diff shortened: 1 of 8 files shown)


benchmark/Linear.hs
```diff
@@ -478,9 +446,63 @@ main = do
...
+    -- can run these conveniently with smaller stream size.
+    --
+    -- These are also the operations that programmers should be aware of and
+    -- should avoid using in a streaming application.
+    , bgroup "buffered"
+      [ -- Inherently non-streaming operations
+
+      -- Right folds for reducing are inherently non-streaming as the
+      -- expression needs to be fully built before it can be reduced.
+        benchIOSink value "foldrM/reduce/IO" Ops.foldrMReduce
```
adithyaov: `IO/reduce/foldrM` and friends?

benchmark/Parallel.hs
```diff
@@ -0,0 +1,102 @@
...
+    :: (t IO Int -> SerialT IO Int)
+    -> String
+    -> (Int -> t IO Int)
+    -> Benchmark
+benchSrcIO t name f
+    = bench name $ nfIO $ randomRIO (1,1) >>= Ops.toNull t . f
+
+{-# INLINE benchMonadicSrcIO #-}
+benchMonadicSrcIO :: String -> (Int -> IO ()) -> Benchmark
+benchMonadicSrcIO name f = bench name $ nfIO $ randomRIO (1,1) >>= f
```
adithyaov: This and `benchIONested` are equivalent. You can remove this I guess?

harendra-kumar: Fixed.

PR #379, opened 2020-01-20
Implement PrimArray

adithyaov:


streamly.cabal
```diff
@@ -285,6 +285,7 @@ library
                        Streamly.Prelude
                      , Streamly
                      , Streamly.Data.Array
+                     , Streamly.Data.PrimArray
```
harendra-kumar: Let's put it in other-modules for now. Give it some time before we expose it.

src/Streamly/Data/PrimArray.hs
```diff
@@ -0,0 +1,44 @@
...
+--
+module Streamly.Data.PrimArray
+    ( PrimArray
+    , Prim
+
+    -- * Construction
+    , A.fromListN
+    , A.fromList
+
+    -- Stream Folds
```
harendra-kumar: Let's have the same APIs exposed across Memory/Array and Prim/Array.

PR #381, opened 2020-01-22
Add distributeAsync_

harendra-kumar:
To tap a stream to multiple folds supplied in a container.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/Parallel.hs
+++ b/src/Streamly/Internal/Data/Stream/Parallel.hs
@@ -21,16 +21,25 @@
 --
 module Streamly.Internal.Data.Stream.Parallel
     (
+    -- * Parallel Stream Type
       ParallelT
     , Parallel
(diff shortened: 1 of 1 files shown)


src/Streamly/Internal/Data/Stream/Parallel.hs
```diff
@@ -386,6 +395,22 @@ tapAsync f m = mkStream $ \st yld sng stp -> do
     sv <- newFoldSVar st f
     foldStreamShared st yld sng stp (teeToSVar sv m)
 
+-- | Concurrently distribute a stream to a collection of fold functions,
+-- discarding the outputs of the folds.
+--
+-- >>> S.drain $ distributeAsync_ [S.mapM_ print, S.mapM_ print] (S.enumerateFromTo 1 2)
```
pranaysashank: We can have a better example maybe with a delay in the `Fold` or different folds in the container

PR #382, opened 2020-01-22
Add broad classification of combinators in doc

harendra-kumar:


harendra-kumar:
We have added notes about shapes where needed. closing this.

Issue #389, opened 2020-01-27
Fix space leak in fromStreamVar (GHC 8.8)


harendra-kumar:
Pranay's investigation suggests that GHC 8.8.x memory leak issue goes away if we remove this rewrite rule:

https://github.com/composewell/streamly/blob/753d091b75bd0d493b130a7eb66b283a949994e3/src/Streamly/Internal/Data/Stream/StreamK/Type.hs#L243

I ran some benchmarks on GHC 8.6.5 before and after removing the rule to see what is the impact of the rule. There seem to be mixed result, some benchmarks become worse while others improve.

The first run is with the rule, the next 2 runs are without the rule.

```
Generating reports for linear...
100,000 elems(time)
Benchmark                                               default(0)(μs) default(1) - default(0)(%) default(2) - default(0)(%) 
------------------------------------------------------- -------------- -------------------------- --------------------------
serially/concat-foldable/foldMapWith                           2819.35                     +62.08                     +63.00                     
serially/concat-foldable/foldMapM                              4798.66                     +36.52                     +38.70                      
serially/concat-foldable/foldMapWithM                          4926.03                     +34.89                     +33.86                    

serially/transformation/foldrSMap                              6226.66                    -114.17                    -117.40                     
```

These same three benchmarks i.e. `foldMapWith`, `foldMapM` and `foldMapWithM` seem to be affected similarly (~30%) in `linear-async` benchmark suite as well.

`nested` and `nested-concurrent` benchmarks seem to have not much impact.

harendra-kumar:
See this GHC ticket for background on this: https://gitlab.haskell.org/ghc/ghc/issues/17745

harendra-kumar:
The `linear-async` results do not seem to be predictable, sometimes we see regressions in the `foldMap*`  functions and at other times no regressions at all after removing the rule.

pranaysashank:
Of course, it's not strictly necessary to delete the rule we can have a differently named function that we can call so the rule doesn't affect it.

harendra-kumar:
Now that we know the cause of the issue, the new fix is to replace "mkStream" function by "MkStream" constructor.

PR #390, opened 2020-01-27
Disable rewrite rule to fix GHC 8.8.x space leak

harendra-kumar:
This rule leads to huge memory leak with GHC 8.8.x. See
https://gitlab.haskell.org/ghc/ghc/issues/17745 and
https://github.com/composewell/streamly/issues/389

This also does not seem to be particularly helpful in performance improvement
so disabling it altogether for now.


harendra-kumar:
We now have a better understanding of the GHC issue. The problem can occur even without the rewrite rule. To avoid the problem with existing compilers we can remove all polymorphism from `fromStreamVar`.

Is it possible to use inspection testing to find other places where this problem may occur?

PR #401, opened 2020-02-08
Add Streamly.FileSystem.DirStream.Posix

hasufell:
I don't see currently how to do this on windows. I also think not everything is reasonably abstractable between platforms.

### DIFF
--- a/src/Streamly/FileSystem/DirStream/Posix.hs
+++ b/src/Streamly/FileSystem/DirStream/Posix.hs
@@ -1,0 +1,53 @@
+{-# LANGUAGE CPP        #-}
+{-# LANGUAGE MultiWayIf #-}
+
+-- |
+-- Module      :  Streamly.FileSystem.DirStream.Posix
+-- Copyright   :  © 2020 Julian Ospald
+-- License     :  BSD3
+--
+-- Maintainer  :  streamly@composewell.com
+-- Stability   :  experimental
+-- Portability :  portable
+--
+-- This module provides working with directory streams (POSIX).
+module Streamly.FileSystem.DirStream.Posix
+  (
+  -- * Directory listing
+    unfoldDirContents
+  )
+where
+
+#include "inline.hs"
+
+import           Control.Monad.IO.Class         ( liftIO
+                                                , MonadIO
+                                                )
+import           Data.Word8
+import           Prelude                 hiding ( readFile )
+import           Streamly.Internal.Data.Unfold.Types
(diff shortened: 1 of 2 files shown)


harendra-kumar:
@hasufell thanks for the PR. We need directory streaming APIs so this is in the right direction. However, I would like to make sure that the approach is correct. We have plans to represent well-typed filesystem paths in streamly, perhaps in a module like `Streamly.FileSystem.Path` and even before that we need a lower level module something on the lines of the `filepath` package but preserving the system returned path wherever possible. In general, for low level path representation:

* We don't want to use bytestrings, we will use Arrays instead. So there should be no dependency on `bytestring` package.
* We won't depend on `posix-paths` package for the same reason and moreover we want Windows support as well.
* We would like to support Windows as well I think it may be possible with some extra effort and burden on maintenance. This is a fundamental thing and must be supported on Windows if we have to support Windows at all.
* In general, we want to keep the external dependencies to the minimum so we would also like to avoid `unix` and `word8` dependencies.

For typed paths we will model it on the `path` package to which I have contributed quite a bit and in general I like it but we may have some changes. Since we have an opportunity here with all IO modules being written afresh for streamly we can rethink the path representation as well.

We can have this evolving inside streamly package but it can be separated into another package once it matures.  I am writing a detailed spec on how to deal with paths in streamly so that we know we have thought about important issues before we commit anything, please hold on until then. It may take a couple of days.

hasufell:
> We have plans to represent well-typed filesystem paths in streamly, perhaps in a module like `Streamly.FileSystem.Path` and even before that we need a lower level module something on the lines of the `filepath` package but preserving the system returned path wherever possible.

Yes, all of that has been done in [hpath](https://github.com/hasufell/hpath).

>     * We don't want to use bytestrings, we will use Arrays instead. So there should be no dependency on `bytestring` package.

For filepaths? Arrays of Word8? Char? For which platform? Linux and Windows have different representations and cannot easily be merged. See https://gitlab.haskell.org/ghc/ghc/wikis/proposal/abstract-file-path

> For typed paths we will model it on the `path` package to which I have contributed quite a bit and in general I like it but we may have some changes.

`path` uses an incorrect representation and should not be used.

hasufell:
Another point why I recommend not using `path` is that it makes a distinction of File vs Directory, which is IMO a misdesign. It's impossible to know whether something is a file or directory just by looking at the path. And even if something is a directory (right now), it might be a file in a split second later.

Issue #406, opened 2020-02-10
Fix hlint errors


harendra-kumar:
1. Let's fix the ones that are reasonable to fix. 
2. For those which require too much work/effort we can annotate at file level/globally so that hlint does not complain 
3. Mark the hlint build not be allowed to fail so that we have a ratchet to not allow more failures.

SanchayanMaity:
@harendra-kumar Can I take this up? Will per directory clean up patches be a good idea?

harendra-kumar:
Yes please. You can organize the commits the way you think is reasonable. Do not keep the commits unmerged for too long, otherwise if code keeps changing you may have to do more work in merging.

SanchayanMaity:
@harendra-kumar Are there any specific benchmarks I should be running or should I run all after a change set? Ran into some issue with cabal build failure because of zlib (even though I have it installed) and changed `USE_STACK` in `bench.sh` to 1 manually to use stack.

harendra-kumar:
Running all of them will take too much time and there will be too much noise. We can identify which benchmarks to run based on specific eta-reduce changes. We can even run single benchmarks by running the benchmark executable via `cabal run` or manually and passing the benchmark name as argument.

pranaysashank:
We now have a [`.hlint.ignore`](https://github.com/composewell/streamly/blob/master/.hlint.ignore). It is a good idea to fix the hlint errors file by file from the ignore file.

SanchayanMaity:
@pranaysashank Ya, I have been doing the same.

PR #408, opened 2020-02-12
Hlint cleanup for Data, FileSystem and Memory directory

SanchayanMaity:
First round of cleanups for #406.


harendra-kumar:
@SanchayanMaity sometimes eta-reduce may affect performance. So it is a good idea to identify benchmarks that involve those operations and run those benchmarks before and after to make sure there is no perf impact.

harendra-kumar:
Merged. Thanks @SanchayanMaity !

Issue #414, opened 2020-02-16
Does not build with network >= 3.0.0.0 && < 3.1.0.0


hasufell:
https://github.com/composewell/streamly/blob/61eb4ebb8f785f4593244d22329846b519cee81d/src/Streamly/Internal/Network/Socket.hs#L252-L256

is wrong. Since 3.0.0.0 we have `fdSocket :: Socket -> IO CInt`: https://hackage.haskell.org/package/network-3.0.0.0/docs/Network-Socket.html#v:fdSocket

As such, it cannot find the instance:

```
src/Streamly/Internal/Network/Socket.hs:255:46: error:
    • No instance for (Integral (IO Foreign.C.Types.CInt))
        arising from a use of ‘fromIntegral’
    • In the second argument of ‘($)’, namely ‘fromIntegral fd’
      In the expression: threadWaitWrite $ fromIntegral fd
      In the second argument of ‘($)’, namely
        ‘let fd = fdSocket s in threadWaitWrite $ fromIntegral fd’
    |
255 |     let fd = fdSocket s in threadWaitWrite $ fromIntegral fd
    |                                              ^^^^^^^^^^^^^^^
```

Unfortunately, those versions don't have `withFdSocket`.

harendra-kumar:
So it would be something like this in those releases:

```
     fdSocket s >>= threadWaitWrite . fromIntegral
```

Would you like to fix this and send a PR? We are readying 0.7.1 for release and can include this in that.

PR #415, opened 2020-02-17
Fix build with network >= 3.0 && < 3.1 wrt #414

hasufell:


harendra-kumar:
Thanks for the PR @hasufell !

PR #418, opened 2020-02-19
Update credits directory for code picked up from primitive.

pranaysashank:


streamly.cabal
```diff
@@ -97,6 +97,7 @@ extra-source-files:
     Changelog.md
     credits/*.md
     credits/base-4.12.0.0.txt
+    credits/primitive-0.7.0.0.txt
```
harendra-kumar: Need to add this to `extra-source-files`.

Issue #255, opened 2019-08-11
Module wise benchmarks


harendra-kumar:
See #146 for module wise test organization. In the same way it may be a good idea to arrange benchmarks as well module wise. We can also have inspection tests along with the benchmarks, see #223. Also see #31.

harendra-kumar:
Some reorganization of benchmarks has been done in #417 . There is more that we need to do, I am listing some of the thoughts below:

1) Till now we used a  specified source stream or fold function to a generic benchmarking function. For example, `source` in the function below:

```
benchIOSink value name f = bench name $ nfIO $ randomRIO (1,1) >>= f . Ops.source value
```

and then we called it as
```
            , benchIOSink value "uncons" Ops.uncons
```

Going forward  for each benchmark we want to have an independent function defined with a `IO a` return type for each benchmark so that we can use the same function in `nfIO` as well as for inspection test. This is to facilitate inspection test for each benchmark. For example,

```
{-# INLINE interleave2 #-}
interleave2 :: Int -> Int -> IO ()
interleave2 value n = S.drain $ Internal.interleave
    (sourceUnfoldrMN (value `div` 2) n)
    (sourceUnfoldrMN (value `div` 2) (n + 1))

#ifdef INSPECTION
inspect $ hasNoTypeClasses 'interleave2
inspect $ 'interleave2 `hasNoType` ''D.InterleaveState
#endif
```

We don't have to do it right away for all benchmarks, but this is something to keep in mind for any new changes.

2) Since we have to write benchmarks for multiple stream types, we can write a common polymorphic function that works on all stream types and then specialize it to different stream types. The specialized functions can be used for inspection as well as benchmarking.

With these guidelines we can do the following:

1) See `benchmark/Streamly/Benchmark/Data/Stream/StreamD.hs` from PR #417 . We can export benchmark groups from Prelude.hs as well in the same way and use those benchmark groups in the top level benchmark driver files. 

2) We now have several benchmark targets. We have to make sure that all of these have corresponding support from `bench.sh` and `Chart.hs`.

3) Also consider if we can merge the `nested` benchmarks with Prelude benchmarks themselves.

4) Organize the Prelude benchmark code in the same way as the documentation in Prelude is organized. 

5) Currently signatures and INLINE pragmas are separated from the definitions. This does not work well for maintenance. We should keep signature/pragma and definitions for each benchmark together.

[...]

PR #431, opened 2020-02-26
Module wise benchmarks

adithyaov:
Solves a few points mentioned in #255 

* [x] See benchmark/Streamly/Benchmark/Data/Stream/StreamD.hs from PR #417 . We can export benchmark groups from Prelude.hs as well in the same way and use those benchmark groups in the top level benchmark driver files.
  * [x] Serial (previously Linear.hs, now Streamly/Benchmark/Prelude/Serial/*.hs)
  * [x] LinearAsync
  * [x] Parallel
  * [x] LinearRate


adithyaov:
@harendra-kumar Please review this PR when you find time. It would be easier if you review this PR commit by commit than everything at once.

PR #435, opened 2020-02-28
Fix bench.sh and charts for changes in the benchmarks

harendra-kumar:


harendra-kumar:
GHC 8.4 build is failing due to cabal crashing and ghcjs build is failing because cabal command failed due to ghc not found. I am unable to reproduce these locally, so merging this for now.

PR #438, opened 2020-03-02
Keep signature/pragma and definitions for each benchmark together

adithyaov:
Solves a few points mentioned in #255 

* [x] Currently signatures and INLINE pragmas are separated from the definitions. This does not work well for maintenance. We should keep signature/pragma and definitions for each benchmark together.


adithyaov:
This can be merged now.

PR #442, opened 2020-03-02
Split Network.Socket module, create Network.Socket.IO

harendra-kumar:
The non-streaming array IO APIs go into the IO module. More IO based
non-streaming Socket APIs can later be exported from this module.


harendra-kumar:
Postponing this as there are very few useful non-streaming (IO) APIs, keeping them in the streaming module for now. Will think about this later.

PR #445, opened 2020-03-04
Prefer Windows APIs over POSIX in Windows systems

adithyaov:


harendra-kumar:
There is no point of bumping the version in configure.ac now, as we are not going to release with this version number. We need to bump it again before the next release.

PR #448, opened 2020-03-08
Mask async exceptions in the "bracketIO" functions.

shlok:
Without this change, there are points (between "bef" and registration of "aft") at which async exceptions can cause resource leaks.

A similar thing is done here: https://github.com/snoyberg/conduit/blob/resourcet-1.2.3/resourcet/Control/Monad/Trans/Resource.hs#L119

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -2951,6 +2951,10 @@ gbracketIO bef exc aft fexc fnormal =
     -- weak pointer to us.
     {-# INLINE_LATE step #-}
     step _ GBracketIOInit = do
+        -- We mask asynchronous exceptions to make the execution
+        -- of 'bef' and the registration of 'aft' atomic.
+        -- A similar thing is done in the resourcet package: https://git.io/JvKV3
+        -- Tutorial: https://markkarpov.com/tutorial/exceptions.html
         (r, ref) <- liftBaseOp_ mask_ $ do
             r <- bef
             ref <- newFinalizedIORef (aft r)
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -320,14 +320,14 @@ where
 
 import Control.Concurrent (killThread, myThreadId, takeMVar, threadDelay)
 import Control.Exception
-       (Exception, SomeException, AsyncException, fromException)
+       (Exception, SomeException, AsyncException, fromException, mask_)
 import Control.Monad (void, when, forever)
 import Control.Monad.Catch (MonadCatch, throwM)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Reader (ReaderT)
 import Control.Monad.State.Strict (StateT)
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -2951,8 +2951,10 @@ gbracketIO bef exc aft fexc fnormal =
     -- weak pointer to us.
     {-# INLINE_LATE step #-}
     step _ GBracketIOInit = do
-        r <- bef
-        ref <- newFinalizedIORef (aft r)
+        (r, ref) <- liftBaseOp_ mask_ $ do
+            r <- bef
+            ref <- newFinalizedIORef (aft r)
```
adithyaov: It would be nice if you could write up small developer documentation regarding the change or paste a few references right above the code.

shlok: Thanks for looking into this. I just pushed a few comments.

harendra-kumar:
Thanks @shlok ! The PR looks good.

It is a good idea to run the bracket benchmarks in the `fileio` benchmark suite. I don't expect any regression but its better to run the benchmarks and check. The relevant benchmarks are `readArray/catBracket*` and `readStream/catBracket*`. Note that readStream benchmarks are available only under the `dev` build flag and you will need to have a largish text file (100 MB or more) to run those. You can run the benchmarks before and after the change and see the difference. 

If you have got time, its a good opportunity to add some tests for `bracketIO` and friends in another PR.

shlok:
Thanks for the feedback, @harendra-kumar. I performed the following steps to do the benchmarks:

* Downloaded the 100 MB `enwik8` text file from http://mattmahoney.net/dc/textdata.html and put it into `benchmark/text-processing/gutenberg-500.txt`
* Made sure `charts/` is empty.
* `git checkout 62bb140` (`master`).
* Ran the following 5 times: `./bench.sh --append --benchmarks fileio --cabal-build-flags '-f dev' -- readArray/catBracket readStream/catBracket`. This gives me a `master` baseline plus 4 additional runs of `master`.
* `git checkout 44f2c57` (this `mask-async` PR branch).
* Ran the same `./bench.sh` command as above 4 more times.
* Generated the charts: `./bench.sh --no-measure --graphs --benchmarks fileio`.

The resulting charts can be seen below. Note again that each chart shows a `master` baseline, followed by 4 more `master` runs, followed by 4 runs for this PR.

Although there is quite some variability between runs, I don’t  see any consistent performance differences.

(Benchmarking  machine: MacBook Pro (15-inch, 2016), 2.6 GHz Intel Core i7, 16 GB 2133 MHz LPDDR3, macOS Mojave 10.14.6.)

fileIO-median-time svg
fileIO-median-cpuTime svg
fileIO-median-maxrss svg

harendra-kumar:
Thanks! @shlok 

To confirm, I just ran it on a 500MB file:

```
File IO(cpuTime)
Benchmark                    default(ms)
---------------------------- -----------
readArray/catBracketIO            110.76
readArray/catBracketStream        110.49
readArray/catBracket              109.60
readArray/catBracketStreamIO      109.38
```

IO functions (with added code) perform the same as non-IO (they do not have the new code) ones, this is a proof that the addition of new code did not make any difference.

We are good to go.

PR #449, opened 2020-03-14
Implement stream parsing

harendra-kumar:
This is the initial version of stream parsing. It implements a "Parse"
type, some parsers based on that type, applicative composition, operations to
run parsers (parse) and to run parsers over chunks of a stream (parseChunks).

Parsers are just an extension of Folds in Streamly.Data.Fold. Parsers just add
backtracking and failure capabilities to folds.

Operations like splitOn to split a stream on a predicate can now be expressed
using a parser applied repeatedly on a stream. For example, line splitting can
be done using parsers. Parsers are as fast as fastest custom code for
line splitting.


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -942,6 +948,187 @@ foldlS fstep begin (Stream step state) = Stream step' (Left (state, begin))
...
+    stepOuter _ (ParseChunksBuf [] s buf pst) =
+        return $ Skip $ ParseChunksStream s buf pst
+
+    -- buffered processing loop
+    stepOuter _ (ParseChunksBuf (x:xs) s buf pst) = do
+        pRes <- pstep pst x
+        -- XXX GHC/fusion issue: if we remove any one case from this case
+        -- statement then everything fuses. Fusion plugin is unable to fuse
+        -- this, but -funfolding-use-threshold=100 fuses it.
+        case pRes of
```
harendra-kumar: @pranaysashank this is the fusion issue I was talking about earlier.

harendra-kumar:
A potential simplification of `Step` is to have only two tags:

```
data Step s = Move Int s | Halt s
```

The `Move` state with a +ve value would mean to release the tail of the buffer by that much and a -ve value would mean to go back in the buffer by that much, a 0 value would mean what currently `Hold` is i.e. add the current item to buffer.

In the current design `Keep n` makes sure there won't be a space leak even if the parser (programmer) specified a larger `n` than required. However, with `Move` if the programmer specified a lower value than actual then we may continuously leak space on each step.

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -0,0 +1,254 @@
...
+    initial = return False
+    step s a = return $
+        if s
+        then Halt True
+        else
+            if predicate a
+            then Halt True
+            else Keep 0 False
+
+-- >>> S.parse (PR.any (== 0)) $ S.fromList [1,0,1]
```
adithyaov: `PR.all` instead of `PR.any`.

src/Streamly/Internal/Data/Parse/Types.hs
```diff
@@ -0,0 +1,169 @@
...
+-- the fold. The driver invokes @step@ with the previous state and an input
+-- element. It results into a new state and a command to the driver represented
+-- by 'Step' type. At any point of time the driver can call @extract@ to
+-- inspect the result of the fold. It may result in an error or an output value
+-- along with the count (zero or more) of last unused elements.
+--
+data Parse m a b =
+    forall s. Parse (s -> a -> m (Step s))
+                    (m s)
+                    (s -> m (Either String (Int, b)))
```
adithyaov: `Either String (Int, b)` does not seem right to me. How will you replay the input if you have an alternative fold? You need to include more formation if the parser fails.

Say, I want something like `takeExact 4 <|> takeExact 3`.

adithyaov: I think I might be in the wrong here. Please don't reply to this yet.

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -0,0 +1,254 @@
...
+
+    where
+
+    initial = (Tuple' 0) <$> finitial
+
+    step (Tuple' i r) a = do
+        res <- fstep r a
+        let i1 = i + 1
+            s1 = Tuple' i1 res
+        return $ Hold s1
```
adithyaov: This `step` makes more sense in my opinion. 
```
    step (Tuple' i r) a = do
        res <- fstep r a
        let i1 = i + 1
            s1 = Tuple' i1 res
        if i1 < n
        then return $ Hold s1
        else return $ Keep 0 s1
```
I'm also curious how `takeAtleast 5 <|> takeExact 10` would work with different implementations of `step`.

adithyaov: In my opinion alternative instance should be as greedy as possible.

src/Streamly/Internal/Data/Parse/Types.hs
adithyaov: Please ignore this.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -830,10 +832,41 @@ or = Fold (\x a -> return $ x || a) (return False) return
 -- Grouping without looking at elements
 ------------------------------------------------------------------------------
 --
+-- XXX Once we have terminating folds, this Parse should get replaced by Fold.
```
adithyaov: `Fold` is a terminating fold?

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -830,10 +832,41 @@ or = Fold (\x a -> return $ x || a) (return False) return
...
+-- | Take at most @n@ input elements and fold them using the supplied fold.
+--
+-- Stops after @n@ elements.
+-- Never fails.
+--
+-- /Internal/
+--
+{-# INLINE take #-}
+take :: Monad m => Int -> Fold m a b -> Parse m a b
+take n (Fold fstep finitial fextract) = Parse step initial extract
```
adithyaov: Move this to `Parse.hs`?

benchmark/lib/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -208,3 +210,18 @@ o_n_heap_serial_folds value =
                 ]
           ]
     ]
+
+o_1_space_serial_parse :: Int -> [Benchmark]
+o_1_space_serial_parse value =
```
adithyaov: Not so important, but you could move this to `benchmark/lib/Streamly/Benchmark/Data/Parse.hs`.

src/Streamly/Internal/Data/Parse/Types.hs
```diff
@@ -0,0 +1,169 @@
...
+-- keep the input for future backtracking or drop it. The constructors of
+-- 'Step' represent the commands to the driver.
+--
+-- /Internal/
+--
+data Step s =
+      Hold s     -- ^ Hold the current input in the buffer, the input is
+                 -- accepted tentatively, but the fold may later fail with the
+                 -- input remaining unused.
+    | Keep Int s -- ^ Keep @n@ most recent inputs, including the input of this
```
adithyaov: Can you replace this with `Discard s`?
Can you give me an example where we would use Keep i s where i > 0?

adithyaov: Turns out `Keep i s` is indeed needed.
Example: A parser that succeeds if atleast the last `n` inputs `(n >= 0 && n < i)` satisfy a certain condition. Such a parser would need to keep history.

adithyaov:
I prefer the current implementation of `Step` instead of `data Step s = Move Int s | Halt s`

src/Streamly/Internal/Data/Parse/Types.hs
adithyaov: > Example: A parser that succeeds if at least the last n inputs (n >= 0 && n < i) satisfy a certain condition. Such a parser would need to keep history even after success.

This can be done while updating the state accordingly though.

harendra-kumar:
I am considering a few more changes. 

1) Earlier I thought I should use constructor names that are different from stream constructors so that we can import them both in the same module unqualified. But I am now considering that it is more important to convey the correspondence of stream constructors with the fold constructors. So I am considering to change the naming as follows:

```
Halt => Stop
Hold => Skip
```

Another thing to note is that `Hold` is just a synonym for `Back 0` and in the same way `Keep 0` can have a synonym called `Yield` which corresponds to the `Yield` constructor in streams. So streams/unfolds and folds are dual and they have precise correspondence with each other.

2) Also, change the name "Parse" to "Parser" to make it consistent with other libraries. There is  another reason for this, I am considering a dual to parsers called a "Generator" type where we can move through or seek in the source buffer in a stream.

3) The existing `Fold` type can be extended to include the `Stop` and `Yield` (i.e. the synonym for `Keep 0`) constructors for the Step type. This can make the existing folds more powerful and the `Parser` type would just extend it to make it even more powerful by adding backtracking.

adithyaov:
@harendra-kumar `Halt` and `Hold` make sense to me. Instead of removing/changing them, it would be nice if you could provide type synonyms.
```
type Stop = Halt
type Skip s = Hold s
type Yield s = Keep 0 s
```

harendra-kumar:
There does not seem to be any difference in Halt and Stop. For folds there won't be a `Keep n s`, so there is no point of synonym, there will only be a `Yield s`. `Keep` will only be in parsers and it will remain with the same name. 

Now only Skip vs Hold remains. If we think it from the point of view of the fold that it consumed an input but did not produce a new output, Skip makes sense. `Hold` only makes sense from the point of view of the buffer. Skip indirectly means Hold, since you are not producing any output you are holding the input.

Having two names for each constructor would create a big confusion, we have to decide this way or that way and use that. Are we going to use whichever name we feel like in a particular combinator? That will be a disaster.

benchmark/lib/Streamly/Benchmark/Data/Fold.hs
harendra-kumar: Done.

src/Streamly/Internal/Data/Parse.hs
harendra-kumar: done.

src/Streamly/Internal/Data/Parse.hs
harendra-kumar: Done.

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -0,0 +1,258 @@
...
+--
+{-# INLINABLE takeAtLeast #-}
+takeAtLeast :: Monad m => Int -> Fold m a b -> Parser m a b
+takeAtLeast n (Fold fstep finitial fextract) = Parser step initial extract
+
+    where
+
+    initial = (Tuple' 0) <$> finitial
+
+    step (Tuple' i r) a = do
```
pranaysashank: Perhaps `i` should be strict, though it'll be in NF anyway due to the comparison later.

src/Streamly/Internal/Data/Parse/Types.hs
```diff
@@ -0,0 +1,170 @@
...
+                Skip n s -> Skip n (SeqParseR f s)
+                Stop n s -> Stop n (SeqParseR f s)
+
+        step (SeqParseLErr _) _ = error "step called in SeqParseLErr"
+
+        -- XXX bimap to add "<*>: Right parse failed" to the error message
+        extract (SeqParseR f s) = fmap (fmap f) (extractR s)
+        extract (SeqParseL _) = return $ Left "<*>: Incomplete left parse"
+        extract (SeqParseLErr err) =
+            return $ Left $ "<*>: Left parse failed\n" ++ err
```
pranaysashank: Is `Parser m a` not a Monad?

src/Streamly/Internal/Data/Parse/Types.hs
```diff
@@ -0,0 +1,170 @@
...
+-- to the driver represented by 'Step' type. At any point of time the driver
+-- can call @extract@ to inspect the result of the fold. It may result in an
+-- error or an output value.
+--
+data Parser m a b =
+    forall s. Parser (s -> a -> m (Step s)) (m s) (s -> m (Either String b))
+
+-- | Maps a function over the output of the fold.
+--
+instance Monad m => Functor (Parser m a) where
```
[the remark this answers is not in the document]
harendra-kumar: Only if/when we replace the definition of `<$`.

src/Streamly/Internal/Data/Parse/Types.hs
```diff
@@ -0,0 +1,170 @@
...
+    {-# INLINE fmap #-}
+    fmap f (Parser step initial extract) =
+        Parser step initial (fmap3 f extract)
+
+        where
+
+        fmap3 g = fmap (fmap (fmap g))
+
+    {-# INLINE (<$) #-}
+    (<$) b = \_ -> pure b
```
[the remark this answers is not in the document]
harendra-kumar: Good catch. It seems okay for pure parser but not okay for effectful ones. We can just remove this and use the default definition which would be correct. We need to remove the same from `Streamly.Data.Fold` as well. Also, we need to remove the incorrect definitions of `<*` and `*>` from `Streamly.Data.Fold`.

src/Streamly/Internal/Data/Parse/Types.hs
harendra-kumar: Not sure how a `Monad` can be written/useful, a `Comonad` is more likely to be useful.

PR #451, opened 2020-03-19
Implement distribute and zip (zipWith) for parsers

harendra-kumar:


src/Streamly/Internal/Data/Parse.hs
```diff
@@ -80,6 +88,23 @@ all predicate = Parse step initial (\x -> return $ Right (0,x))
             else Halt False
         else Halt False
 
+{-# INLINE sum #-}
+sum :: (Monad m, Num a) => Parse m a a
+sum = Parse (\s a -> return $ Keep 0 (s + a))   -- step
```
adithyaov: There is a `sum` in `Fold`. You can remove this as we have `fromFold`.

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -252,3 +277,101 @@ endBefore = undefined
...
 finishBy :: Parse m a x -> Fold m a y -> Parse m a (x, y)
 finishBy = undefined
+
+-------------------------------------------------------------------------------
+-- Zipping Parses
+-------------------------------------------------------------------------------
+--
+
+-- | State of the pair of parsers in a zip composition
+-- Note: strictness annotation is important for fusing the constructors
```
adithyaov: Is it the fusion of constructors or constructor elimination via specialization?

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -252,3 +277,101 @@ endBefore = undefined
...
+
+    {-# INLINE_LATE initial #-}
+    initial = do
+        sL <- initialL
+        sR <- initialR
+        return $ ParPair ([], Right sL, [], []) ([], Right sR, [], [])
+
+    {-# INLINE_LATE step #-}
+    step (ParPair (bufL, Right sL, inpL1, inpL2)
+                  (bufR, Right sR, inpR1, inpR2)) x = do
```
adithyaov: What is `inpL1` and `inpL2` here?

adithyaov: Won't `inpR2` and `inpL2` always be `[]`?

adithyaov: I believe you are using these for efficiency? Similar to how d-list is implemented for efficient `snoc` and `reverse`?

src/Streamly/Internal/Data/Parse.hs
```diff
@@ -252,3 +277,101 @@ endBefore = undefined
...
+    {-# INLINE extractBoth #-}
+    extractBoth sL sR = do
+        rL <- extractL sL
+        case rL of
+            Left err -> return $ Left err
+            Right (_, x) -> do
+                rR <- extractR sR
+                return $ case rR of
+                    Left err -> Left err
+                    Right (_, y) -> Right (0, zf x y)
```
adithyaov: Is `Right (0 ...` correct?
Shouldn't it be more like `Right (min (length bufL1 + length bufL2) (length bufR1 + length bufR2)) ...`

harendra-kumar:
I will push this again, there are several changes to the previous code.

src/Streamly/Internal/Data/Parse.hs
adithyaov: Cool, Cool!

PR #459, opened 2020-03-24
Make individual function for each benchmark of Data.Fold.

adithyaov:


harendra-kumar:
Its good to make individual functions, that will help in writing inspections tests if any. But more than that I wanted to:

1. Move this file from "benchmark/lib/..." to "benchmark/Streamly/..." namespace.
2. Have an individual benchmark executable for folds like we now have for parsers, i.e. separate it from "linear". In the same way we did for parsers.
3.  Move the fold benchmarks under top level "o-1-space/, o-n-space/,..." benchmark categories. Later when we run these through bench.sh we can pass the required heap and stack memory restriction flags to the executable when running it based on the top level category we are running. The motivation for this is to scheme is to reduce the number of executables we have to create in "streamly-benchmarks.cabal", and still provide separate executables for different modules.

adithyaov:
> Move the fold benchmarks under top level "o-1-space/, o-n-space"

Do you mean something like `benchmark/o-1-space/Streamly/Data/Fold.hs`?

Although we could do this, `o-n-heap` in `Fold.hs` has only 1 benchmark group. Wouldn't this be overkill in such cases? I would prefer it anyway. Please let me know your thoughts on this.

adithyaov:
This PR is rather unnecessary, will split when required.

harendra-kumar:
This was good change, we will be able to write inspection tests for individual fold functions with this refactoring. why would you not do it? We can merge it right away.

adithyaov:
With the way the functions are split, it is not possible to write inspection tests. I'll move this task to #462 with appropriate changes.

PR #462, opened 2020-03-26
Make a sandalone fold executable.

adithyaov:


adithyaov:
@harendra-kumar Please merge this for now.

benchmark/streamly-benchmarks.cabal
```diff
@@ -226,6 +225,18 @@ benchmark serial-o-n-space
     build-depends: lib-prelude
     buildable: True
 
+-- XXX Default memory restriction = o-1-space
+benchmark fold
+  import: bench-options
+  type: exitcode-stdio-1.0
+  ghc-options: -with-rtsopts "-T -K36K -M16M"
```
harendra-kumar: is 16M heap enough for o-n-heap fold benchmarks?

adithyaov: I think we should remove ghc-options all togather. We'll control that via `bench.sh`.

Issue #466, opened 2020-04-04
Build fails with primitive package 0.6.3.0


pranaysashank:
That is the primitive version provided on stackage lts-12.1, since we support ghc-8.4.3 we should fix this. The problem seem to be in the duplicated code in `Internal.Data.Prim.Array.Types`.

@saurabhnanda brought this to notice on gitter.

adithyaov:
`primitive-0.6.3.0` is deprecated. I think the better option is to maybe update the `primitive` version on stackage ecosystem?

pranaysashank:
I don’t think that’s realistic or possible

adithyaov:
Do you mean `lts-12.1` is locked? Can't the package versions be updated? If so, I think a simple solution is to add `primitive` to `extra-deps`.

adithyaov:
Build fails with primitive package < 0.6.4. I don't think deprication is the problem.

pranaysashank:
Can't close until there's a release.

pranaysashank:
streamly-0.7.2 is now released.

PR #469, opened 2020-04-05
Fix build for primitive-0.6.3.0. Solves #466.

adithyaov:


harendra-kumar:
When adding this module I asked Adithya to keep only the bare minimum that we need but he argued that there is no harm in keeping everything. Unwanted code is always a maintenance headache plus added compilation times. We should try to keep the bare minimum needed or if we expect something to be used in future.

adithyaov:
> This entire module is a mess that ideally we shouldn't have to maintain.

I agree. At that point, it seemed like a good idea because I thought I'll use these functions in `Streamly.External.Text`. I think it's better to keep it at bare minimum like @harendra-kumar suggested.

adithyaov:
Hopefully users of streamly did not use that part of the code. Removing unecessary code also got rid of the CPP macros which is nice.

harendra-kumar:
Looks good. @pranaysashank please merge if it looks fine to you as well.

harendra-kumar:
Can we add a changelog entry about the build fix?

pranaysashank:
Please fix the failing builds

PR #470, opened 2020-04-06
Benchmark Reorganization

### LINKED ISSUE #468: Move benchmark/Array* files to benchmark/Streamly/Benchmark/Memory/*

adithyaov:
Closes #468


adithyaov:
We can potentially move nested unfolds out of lib. I think it's better to add an appropriate bench group regarding the mem restriction as well in the files that have been renamed.

benchmark/Streamly/Benchmark/Data/NestedUnfoldOps.hs
```diff
@@ -5,7 +5,7 @@
 -- License     : BSD3
 -- Maintainer  : streamly@composewell.com
 
-module NestedUnfoldOps where
+module Streamly.Benchmark.Data.NestedUnfoldOps where
```
harendra-kumar, line 8: 1. Rename this to `Data.Unfold` to reflect the path of the module in the library.
2. Move the unfold benchmarks from Prelude here.

benchmark/streamly-benchmarks.cabal
```diff
@@ -324,7 +324,7 @@ benchmark linear-async
   import: bench-options-threaded
   type: exitcode-stdio-1.0
   ghc-options: -with-rtsopts "-T -N2 -K64K -M16M"
-  hs-source-dirs: .
+  hs-source-dirs: ., Streamly/Benchmark/Prelude
```
harendra-kumar, line 327: is "." still needed here?

adithyaov: You're right, I'll be removing it.

benchmark/Streamly/Benchmark/Data/NestedUnfoldOps.hs
adithyaov: Yes, I've done it already, didn't push it yet.

benchmark/Chart.hs
```diff
@@ -423,3 +427,13 @@ main = do
                             (makeGraphs "fold-o-n-heap")
                             "charts/fold-o-n-heap/results.csv"
                             "charts/fold-o-n-heap"
+                UnfoldO1Space -> benchShow opts cfg
```
harendra-kumar, line 430: This is all mechanically repeated code, at some point we have to do something about it, maybe use TH or some other way to remove duplication.

PR #471, opened 2020-04-06
Add coverage build to travis.

pranaysashank:


.travis.yml
```diff
@@ -61,8 +61,8 @@ env:
   - PACKCHECK_LOCAL_PATH="./packcheck.sh"
   # If you have not committed packcheck.sh in your repo at PACKCHECK_LOCAL_PATH
   # then it is automatically pulled from this URL.
-  - PACKCHECK_GITHUB_URL="https://raw.githubusercontent.com/harendra-kumar/packcheck"
-  - PACKCHECK_GITHUB_COMMIT="b8d39c9cf336e4627e12eb4b1f5827bb7b6afc91"
+  - PACKCHECK_GITHUB_URL="https://raw.githubusercontent.com/pranaysashank/packcheck"
```
harendra-kumar, line 64: URL should be changed to `composewell/packcheck`.

.circleci/config.yml
```diff
@@ -208,6 +208,28 @@ jobs:
...
+      <<: *env
+      steps:
+        - checkout
+        - *preinstall
+        - *restore
+        - run:
+            environment:
+              BUILD: cabal-v2
+              COVERALLS_OPTIONS: "--repo-token=$REPO_TOKEN --coverage-mode=StrictlyFullLines --exclude-dir=test"
+              CABAL_GLOBAL_INSTALL_OPTIONS: " --project-file=cabal.project.coveralls"
```
harendra-kumar, line 221: We have a `CABAL_PROJECT` environment variable in packcheck to specify the project file. Won't that work here?

pranaysashank: I can use it here, because cabal doesn't work properly I had to essentially do it anyway.

pranaysashank:
Gonna merge this, although two things:
1) Something is wrong with circle ci, tests take too long for some reason and the build stops sometimes perhaps we can extend the allowed time with no output.
2) git seems to be in a detached state and coveralls is receiving the current branch name as "HEAD"

harendra-kumar:
Branch name being received as "HEAD" may be the reason why it was not being updated. I was suspecting something like this. Because coveralls was correctly showing data for all other branches.

PR #472, opened 2020-04-07
Multiple changes corresponding to the fold benchmark.

adithyaov:
1. Removed rts options at compile time for fold benchmark.
2. Added fold-o-1-space and fold-o-1-heap benchmark to bench.sh
3. Added fold-o-1-space and fold-o-1-heap to bench.sh


harendra-kumar:
Looks good! Thanks!

PR #478, opened 2020-04-09
Change the return type of "extract" and implement "teeWith"

harendra-kumar:
See the commit messages for details. Its better to review each commit individually.

### DIFF
--- a/src/Streamly/Internal/Data/Parser.hs
+++ b/src/Streamly/Internal/Data/Parser.hs
@@ -1,7 +1,5 @@
-{-# LANGUAGE BangPatterns              #-}
 {-# LANGUAGE CPP                       #-}
 {-# LANGUAGE RankNTypes #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 
 -- |
@@ -44,8 +42,8 @@ where
 import Prelude
        hiding (any, all, takeWhile)
 
-import Streamly.Internal.Data.Parser.Types (Parser(..), Step(..))
 import Streamly.Internal.Data.Fold.Types (Fold(..))
+import Streamly.Internal.Data.Parser.Types (Parser(..), Step(..))
 
 import Streamly.Internal.Data.Strict
 
@@ -57,17 +55,14 @@ import Streamly.Internal.Data.Strict
 --
 {-# INLINE fromFold #-}
 fromFold :: Monad m => Fold m a b -> Parser m a b
-fromFold (Fold fstep finitial fextract) = Parser step initial extract
+fromFold (Fold fstep finitial fextract) = Parser step finitial fextract
+
     where
 
-    initial = finitial
     step s a = Yield 0 <$> fstep s a
-    extract s = do
-        r <- fextract s
-        return $ Right r
 
 -------------------------------------------------------------------------------
--- Terminating folds
+-- Terminating but not failing folds
 -------------------------------------------------------------------------------
 --
 -- |
@@ -76,9 +71,12 @@ fromFold (Fold fstep finitial fextract) = Parser step initial extract
 --
 {-# INLINABLE any #-}
 any :: Monad m => (a -> Bool) -> Parser m a Bool
-any predicate = Parser step initial (return . Right)
+any predicate = Parser step initial return
+
     where
+
     initial = return False
+
     step s a = return $
         if s
         then Stop 0 True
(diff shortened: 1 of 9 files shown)


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1077,22 +1077,17 @@ splitParse (Parser pstep initial extract) (Stream step state) =
...
+                            ParseChunksYield b (ParseChunksInit src s)
             Skip s -> return $ Skip $ ParseChunksStream s buf pst
             Stop   -> do
-                res <- extract pst
-                case res of
-                    Left err -> error err
-                    Right a -> do
-                        let src = Prelude.reverse buf
-                        return $ Skip $ ParseChunksYield a (ParseChunksInitLeftOver src)
+                b <- extract pst
```
adithyaov, line 1087:
```
Stop   -> do
    b <- extract pst
    let src = Prelude.reverse buf
    return $ Skip $
        ParseChunksYield b (ParseChunksInitLeftOver src)
```
The extract at the end seems inconsistent here. I would assume the parsers that go into `splitParse` are terminating and would give a proper result on either `Yield` or `Stop`. Using `extract` here seems inconsistent.  We could save the most recent `Yield` state, which should be extracted, given the stream ends. I don't think that is the behavior right now.

harendra-kumar: `extract` is implemented by the parser, it knows whether there is a consistent partial result available or not, if the parser is implemented correctly extract is guaranteed to  either return a partial result from the last yield or an error. It is doing exactly what you are expecting it to do.

src/Streamly/Internal/Data/Parser/Tee.hs
```diff
@@ -0,0 +1,297 @@
...
+-- Distributing Parsers
+-------------------------------------------------------------------------------
+
+data StepState s a = StepState s | StepResult a
+
+-- | State of the pair of parsers in a tee composition
+-- Note: strictness annotation is important for fusing the constructors
+{-# ANN type TeeState Fuse #-}
+data TeeState sL sR x a b =
+    TeePair !([x], StepState sL a, [x], [x]) !([x], StepState sR b, [x], [x])
```
pranaysashank, line 52: I think it'd be helpful if you can document what all the arguments here are supposed to be, would be useful while visiting this function in the future.

src/Streamly/Internal/Data/Parser/Tee.hs
```diff
@@ -0,0 +1,297 @@
...
+
+-- XXX: With the current "Step" semantics, it is hard to write, and not sure
+-- how useful, an efficient teeWith that returns a correct unused input count.
+--
+-- | @teeWith f p1 p2@ distributes its input to both @p1@ and @p2@ until both
+-- of them succeed or fail and combines their output using @f@. The parser
+-- succeeds if both the parsers succeed.
+--
+-- /Internal/
+--
```
pranaysashank, line 64: Just a reminded that this function's implementation is complicated, we shouldn't expose this combinator from non-internal modules without tests.

harendra-kumar:
I will merge this and update the comments once when all the PRs are merged.

harendra-kumar:
This has been merged, I have to incorporate some comments from the review, which I will do once all the parser's PRs are merged and then close this.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -118,33 +127,31 @@ all predicate = Parser step initial (return . Right)
 --
 -- /Internal/
 --
```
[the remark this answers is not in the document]
harendra-kumar: This has been take care of in the next PR.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -118,33 +127,31 @@ all predicate = Parser step initial (return . Right)
...
             ++ " elements, got " ++ show i
 
-        f x =
-            if n == i
-            then Right x
-            else Left err
-
 -- | Take at least @n@ input elements, but can collect more.
 --
 -- * Stops - never.
```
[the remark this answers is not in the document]
harendra-kumar: The fold never terminates, it never returns a Stop result. When the stream ends the parse driver needs to extract the last value. Not sure if the confusion here can be addressed succinctly.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1077,22 +1077,17 @@ splitParse (Parser pstep initial extract) (Stream step state) =
                             src  = Prelude.reverse src0
                         return $ Skip $ ParseChunksBuf src s buf1 pst1
                     -- XXX Halt 0 common case?
```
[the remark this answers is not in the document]
harendra-kumar: will do.

src/Streamly/Internal/Data/Parser/Tee.hs
harendra-kumar: yes.

harendra-kumar:
Addressed all the comments in #479 .

PR #498, opened 2020-04-17
fixed output explanation of wAsyncly & mentioned about concurrent mapping through mapM and sequence

wildcat26:
From this link : - 
https://hackage.haskell.org/package/streamly-0.7.1/docs/Streamly.html#t:WAsyncT

### DIFF
--- a/src/Streamly/Tutorial.hs
+++ b/src/Streamly/Tutorial.hs
@@ -1263,8 +1263,8 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- main = (S.'toList' $ fmap show $ S.'fromFoldable' [1..10]) >>= print
 -- @
 --
--- Also see the 'mapM' and 'sequence' functions for mapping actions, in the
--- "Streamly.Prelude" module.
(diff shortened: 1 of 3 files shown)


src/Streamly/Tutorial.hs
```diff
@@ -1263,7 +1263,7 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- main = (S.'toList' $ fmap show $ S.'fromFoldable' [1..10]) >>= print
 -- @
 --
--- Also see the 'mapM' and 'sequence' functions for mapping actions, in the
+-- Also see the 'mapM' and 'sequence' functions for mapping actions concurrently, in the
```
adithyaov, line 1266: This isn't always true. For serial streams, the actions are run sequentially.

harendra-kumar: The statement does not say that you can map _only_ concurrently, it is to say that you _can_ map concurrently too using those. We can perhaps rephrase it something like this:

```
'fmap' always maps serially, see 'mapM' which can map concurrently.
```

And we can remove "it is always serial" from the previous paragraph.

pranaysashank:
Merged with 0911988b

PR #500, opened 2020-04-17
Add CPS parser, use both CPS and direct style parsers

harendra-kumar:
CPS is good at scalable applicative/monad/alternative operations and direct style is good at stream fusion providing better performance. We use both via rewrite rules depending on the situation.

### DIFF
--- a/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
@@ -6,7 +6,7 @@
 #include "inline.hs"
 
 -- |
--- Module      : Streamly.ParserK.Types
+-- Module      : Streamly.Internal.Data.Parser.ParserK.Types
 -- Copyright   : (c) 2020 Composewell Technologies
 -- License     : BSD3
 -- Maintainer  : streamly@composewell.com
@@ -15,7 +15,7 @@
 --
 -- CPS style implementation of parsers.
 
-module Streamly.Internal.Data.ParserK.Types
+module Streamly.Internal.Data.Parser.ParserK.Types
     (
       Parser (..)
     )
--- a/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
@@ -14,10 +14,20 @@
 -- Portability : GHC
 --
 -- CPS style implementation of parsers.
+--
+-- The CPS representation allows linear performance for Applicative, sequenceA,
+-- Monad, sequence, and Alternative, choice operations compared to the
+-- quadratic complexity of the corresponding direct style operations. However,
+-- direct style operations allow fusion with ~10x better performance than CPS.
+--
+-- The direct style representation does not allow for recursive definitions of
+-- "some" and "many" whereas CPS allows that.
 
 module Streamly.Internal.Data.Parser.ParserK.Types
     (
       Parser (..)
+    , yield
+    , die
     )
 where
 
@@ -37,6 +47,10 @@ instance Functor m => Functor (Parser m a) where
     fmap f parser =
         MkParser $ \inp yieldk -> runParser parser inp (yieldk . fmap (fmap f))
 
+-- | See 'Streamly.Internal.Data.Parser.yield'.
+--
+-- /Internal/
+--
 {-# INLINE yield #-}
 yield :: b -> Parser m a b
 yield b = MkParser (\_ yieldk -> yieldk (Z.nil, Right b))
@@ -45,7 +59,10 @@ yield b = MkParser (\_ yieldk -> yieldk (Z.nil, Right b))
 -- Sequential applicative
 -------------------------------------------------------------------------------
 --
--- | 'Applicative' form of 'splitWith'.
+-- | 'Applicative' form of 'Streamly.Internal.Data.Parser.splitWith'. Note that
+-- this operation does not fuse, use 'Streamly.Internal.Data.Parser.splitWith'
+-- when fusion is important.
+--
 instance Monad m => Applicative (Parser m a) where
     {-# INLINE pure #-}
     pure = yield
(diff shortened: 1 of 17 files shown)


src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+-- /Internal/
+--
+{-# INLINE release #-}
+release :: Zipper m a -> Zipper m a
+release (Zipper [] _ _ _) = error "Bug: release, no checkpoint exists!"
+release (Zipper (n:cps) xs ys stream) =
+    assert (n <= length xs) $
+            case cps of
+                [] -> assert (n == length xs) $ Zipper [] [] ys stream
+                cp:rest -> Zipper ((cp + n) : rest) xs ys stream
```
adithyaov, line 153: Is this releasing the values held by the checkpoint?

src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+fromList xs = Zipper [] [] xs K.nil
+
+-- | Add a checkpoint to the Zipper so that we do not release the buffer beyond
+-- the checkpoint.
+--
+-- /Internal/
+--
+{-# INLINE checkpoint #-}
+checkpoint :: Zipper m a -> Zipper m a
+checkpoint (Zipper cps xs ys stream) = Zipper (0:cps) xs ys stream
```
adithyaov, line 140: Can you please elaborate on this?

src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+-------------------------------------------------------------------------------
+
+-- | @Zipper checkpoints lefts rights tail@.  The focus is on the first element
+-- of @rights@.  @lefts@ is buffered data on the right of the cursor.  @tail@
+-- is a stream that is used to generate more data if the cursor moves past
+-- @rights@.
+--
+-- @checkpoints@ is a stack of checkpoints. A new checkpoint is created by a
+-- @checkpoint@ operation. A checkpoint consists of a count that tracks how
+-- many elements we have yielded after the checkpoint was taken.  We need this
```
adithyaov, line 93: > A checkpoint consists of a count that tracks how many elements we have yielded after the checkpoint was taken.
I'm unable to follow this statement.

adithyaov, line 93: "after the checkpoint was taken" == "after the checkpoint was created"?

src/Streamly/Internal/Data/Parser/ParserK/Types.hs
```diff
@@ -0,0 +1,169 @@
...
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+#include "inline.hs"
+
+-- |
+-- Module      : Streamly.Internal.Data.Parser.ParserK.Types
+-- Copyright   : (c) 2020 Composewell Technologies
+-- License     : BSD3
```
pranaysashank, line 11: The appropriate SPDX identifier that cabal uses is "`BSD-3-Clause`"

src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+-- | Add a checkpoint to the Zipper so that we do not release the buffer beyond
+-- the checkpoint.
+--
+-- /Internal/
+--
+{-# INLINE checkpoint #-}
+checkpoint :: Zipper m a -> Zipper m a
+checkpoint (Zipper cps xs ys stream) = Zipper (0:cps) xs ys stream
+
+-- | Release the latest checkpoint, releases any values held by the checkpoint.
```
adithyaov, line 142: It does not necessarily release values held by the checkpoint. Or did I misunderstand what releasing values mean? According to me, releasing values = discarding values from the buffer.

src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+
+    where
+
+    {-# INLINE go #-}
+    go st buf !acc =
+        let stop = do
+                pst <- acc
+                r <- try $ extract pst
+                return $ case r of
+                    Left (e :: SomeException) ->
```
pranaysashank, line 214: We should `catch` `ParseError` here.

src/Streamly/Internal/Data/Parser/ParserK/Types.hs
harendra-kumar: Changed. But we need to change all the old files in the source code as well.

src/Streamly/Internal/Data/Zipper.hs
harendra-kumar: It releases the latest checkpoint, if the values are held by that checkpoint they are released from that checkpoint's perspective but the values may be held by other checkpoints so they will keep holding those. I am adding this note:

```
-- Note that the values may still be held by other checkpoints in the stack of
-- checkpoints.
```

src/Streamly/Internal/Data/Zipper.hs
harendra-kumar: Yes, thanks!

src/Streamly/Internal/Data/Zipper.hs
```diff
@@ -0,0 +1,380 @@
...
+                        (nil, Left (displayException e))
+                    Right b -> (nil, Right b)
+            single x = do
```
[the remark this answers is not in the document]
harendra-kumar: Good suggestion. Changed. If you have a good proposal for merging go/gobuf you can try it if the performance remains good we can make the change.

src/Streamly/Internal/Data/Zipper.hs
harendra-kumar: done.

PR #516, opened 2020-04-29
Update contributing file

tymefighter:
updated contributing file since newer versions of cabal allow us to do haddock instead of v2-haddock, also extra haddock arguments are also not needed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -135,8 +135,8 @@ To build haddock documentation for all modules, including the ones not exposed
 to users of the library, use the following commands:
 
 ```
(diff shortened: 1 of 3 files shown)


pranaysashank:
LGTM! Update the changelog to include this as a breaking change under the release version as Unreleased and it's good to go.

harendra-kumar:
@pranaysashank we can hold off the breaking change for now. We may have to do another patch release before the next major release. We can hold off minor breaking changes until we have to commit a major one.

pranaysashank:
Merged into master with e204d9c

PR #517, opened 2020-04-30
Split lib/benchmark/Prelude to decrease memory consumed during compilation

### LINKED ISSUE #490: Break benchmark/lib/Streamly/Benchmark/Prelude.hs into smaller modules
`benchmark/lib/Streamly/Benchmark/Prelude.hs` takes 5GB memory to build which is quite a lot. We break this down into smaller modules to reduce the memory used by GHC.

See also #433 .

adithyaov:
Closes #490


adithyaov:
The peak mempry usage for `cabal v2-build --enable-tests --enable-benchmarks --flag debug all -w /opt/ghc/8.8.1/bin/ghc` is about 2.5GB

harendra-kumar:
Some perf difference observations after the change:

* The lowest timings earlier were 32 us, now mostly 42 us in serial stream benchmarks
* serial zip benchmarks seem to have regressed from 110 ms to 126 ms.
* asyncly/toNull 29 ms to 73 ms, and similarly most other monad benchmarks
* tapAsync/tapAsyncS seem to have regressed

rest all look the same.

harendra-kumar:
the `tapAsync` benchmark change is because we now run it with threaded RTS and using two CPUs. Other benchmark regressions I have fixed, so we are now mostly at par with earlier benchmark results.

PR #519, opened 2020-05-01
Parser test

tymefighter:
simple test for yield and yieldM

### DIFF
--- a/test/Streamly/Test/Internal/Data/Parser.hs
+++ b/test/Streamly/Test/Internal/Data/Parser.hs
@@ -170,6 +170,15 @@ testTakeWhile1 =
                 (x : _) -> not $ predicate x
         where
             predicate = (== 0)
+    
+testSliceSepBy :: Property
+testSliceSepBy =
+    forAll (listOf (chooseInt (0, 1))) $ \ls ->
+        case S.parse (sliceSepBy predicate FL.toList) (S.fromList ls) of
+            Right parsed_list -> parsed_list == Prelude.takeWhile (not . predicate) ls
+            Left _ -> False
+        where
+            predicate = (== 1)
 
 main :: IO ()
 main = hspec $ do
@@ -194,3 +203,4 @@ main = hspec $ do
         prop "test for LookAhead function" testLookAhead
         prop "test for takeWhile function" testTakeWhile
         prop "test for takeWhile1 function" testTakeWhile1
+        prop "test for sliceSepBy function" testSliceSepBy
--- a/test/Streamly/Test/Internal/Data/Parser.hs
+++ b/test/Streamly/Test/Internal/Data/Parser.hs
@@ -8,7 +8,7 @@ import qualified Streamly.Internal.Data.Fold as FL
 
 import Test.Hspec(hspec, describe)
 import Test.Hspec.QuickCheck
-import Test.QuickCheck (forAll, chooseInt, Property, property, listOf, vectorOf, counterexample, (.&&.))
+import Test.QuickCheck (forAll, choose, Property, property, listOf, vectorOf, counterexample, (.&&.), Gen)
 
 import Test.QuickCheck.Monadic (monadicIO, PropertyM, assert, monitor)
 import Control.Monad.IO.Class (MonadIO(..))
@@ -47,6 +47,9 @@ listEquals eq parsed_list list = do
 
 checkListEqual :: (Show a, Eq a) => [a] -> [a] -> Property
 checkListEqual ls_1 ls_2 = monadicIO (listEquals (==) ls_1 ls_2)
+
+chooseInt :: (Int, Int) -> Gen Int
+chooseInt = choose
 
 -- Accumulator Tests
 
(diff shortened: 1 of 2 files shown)


pranaysashank:
Please rebase over master, and only retain the Parser test commits.

harendra-kumar:
Let's drop the "test" prefix from the test names and name them exactly the same as the combinator we are testing.

tymefighter:
@pranaysashank Can you take a look at this ?

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+--     property (case S.parse (P.shortest (P.yield (1 :: Int)) (P.die "die")) (S.fromList [1 :: Int]) of
+--         Right r -> r == 1
+--         Left _ -> False)
+--     .&&.
+--     property (case S.parse (P.shortest (P.die "die") (P.die "die")) (S.fromList [1 :: Int]) of
+--         Right _ -> False
+--         Left _ -> True)
+
+main :: IO ()
+main = hspec $ do
```
pranaysashank, line 278: Do write the description in line with how other test files have it. For example, take a look at other  test files `test/Streamly/Test/Array.hs`

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+import qualified Streamly.Internal.Prelude as S
+import qualified Streamly.Internal.Data.Fold as FL
+
+-- import Data.List (partition)
+
+import Test.Hspec(hspec, describe)
+import Test.Hspec.QuickCheck
+import Test.QuickCheck (forAll, chooseInt, Property, property, listOf, vectorOf, (.&&.))
+
+-- Accumulator Tests
```
pranaysashank, line 13: Refactor the number of elements or the list length (10000) that's used in every function to a top level variable like `max_value` or `max_elements` like in other test files.

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+takeEQ :: Property
+takeEQ =
+    forAll (chooseInt (0, 10000)) $ \n ->
+        forAll (listOf (chooseInt (0, 10000))) $ \ls ->
+            let 
+                list_length = Prelude.length ls
+            in
+                case S.parse (P.takeEQ n FL.toList) (S.fromList ls) of
+                    Right parsed_list -> (n <= list_length) && (parsed_list == Prelude.take n ls)
+                    Left _ -> n > list_length
```
pranaysashank, line 124: Separate a failure test to a separate property (n > list_length) and in one make property make sure at least `n` elements are always in the list and in another it need not be guaranteed.

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+
+-- Sequence Parsers Tests
+
+take :: Property
+take = 
+    forAll (chooseInt (0, 10000)) $ \n ->
+        forAll (listOf (chooseInt (0, 10000))) $ \ls ->
+            case S.parse (P.take n FL.toList) (S.fromList ls) of
+                Right parsed_list -> parsed_list == Prelude.take n ls
+                Left _ -> False
```
pranaysashank, line 113: In all the properties, instead of checking for equality just this normal way, you can do something equivalent to the `equal` or `listEquals` function in `test/Prop.hs` where it can provide more info for the failure case.

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+                case S.parse (P.takeEQ n FL.toList) (S.fromList ls) of
+                    Right parsed_list -> (n <= list_length) && (parsed_list == Prelude.take n ls)
+                    Left _ -> n > list_length
+
+takeGE :: Property
+takeGE =
+    forAll (chooseInt (0, 10000)) $ \n ->
+        forAll (listOf (chooseInt (0, 10000))) $ \ls ->
+            let 
+                list_length = Prelude.length ls
```
pranaysashank, line 131: Similar comments to the `takeEQ` function apply here

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+            in
+                case S.parse (P.takeGE n FL.toList) (S.fromList ls) of
+                    Right parsed_list -> (n <= list_length) && (parsed_list == ls)
+                    Left _ -> n > list_length
+
+lookAhead :: Property
+lookAhead =
```
pranaysashank, line 138: Add another test where you run a parser that consumes after `lookAhead`

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+            takeWithoutConsume = P.lookAhead $ P.take n FL.toList
+            parseTwice = do
+                parsed_list_1 <- takeWithoutConsume
+                parsed_list_2 <- takeWithoutConsume
+                return (parsed_list_1, parsed_list_2)
+        in
+            forAll (listOf (chooseInt (0, 10000))) $ \ls ->
+                case S.parse parseTwice (S.fromList ls) of
+                    Right (ls_1, ls_2) -> (ls_1 == ls_2) && (ls_1 == Prelude.take n ls)
+                    Left _ -> (list_length < n) || (list_length == n && n == 0)
```
pranaysashank, line 150: Here as well, make sure the list you generate have at least n elements. Instead of testing for the parse failures and successes in the same property you can separate them into separate ones where in one case the list will at least have n elements.

test/Streamly/Test/Internal/Data/Parser.hs
```diff
@@ -0,0 +1,305 @@
...
+        case S.parse (P.takeWhile1 predicate  FL.toList) (S.fromList ls) of
+            Right parsed_list -> case ls of
+                [] -> False
+                (x : _) -> predicate x && (parsed_list == Prelude.takeWhile predicate ls)
+            Left _ -> case ls of
+                [] -> True
+                (x : _) -> not $ predicate x
+        where
+            predicate = (== 0)
+
```
pranaysashank, line 175: In general, try to write one property to test one condition.

pranaysashank:
Merged in 646cb86f0d81032f7ddf94208f7e969c6f6d478d

PR #523, opened 2020-05-03
Cleanup hlint warnings

SanchayanMaity:
Another round of cleanups for #406


harendra-kumar:
@SanchayanMaity thanks for the PR. I have added a .hlint.ignore file in the streamly repo root. Now onwards hlint CI build will fail on the files that have already been cleaned. The files that are not yet clean are in `.hlint.ignore` file. So when you clean a file, also remove it from the .hlint.ignore file so that it is now checked in CI.

harendra-kumar:
merged in master.

PR #529, opened 2020-05-14
updated definition of parseK to use throwM instead of error

tymefighter:


pranaysashank:
Looks good! Just rebase over master and we can merge.

tymefighter:
I rebased over master

src/Streamly/Internal/Prelude.hs
```diff
@@ -517,7 +517,7 @@ where
 import Control.Concurrent (threadDelay)
 import Control.Exception (Exception, assert)
 import Control.Monad (void)
-import Control.Monad.Catch (MonadCatch, MonadThrow)
+import Control.Monad.Catch (MonadCatch, MonadThrow, throwM)
```
adithyaov, line 520: `MonadThrow(..)` instead of `MonadThrow, throwM` would have been better.

PR #531, opened 2020-05-15
Add exception handling routines for folds

harendra-kumar:


harendra-kumar:
Will merge this after #488

harendra-kumar:
Committed via #3011

PR #537, opened 2020-05-18
Fix sliceSepByMax

pranaysashank:
- Fix `sliceSepByMax`
- Add examples for `Parser.eqBy`

### DIFF
--- a/src/Streamly/Internal/Data/Parser.hs
+++ b/src/Streamly/Internal/Data/Parser.hs
@@ -570,6 +570,11 @@ groupBy = undefined
 
 -- | Match the given sequence of elements using the given comparison function.
 --
+-- >>> S.parse $ S.eqBy (==) "string" $ S.fromList "string"
+--
+-- >>> S.parse $ S.eqBy (==) "mismatch" $ S.fromList "match"
+-- > *** Exception: ParseError "eqBy: failed, yet to match 7 elements"
+--
 -- /Internal/
 --
 {-# INLINE eqBy #-}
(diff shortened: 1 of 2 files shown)


pranaysashank:
Circle CI failure is due to the token change and can be ignored.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -570,6 +570,11 @@ groupBy = undefined
 
 -- | Match the given sequence of elements using the given comparison function.
 --
+-- >>> S.parse $ S.eqBy (==) "string" $ S.fromList "string"
+--
+-- >>> S.parse $ S.eqBy (==) "mismatch" $ S.fromList "match"
+-- > *** Exception: ParseError "eqBy: failed, yet to match 6 elements"
```
harendra-kumar, line 576: There is something wrong with the count? yet to match 6 elements?

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -505,15 +505,17 @@ sliceSepByMax predicate cnt (Fold fstep finitial fextract) =
...
-            b <- fextract res
-            return $ Stop 0 b
+    step (Tuple' i r) a
+        | not (predicate a) = do
+            res <- fstep r a
+            let i1 = i + 1
+                s1 = Tuple' i1 res
+            if i1 < cnt
+                then return $ Yield 0 s1
+                else do
```
harendra-kumar, line 515: if-then-else style, don't indent `then` and `else`.

src/Streamly/Internal/Data/Parser.hs
pranaysashank: The word "match" isn't matched yet, and it has 6 letters in it.

PR #540, opened 2020-05-21
Add some time related combinators

harendra-kumar:
add: times, relTimes, timestamped
unimplemented skeletons: durations, ticks, timeout

Changes to the original currentTime combinator: remove delay from the first
event, cap the granularity to 1 ms to guarantee reasonable cpu usage.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -4071,7 +4071,7 @@ intersperseSuffix_ m (Stream step1 state1) = Stream step (Left state1)
         case r of
             Yield x s -> return $ Yield x (Right s)
             Skip s -> return $ Skip $ Left s
-            Stop -> m >> return Stop
+            Stop -> return Stop
 
     step _ (Right st) = m >> return (Skip (Left st))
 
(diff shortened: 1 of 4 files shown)


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -4583,40 +4583,55 @@ dropByTime duration (Stream step1 state1) = Stream step (DropByTimeInit state1)
...
+        return $ Skip $ Just (timeVar, tid, a)
 
-    step _ s@(Just (timeVar, _)) = do
+    step _ s@(Just (timeVar, _, t0)) = do
         a <- liftIO $ readVar timeVar
         -- XXX we can perhaps use an AbsTime64 using a 64 bit Int for
         -- efficiency.  or maybe we can use a representation using Double for
         -- floating precision time
-        return $ Yield (toAbsTime (MicroSecond64 a)) s
+        return $ Yield (toAbsTime (MicroSecond64 t0),
```
adithyaov, line 4636: `toAbsTime (MicroSecond64 t0)` Is this correct?
It never changes.

adithyaov: Ahhh..... NVM. (Coming back after reading `absTime`.)

adithyaov: The signature can be a little misleading.

PR #541, opened 2020-05-23
Streamly: Internal: Data: Clean up hlint warnings

SanchayanMaity:
Another round of clean ups for #406 

Have not removed `SmallArray/Types.hs` and `Pipe/Types.hs` from `.hlint.ignore` as they still need some clean up which I am not sure if they should be made or silenced. Will come back to it in the end I guess.

Signed-off-by: Sanchayan Maity <maitysanchayan@gmail.com>

### DIFF
--- a/.hlint.ignore
+++ b/.hlint.ignore
@@ -1,6 +1,5 @@
 src/Streamly/Internal/BaseCompat.hs
 src/Streamly/Internal/Control/Monad.hs
-src/Streamly/Internal/Data/SVar.hs
 src/Streamly/Internal/Data/Stream/SVar.hs
 src/Streamly/Internal/Data/Stream/Serial.hs
 src/Streamly/Internal/Data/Stream/Zip.hs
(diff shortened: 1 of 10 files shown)


SanchayanMaity:
```
src/Streamly/Internal/Data/Prim/Array/Types.hs:67:1-66: Suggestion: Use newtype instead of data

Found:

  data MutablePrimArray s a = MutablePrimArray (MutableByteArray# s)

Perhaps:

  newtype MutablePrimArray s a

    = MutablePrimArray (MutableByteArray# s)

Note: decreases laziness
```
The CI job hlint gives the above suggestion however I do not get hints on my local setup for the same. Also wouldn't think of changing this to newtype. Trying out anyways gives the below error.

```
src/Streamly/Internal/Data/Prim/Array/Types.hs:67:32: error:
    * A newtype cannot have an unlifted argument type
    * In the definition of data constructor `MutablePrimArray'
      In the newtype declaration for `MutablePrimArray'
   |
67 | newtype MutablePrimArray s a = MutablePrimArray (MutableByteArray# s)
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
hlint false positive? I can see an issue ticket [here](https://github.com/ndmitchell/hlint/issues/316) and [here](https://github.com/ndmitchell/hlint/pull/308).

pranaysashank:
Yeah the hlint suggestion is wrong. It makes sense only on the latest ghc with `UnliftedNewtypes` extension.

src/Streamly/Internal/Data/Sink.hs
```diff
@@ -57,7 +57,7 @@ module Streamly.Internal.Data.Sink
     )
 where
 
-import Control.Monad (when, void)
```
adithyaov, line 60: Maybe we could avoid the use of `arrows`?

src/Streamly/Internal/Data/Sink.hs
```diff
@@ -57,7 +57,7 @@ module Streamly.Internal.Data.Sink
...
+import Control.Monad (when, void, (>=>))
```
adithyaov, line 60: Maybe we could avoid the use of arrows?

PR #542, opened 2020-05-28
Fix classifySessionsOf

harendra-kumar:


benchmark/lib/Streamly/Benchmark/Prelude.hs
```diff
@@ -2141,6 +2151,8 @@ o_n_space_serial_toList value =
           -- , benchIOSink value "toPure" toPure
           -- , benchIOSink value "toPureRev" toPureRev
                 ]
+          , benchIOSink (value `div` 10) "classifySessionsOf"
```
adithyaov, line 2154: Change `classifySessionsOf` to `classifySessionsOf/10` maybe to indicate that you're using something other than the default value. It's not a big deal though.

harendra-kumar:
This got merged via CLI, but github does not let us change the merge target to master. Closing.

PR #545, opened 2020-05-30
Make changes to unfolds from stream

harendra-kumar:
* Add fromStreamK
* fromStream: signature change, use fromStreamK as fromStreamD has very bad
  performance characterstics.
* Add INLINEs to fromStream
* Remove fromStream1 and fromStream2 as they can be expressed trivially in
  terms of fromStream and lmap.

### DIFF
--- a/src/Streamly/Internal/Prelude.hs
+++ b/src/Streamly/Internal/Prelude.hs
@@ -3038,6 +3038,7 @@ concatMapWith = K.concatMapBy
 -- @
 -- concatMap = 'concatMapWith' 'Serial.serial'
 -- concatMap f = 'concatMapM' (return . f)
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Prelude.hs
```diff
@@ -3038,6 +3038,7 @@ concatMapWith = K.concatMapBy
 -- @
 -- concatMap = 'concatMapWith' 'Serial.serial'
 -- concatMap f = 'concatMapM' (return . f)
+-- concatMap f = concatUnfold (UF.lmap f UF.fromStream)
```
adithyaov, line 3041: Use apostrophes on RHS?

PR #550, opened 2020-06-05
Extend Prim Array to be the default

adithyaov:
* [x] Make data mutable array
* [x] Modify immutable data array to wrap over mutable data array
* [x] Extend immutable data array with additional functions
* [x] Make a new pinned data array (mutable & immutable)
* [x] Wrap almost all the functions using existing immutable data array except `resizeArray` and `newArray`
* [ ] Add groupIOVecsOf (optional?)
* [x] Change the main datatype
* [ ] Move extra files to a better location (optional)
* [x] Final cleanup (remove warnings etc.)
* [x] Fix fusion problems


src/mutable-prim-array-types.hs
```diff
@@ -1,27 +1,31 @@
...
 import qualified Streamly.Internal.Data.Stream.StreamD.Type as D
 
+import GHC.Exts
+import Prelude hiding (length, unlines)
+
 -------------------------------------------------------------------------------
 -- Array Data Type
 -------------------------------------------------------------------------------
 
+-- XXX why is this not supposed to be used?
```
adithyaov, line 19: I've written this at the beginning with something in mind about a clean data type.

src/mutable-prim-array-types.hs
```diff
@@ -1,27 +1,31 @@
...
 -- This is not supposed to be used
 data Array s a = Array (MutableByteArray# s)
 
+-------------------------------------------------------------------------------
+-- Utilities
+-------------------------------------------------------------------------------
+
+-- XXX Document important details about semantics e.g. is overlapped copy
```
adithyaov, line 27: Will do, overlapped copy is allowed.

src/mutable-prim-array-types.hs
```diff
@@ -31,74 +35,106 @@ unsafeCopy ::
...
+-------------------------------------------------------------------------------
+-- Length
+-------------------------------------------------------------------------------
+
+-- XXX rename to byteCount?
+{-# INLINE byteLength #-}
+byteLength :: Array s a -> Int
+byteLength (Array arr#) = I# (sizeofMutableByteArray# arr#)
+
+-- XXX Rename length to elemCount so that there is no confusion bout what it
```
adithyaov, line 53: I don't think there would be any confusion with `length` but `elemCount` sounds to the point.

src/mutable-prim-array-types.hs
```diff
@@ -31,74 +35,106 @@ unsafeCopy ::
...
-    a3 <- resizeArray a1 (l1 + l2)
-    unsafeCopy a2 0 a3 l1 l2
-    return a3
-  where
-    l1 = length a1
-    l2 = length a2
+    -> Int
+    -> m a
+unsafeIndexM (Array arr#) (I# i#) =
+    -- XXX XXX why not use readByteArray#
```
adithyaov, line 77: Will change, this isn't a bug though, is it?

src/mutable-prim-array-types.hs
```diff
@@ -31,74 +35,106 @@ unsafeCopy ::
...
 
+-- Note: We do not store the actual length of the array in the Array
+-- constructor. Therefore, for "length" API to work correctly we need to match
+-- the ByteArray length with the used length by shrinking it.
+--
+-- However, it may be expensive to always shrink the array. We may want to
+-- shrink only if significant space is being wasted. If we want to do that then
+-- we will have to store the used length separately. Or does GHC take care of
+-- that?
+--
```
adithyaov, line 104: GHC does not take care of it. Also, I don't think in place shrinking is expensive though. Will check. Storing the length separately looks very ugly. I would not recommend it if it isn't required.

src/mutable-prim-array-types.hs
```diff
@@ -128,11 +167,9 @@ fromStreamDN ::
...
 fromStreamDN limit str = do
     marr <- newArray (max limit 0)
-    _ <-
-        D.foldlM'
-            (\i x -> i `seq` (writeArray marr i x) >> return (i + 1))
-            0 $
-        D.take limit str
+    let step i x = i `seq` (writeArray marr i x) >> return (i + 1)
+    void $ D.foldlM' step 0 $ D.take limit str
+    -- XXX XXX we need to shrink the array in case the stream terminated early.
```
adithyaov, line 172: Yes, Good catch!

adithyaov:
> Is it possible to cleanly separate the commits for:
>
>    implementation of new arrays
>   replacement of memory array

I'll see what I can do.

> Let's not create the ArrayStream module. Most operations are trivially
> expressed in terms of other combinators. The remaining ones will also be
> changed such that they can be expressed in terms of others.

This PR has too many changes already, Other than some trivial changes + documentation, I'd like to make the remaining changes seperately.

src/mutable-prim-array-types.hs
harendra-kumar: `length` is used everywhere, in lists, streams etc, so that way it is consistent. Maybe we can just leave it like that.

src/mutable-prim-array-types.hs
harendra-kumar: Does GHC actually reallocate and copy the array even if it is to be shrunk by one byte? Or does it just adjust the length in some cases?

src/mutable-prim-array-types.hs
harendra-kumar: The state token is not used during the read.

harendra-kumar:
> This PR has too many changes already, Other than some trivial changes + documentation, I'd like to make the remaining changes seperately.

This is mostly about removing some of the changes that have been made. The `ArrayStream` module is not required and there are two of them. The one combinator (`compact`) that we actually need can be pushed to the Array module itself.

I am ok with doing it in another PR, but we will unnecessarily add a lot of code and then remove it.

harendra-kumar:
The new version of `primitive` adds these:

```
class (PrimMonad m, s ~ PrimState m) => MonadPrim s m 
class (PrimBase m, MonadPrim s m) => MonadPrimBase s m 
```

We should make use of these. I also like the name `MonadPrim`, it is also consistent with `MonadIO`. In fact I was thinking of creating these definitions. For older versions of primitive we can use these definitions under a CPP conditional.

src/mutable-prim-array-types.hs
adithyaov: The shrinking is always inplace.

adithyaov:
> We should make use of these. I also like the name MonadPrim, it is also consistent with MonadIO. In fact I was thinking of creating these definitions. For older versions of primitive we can use these definitions under a CPP conditional.

`MonadPrim` does sound better. Maybe this change can be made later?

src/mutable-prim-array-types.hs
adithyaov: The idea is that I can only read from a mutable byte given its current state after changes etc. Else, I think the results wont be as expected due to optimizations. Eitherway, we have `readByteArray#` now.

harendra-kumar:
I added the proposed naming plan here: https://github.com/composewell/streamly/blob/module-org/design/module-organization.md . Take a look at the last section in this.

We also need to move the header files to the right places. We cannot have them in `src`.

adithyaov:
> We also need to move the header files to the right places. We cannot have them in src.
Yes, that is one of the milestones of this PR.

adithyaov:
This is replaced by #601 and #602

PR #566, opened 2020-06-28
Use extensions by default instead of using them per file

harendra-kumar:
Most extensions do not show any compilation, functionality or performance related issues. Except for the following three we can enable all others:

* UndecidableInstances -- Don't want to use unless indicated
* TypeFamilies         -- Causes large regressions and improvements too!
* UnboxedTuples        -- interferes with (#.)

We need to investigate TypeFamilies, why does it cause regressions and improvements.


pranaysashank:
* Do we add `TemplateHaskell` to the list of default-extensions? I think
it might create confusing error messages.
* Use `default-extensions` for tests and benchmarks?

harendra-kumar:
No for `TemplateHaskell`, yes for tests and benchmarks.

pranaysashank:
@harendra-kumar Can you restart appveyor?

harendra-kumar:
restarted.

PR #568, opened 2020-06-28
Fix take, takeEQ, sliceSepByMax.

pranaysashank:
- `take` and `takeEQ` now work when the count is 0.
- `sliceSepByMax` behaviour is now more in line with what the
  documentaton says in `Streamly.Internal.Data.Parser`.

- TODO: Tests need to be fixed, a bit unclear how best to proceed with them.

### DIFF
--- a/src/Streamly/Internal/Data/Parser/ParserD.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserD.hs
@@ -346,9 +346,11 @@ take n (Fold fstep finitial fextract) = Parser step initial extract
 --
 {-# INLINE takeEQ #-}
 takeEQ :: MonadThrow m => Int -> Fold m a b -> Parser m a b
-takeEQ n (Fold fstep finitial fextract) = Parser step initial extract
+takeEQ cnt (Fold fstep finitial fextract) = Parser step initial extract
 
     where
+
+    n = max cnt 0
 
     initial = Tuple' 0 <$> finitial
 
--- a/src/Streamly/Internal/Data/Parser/ParserD.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserD.hs
@@ -328,13 +328,15 @@ take n (Fold fstep finitial fextract) = Parser step initial extract
 
     initial = Tuple' 0 <$> finitial
 
-    step (Tuple' i r) a = do
-        res <- fstep r a
-        let i1 = i + 1
-            s1 = Tuple' i1 res
-        if i1 < n
-        then return $ Partial 0 s1
-        else Done 0 <$> fextract res
+    step (Tuple' i r) a
+        | i < n = do
+            res <- fstep r a
+            let i1 = i + 1
+                s1 = Tuple' i1 res
+            if i1 < n
+            then return $ Partial 0 s1
+            else Done 0 <$> fextract res
+        | otherwise = Done 1 <$> fextract r
 
     extract (Tuple' _ r) = fextract r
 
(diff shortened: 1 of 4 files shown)


pranaysashank:
What's the error in appveyor? I don't see anything in the log

adithyaov:
> What's the error in appveyor? I don't see anything in the log

Test suite properties failed

harendra-kumar:
I have mentioned this error several times. This comes once in a while on appveyor. This is not related to your change.

pranaysashank:
@harendra-kumar Can you restart it?

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -330,15 +330,15 @@ take n (Fold fstep finitial fextract) = Parser step initial extract
 
     where
 
-    initial = Tuple' 0 <$> finitial
+    initial = Tuple' n <$> finitial
 
-    step (Tuple' i r) a = do
+    step (Tuple' i r) a
+      | i == 0 = Done 1 <$> fextract r
```
harendra-kumar, line 336: We can write the code to use `Done 0` instead of `Done 1` so that we can return the result in the earlier step itself and we do not have to return an element back to the stream. The previous code in fact was doing that. We just need to replace `res` with `s1` in the `Done` constructor in the earlier code to fix the issue.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -352,16 +352,18 @@ takeEQ n (Fold fstep finitial fextract) = Parser step initial extract
...
```
harendra-kumar, line 358: The comment I made in `take` applies here as well. And the fix is also the same. In the old code just replace `res` with `s1`.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -504,18 +506,19 @@ sliceSepByMax predicate cnt (Fold fstep finitial fextract) =
...
     where
 
-    initial = Tuple' 0 <$> finitial
+    initial = Tuple' cnt <$> finitial
+
     step (Tuple' i r) a
-        | not (predicate a) = do
+      | not (predicate a) =
+        if i == 0
+        then Done 1 <$> fextract r
```
harendra-kumar, line 514: Ditto here as well.

src/Streamly/Internal/Data/Parser/ParserD.hs
harendra-kumar: Sorry, scratch the previous fix suggestion. How about this, with minimal changes to the old code and keeping the common branch case first:

```
    step (Tuple' i r) a
        | i < n = do
            res <- fstep r a
            let i1 = i + 1
                s1 = Tuple' i1 res
            if i1 < n
            then return $ Partial 0 s1
            else Done 0 <$> fextract res
        | otherwise = Done 1 <$> fextract r
```

harendra-kumar: Also the current fix has a bug when `n` is negative, it will never terminate.

Can we also add these two test cases:

1) when n is 0
2) when n is negative

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -501,17 +507,18 @@ sliceSepByMax predicate cnt (Fold fstep finitial fextract) =
...
-                b <- fextract res
-                return $ Done 0 b
+        | not (predicate a) =
+            if i < cnt
+            then do
+                res <- fstep r a
+                let i1 = i + 1
+                    s1 = Tuple' i1 res
+                return $ Partial 0 s1
+            else Done 1 <$> fextract r
```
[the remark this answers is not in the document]
harendra-kumar: I took another look at it, and it looks fine. Because when we reach the Done case we are already looking at the next element and we have checked that it is not passing the predicate, so we return it back. If it is passing the predicate then we will reach the Done 0 case where we drop the separator.

The ultimate proof is a test case for this case.

harendra-kumar:
I am not waiting any more on this, merging it. If there are any more changes they can be made via a separate PR.

PR #570, opened 2020-06-29
Add MonadFail instance for Parser.

pranaysashank:
I didn't add a MonadFail instance for ParserD type as it doesn't feel
useful and also necessitates adding a `MonadThrow m` constraint to the
`Monad` instance of `ParserD`. Let me know if I should add it.

### DIFF
--- a/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserK/Types.hs
@@ -31,6 +31,10 @@ import Control.Applicative (Alternative(..))
 import Streamly.Internal.Data.Zipper (Zipper)
 import qualified Streamly.Internal.Data.Zipper as Z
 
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Data/Parser/ParserK/Types.hs
```diff
@@ -145,6 +149,19 @@ instance Monad m => Monad (Parser m a) where
                 Left err -> runParser (die err) z yieldk
         in runParser m inp yield1
 
+#if !(MIN_VERSION_base(4,13,0))
```
harendra-kumar, line 152: Do we need this for base > 4.8 && < 4.13?

pranaysashank: Yes, `fail` is a method of `Monad` in those versions

pranaysashank:
Merged with 51127df9

Issue #150, opened 2018-11-30
Use a monadic zero element in monadic folds


harendra-kumar:
We changed the signature of `foldrM` to be consistent with that of lists, like this:

```
foldrM :: Monad m => (a -> b -> m b) -> b -> SerialT m a -> m b
```

We used `b` instead of `m b` as zero element. However, this has a drawback, now we cannot perform a monadic action automatically whenever the stream ends. It seems to me that the previous signature was better.

We should change this for all monadic folds.

@Abhiroop do you have any feedback on this?

harendra-kumar:
This may not be much of an issue for foldlM' as we can compose a monadic action like this, to get the equivalent effect:

```
m >>= \z -> foldlM' f z xs
```

However, this is not possible in case of foldrM as we should run the action when `z` is used and we do not have control over when `z` is used. But for consistency of right and left folds we should make this change in foldlM as well.

We have to keep scanlM variants also consistent with the same.

harendra-kumar:
Commit 0ce4a6d5adbce5673d820817b2493d52bfa29fbe changed the foldrM signature to use a monadic zero element. This was needed to keep foldrM lazy.

For left folds a pure initial element is ok as we have to execute the action immediately anyway to initialize the accumulator.  So we have decided to keep the existing signatures.

harendra-kumar:
Reopening this issue. We need to have the zero as monadic in `foldlM` and `scanlM`. The reason is e.g. when implementing `classifySessionsBy` using the system time instead of stream timestamps, we want to initialize the accumulator with a current system time. To be able to do that when we have a non-monadic zero, we need to call the time function before calling `scanlM` which means we need the stream to be a monad and need to use a bind. This could affect performance badly. It is much more efficient if we perform the monadic action inside the fold instead of inside the stream monad.

harendra-kumar:
In addition to a monadic zero we need to pass the accumulator as a monadic value to the step function. The signature should look like:

```
foldlM' :: Monad m => (m b -> a -> m b) -> m b -> SerialT m a -> m b 
```

There are two reasons for this:

1) With the current signature we cannot have the monad `m` as a list monad or a stream monad e.g.  we should be able to fold `SerialT (SerialT m) a`. In the existing signature, the accumulator is using just the value `b` so we cannot provide the whole stream to it, but only individual elements.

2) That will make foldlM' dual of foldrM with `foldr f z xs = foldl (flip f) z (reverse xs)` holding.

[...]

PR #571, opened 2020-06-29
Change the signature of foldlM' to use a monadic accumulator.

pranaysashank:
It feels a bit awkward to use this new fold signature. Maybe it'd be
better if we provide functions with both the fold signatures.

- [x] Fixed: foldl' and foldlM' strictness tests are failing.


harendra-kumar:
I updated #150. We only need a monadic initial action and not a monadic accumulator. Please change the PR accordingly.

pranaysashank:
Changed the PR to only use a monadic seed.

harendra-kumar:
Most benchmarks look ok except `writeN` and `foldl'` in Array benchmarks show some regression. I don't have much time to further investigate so I am ignoring that and merging it.

PR #572, opened 2020-06-29
Rework the installation guide.

jkoshy:
- Use an informal register.
- Use the active voice.
- Make the guide accessible to people new to Haskell.
- Add a section suggesting a few next steps.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,76 +1,227 @@
-# Using `streamly` package
+<!--
+Portions (c) 2020, Google LLC.
+SPDX-License-Identifer: BSD-3-Clause
+-->
 
-If you are new to Haskell please take a look at [Haskell Getting Started
-Guide](https://github.com/composewell/haskell-dev/blob/master/getting-started.rst).
+# Getting started with the `streamly` package
 
-The following instructions assume that you are using a shell on a POSIX
-system. On Windows you need a shell under `msys`, some of the commands
-may have slight differences.
+This guide will show you how to write a simple program that uses
+`streamly`, using the [`cabal`](https://www.haskell.org/cabal/) build
+manager for Haskell.
 
-## Which version to use?
+<!-- TODO: Add instructions for `stack` and `nix`.
 
-* [Latest stable release](https://hackage.haskell.org/package/streamly) is available on
-  Hackage.
-* [Latest development version](https://github.com/composewell/streamly) is
-  available on github.
+If you are using `stack` or `nix` please make sure to add the latest
+version from Hackage to your tool configuration.  -->
 
-If you are using `stack` or `nix` please make sure to add the latest version
-from Hackage to your tool configuration.
+No prior knowledge of either Haskell or of `cabal` is needed.  We do
+however assume that you are using a command-line shell on a POSIX
+operating system.  If you are running Windows&trade; then you may need
+to run a command-line shell under `msys` and some of the commands
+below may also need to be changed in small ways to make them work.
 
-## Recommended build options
+**Note**: If you are new to Haskell you may find the [Haskell Getting
+Started
+Guide](https://github.com/composewell/haskell-dev/blob/master/getting-started.rst)
+to be useful.
 
-See [recommended compilation options here](docs/Build.md).
+If you know your way around Haskell, and have an up to date toolchain
+already installed, then you can jump straight to the section titled
+"[_Prepare Your Build Directory_](#user-content-prepare-your-build-directory)".
 
-## Using with `cabal`
+## Getting started with `streamly` using `cabal`
 
-Make sure you have `cabal` version 3.0 or later installed and you have `ghc`
-available in your PATH. Refresh your package list:
+To get started, you will need a recent `cabal` binary to be installed
+on your system.
+
+### Installing `cabal`
+
+A pre-packaged `cabal` binary may already be available for your
+operating system.
+
+For example, on Debian GNU/Linux, you could install `cabal` using::
+
+```
+$ sudo apt install cabal-install
+```
+
+### Installing `ghc`
+
+You would also need to install the Glasgow Haskell Compiler `ghc`.  In
+most operating systems, using the system's package manager to install
+`cabal` would also bring in a version of `ghc` as a dependency.
+
+Please verify that `ghc` is present:
+
+```
+$ ghc --version
+The Glorious Glasgow Haskell Compilation System, version 8.0.2
+```
+
+If `ghc` isn't present, please install it using your operating
+system's package manager. For example, on Debian GNU/Linux, use:
+
+```
+$ sudo apt install ghc
+```
+
+### Update `cabal` if necessary
+
+Many operating systems ship an old version of `cabal`; in order to use
+`streamly` you should use `cabal` version 3.0 (or later).
+
+You can check the installed version of `cabal` by using its
+`--version` flag:
+
+```
+$ which cabal
+/usr/bin/cabal
+$ /usr/bin/cabal --version
+cabal-install version 1.24.0.2
+compiled using version 1.24.2.0 of the Cabal library
+```
+
+In this example, the installed `cabal` binary is too old for our
+needs. It would need to be upgraded.
+
+You can ask `cabal` to upgrade itself (please see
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -1,76 +1,217 @@
...
 
-## Recommended build options
+## Getting started with `streamly` using `cabal`
 
-See [recommended compilation options here](docs/Build.md).
+To get started, you will need a recent `cabal` binary to be installed
+on your system.
 
-## Using with `cabal`
+### Installing (and updating) `cabal`
```
harendra-kumar, line 28: I am wondering if this section has gotten too big and a bit distracting for this guide? An alternative is to move this to the "Haskell getting started guide" and put a reference to that section here? I know that guide may also need some rework to fit it there.

INSTALL.md
```diff
@@ -1,76 +1,217 @@
...
+the `build-depends` section of the generated cabal file
+`streamly-play.cabal`.  Please see the [Cabal User
+Guide](https://www.haskell.org/cabal/users-guide/) for more
+information on configuring `cabal`.
+
+This invocation also creates a skeletal `Main.hs` program which we
+will use later.
 
-### Using hackage version in repl
+### Try `streamly` at the GHCi REPL
```
harendra-kumar, line 147: Shall we say `in the` GHCi REPL instead of `at the`? For example, I usually say `in the shell` rather than `at the shell`.

jkoshy: Done.  Looking around, common usage seems to be 'in the REPL' or 'at the REPL prompt'.

INSTALL.md
jkoshy: Yes there is some degree of overlap currently.

For this rewrite, I assumed that the reader of this guide is likely to be curious about `streamly` but also new to Haskell -- since, statistically speaking, that is likely to be the case today. The idea then was offer this new user a simple, 'linear', series of steps that they could follow to get `streamly` up and running quickly on their system.

The 'Getting Started' guide is a great resource, but it covers more topics than strictly needed to bring up a `streamly` environment.  This means that the burden of sifting out the parts of the document that are relevant to bringing up `streamly` would then fall onto the reader, who may or may not know enough to decide.

One way out could be to split the 'Getting Started' guide into focussed sections that could perhaps be referenced from this guide.

Even better perhaps could be to offer a `stack` template that offers everything preconfigured.  With this approach, new users could get started using `streamly` with just two steps:

1. Install `stack`.
2. Run `stack new project-name streamly-template`.

harendra-kumar: > One way out could be to split the 'Getting Started' guide into focussed sections that could perhaps be referenced from this guide.

Yes, that's what I meant. But we can keep it the way you have written for now to not get distracted. I agree with all the previous paragraphs as well.

I think one complication here is because of the cabal upgrade instructions. If we install cabal directly then we do not need upgrade. To make the install simple I have used the `ghcup` method as the preferred method in the Haskell guide. It installs latest versions of cabal and ghc both in one simple step.

I like the direct method instead of using templates because otherwise the users have to be aware of what the templates are, how they work, what exactly they will do etc. And templates are not very useful beyond beginners. If we keep that out of the way, then in both cabal and stack the process is pretty similar except that in cabal you have to install ghc separately:

* install cabal/ghc, stack
* cabal/stack init/new
* ...the rest

If we use `ghcup` to install cabal/ghc then both are equivalent.

INSTALL.md
```diff
@@ -1,76 +1,221 @@
...
+compiled using version 1.24.2.0 of the Cabal library
+```
+
+In this example, the installed `cabal` binary is too old for our
+needs. It would need to be upgraded.
+
+You can ask `cabal` to upgrade itself:
+
+```
+$ cabal install Cabal cabal-install
```
harendra-kumar, line 85: This is a bit heavy upgrade method, may take a long time. Also, the newer `cabal install` requires `--installdir` option.  This is another reason to prefer `ghcup`. 

Also, I am wondering why we need `Cabal` here, won't just `cabal install cabal-install` work?

INSTALL.md
```diff
@@ -1,76 +1,221 @@
...
+cabal-install version 3.0.0.0
+compiled using version 3.0.2.0 of the Cabal library
+```
+
+If you have upgraded `cabal`, do be sure to specify `$HOME/.cabal/bin`
+early in your `PATH` shell variable, so that the new binary gets used
+instead of the older version on your system.
+
+```
+$ PATH=$HOME/.cabal/bin:$PATH
```
harendra-kumar, line 110: We will also need to put that in `.bash_profile` or equivalent. `ghcup` takes care of that as well.

INSTALL.md
```diff
@@ -1,76 +1,221 @@
...
+
+This invocation also creates a skeletal `Main.hs` program which we
+will use later.
 
-### Using hackage version in repl
+### Try `streamly` in the GHCi REPL
 
-Run repl with the latest version from Hackage:
+You can now try out `streamly` using the `GHCi` interpreter's
+read-eval-print (REPL) facility.
```
harendra-kumar, line 158: change to `read-eval-print-loop` to make it consistent with the acronym?

INSTALL.md
jkoshy: > If we install cabal directly then we do not need upgrade ...

True.  However, the downside of these "direct" installation methods is that they require users to download and run pre-compiled binaries of unknown provenance.

INSTALL.md
jkoshy: Done.

INSTALL.md
jkoshy: That is what <https://www.haskell.org/cabal/> recommends.  I've added a link to that page.

jkoshy: > This is a bit heavy upgrade method, may take a long time.

It does take a while, but it rebuilds `cabal` from source. If the local Haskell tool chain has not been compromised we should end up with a binary that reflects the source code for `cabal`.

> This is another reason to prefer ghcup.

Unfortunately `ghcup` seems to require us to download and run a binary of unknown provenance.

INSTALL.md
harendra-kumar: I guess stack install is similar in nature to ghcup and cabal binary download. So all of them should suffer from the same security issues. I am wondering if you trust the website, are using https and verifying the signatures then is it sufficient?

jkoshy:
> Looks good, we can merge it. Unless you have something more in the works. Would you like to rebase and squash the commits?

I've squash-merged the changes into a single commit.  Please let me know if anything else is needed.

> Also, you may want to add Contributor/Copyright info as per Google's legal requirements.

Done (per my understanding of their requirements).

INSTALL.md
jkoshy: I believe `stack` offers a mode for upgrading from source code (`stack upgrade --source-only`).

I noticed that the installation instructions for a number of new installers (`ghcup`, `rustup`, etc) invite the user to directly execute a binary downloaded from an external website.

This means that the user would need to trust that:

1. That the source code used to build that tool was trustworthy.
2. That the source code for the tool reached the compilation toolchain on the development machine unmodified.
3. That the compilation toolchain on that machine had not been tampered with.
4. That the executable that was made available on the website was exactly that which was built.
5. That the executable that reached the local disk was exactly that which was made available on the website.

For open-source code, we can protect against (1) with a manual audit.  For (4) and (5) we can sign the built binary and insist that the user verify its signature before proceeding further. However (2) and (3) to defend against.

Building from source reduces the attack surface somewhat, although we would still need to trust that the versions of `stack`, `cabal`, `ghc` and the rest of the toolchain in the default package repository had been bootstrapped correctly.

harendra-kumar: Trust is hard and at odds with convenience. For example, (1) involves a very high cost, you need someone who is competent and trustworthy to do it. The rest can be automated using a trust chain (perhaps only in theory as of now) using the right tools which are themselves verified for trust. But then again there could be tricky issues like the possibility of code injection if the code is running on an untrusted machine. I guess we can stop at a point where the ROI is sufficiently low.

harendra-kumar:
Thanks @jkoshy! Looking forward to more awesome stuff from you.

PR #579, opened 2020-07-06
Use default-extensions for tests.

pranaysashank:
- Should we add MonadComprehensions, OverloadedStrings,
  OverloadedLists to the list of default extensions?


harendra-kumar:
No we should not add these. Also, before we add any new extensions to this list we should run all performance tests.

The PR looks good, let's merge it.

PR #589, opened 2020-07-11
Make bench.sh compatible with the cabal option "--builddir"

adithyaov:
Something like this is possible:
`./bench.sh --cabal-build-flags "--builddir=dist-newstyle/$(git rev-parse --abbrev-ref HEAD)"`
which will help reduce recompilation when switching branches.

### DIFF
--- a/bench.sh
+++ b/bench.sh
@@ -184,7 +184,6 @@ set_benchmarks() {
   fi
 }
 
-# $1: benchmark name (linear, nested, base)
 find_report_prog() {
     local prog_name="chart"
     hash -r # XXX Why is this required?
@@ -208,7 +207,7 @@ find_report_prog() {
 # The path is dependent on the architecture and cabal version.
 # Use this command to find the exe if this script fails with an error:
 # find .stack-work/ -type f -name "benchmarks"
-
+# $1: benchmark name (Prelude.Serial, Prelude.WSerial)
 stack_bench_prog () {
   local bench_name=$1
   local bench_prog=`stack path --dist-dir`/build/$bench_name/$bench_name
@@ -220,6 +219,7 @@ stack_bench_prog () {
   fi
 }
 
+# $1: benchmark name (Prelude.Serial, Prelude.WSerial)
 cabal_bench_prog () {
   local bench_name=$1
   local bench_prog=`$WHICH_COMMAND $1`
@@ -231,6 +231,7 @@ cabal_bench_prog () {
   fi
 }
 
+# $1: benchmark name (Prelude.Serial, Prelude.WSerial)
 bench_output_file() {
     local bench_name=$1
     echo "charts/$bench_name/results.csv"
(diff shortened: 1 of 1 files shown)


adithyaov:
This has one downside of building `chart` for every different `build-dir`, but only once. I cannot find a simple and clean way to avoid this. One solution includes to only build it in default builddir `dist-newstyle` but this would require a different `WHICH_COMMAND`.

harendra-kumar:
We can build chart once and install it in the PATH and bench.sh can pick it up from PATH. It is a very stable binary now, it would rarely change.

bench.sh
```diff
@@ -429,8 +429,17 @@ CABAL_BUILD_FLAGS=""
...
 
+CABAL_BUILD_DIR="dist-newstyle"
+
 cabal_which() {
-  find dist-newstyle -type f -path "*${GHC_VERSION}*/$1"
+  if test -z "$CABAL_BUILD_DIR"
+  then
+    CABAL_BUILD_DIR="dist-newstyle"
+  fi
+  if test -d "$CABAL_BUILD_DIR/build"
```
harendra-kumar, line 439: Why do you need this test? find would anyway fail if the directory does not exist.

adithyaov: Yeah, I did not like the failure message:`find: ‘dist-newstyle/branch/build’: No such file or directory`. It gives the idea that something has gone wrong when it is completely normal.

adithyaov: But then again, I'll be changing a few things so this might be removed. The problem was chart related.

bench.sh
```diff
@@ -536,7 +553,7 @@ else
   # XXX cabal issue "cabal v2-exec which" cannot find benchmark/test executables
   #WHICH_COMMAND="cabal v2-exec which"
   WHICH_COMMAND=cabal_which
-  BUILD_CHART_EXE="cabal v2-build --flags dev chart"
+  BUILD_CHART_EXE="cabal v2-build --builddir=$CABAL_BUILD_DIR --flags dev chart"
```
harendra-kumar: Is this still needed?

adithyaov: You're commenting on outdated code sir. Outdated I say.

bench.sh
```diff
@@ -457,13 +448,25 @@ do
...
 GAUGE_ARGS=$*
 
+# Parse cabal build flags for any quirks
+# Dont specify the "build" directory of another build.
+# Legal: dist-newstyle, dist-newstyle/commit-hash
+# Illegal: dist-newstyle/build
+for option in "$CABAL_BUILD_FLAGS"
+do
+  case $option in
+    --builddir=*) CABAL_BUILD_DIR=${option:11}; break ;;
```
harendra-kumar, line 466: You may want to quote this (`CABAL_BUILD_DIR="${option:11}"`) in case there are spaces in the dir name.

adithyaov:
I'm going to close this PR. We can open it once `chart` is merged with `bench-show`. It's inconvinient for users to think about multiple executables.

PR #600, opened 2020-07-15
Add tests & fix doc for groupsBy

wildcat26:


test/Prop.hs
```diff
@@ -990,6 +1000,55 @@ monadBind constr eq t (a, b) = withMaxSuccess maxTestCount $
...
+
+-- after grouping (and folding) Int stream using (>) operation,
+-- the first Int of every [Int] (in the [Int] stream) should be the minimum
+testGroupsBy :: Property
+testGroupsBy =
+    forAll (choose (0, maxStreamLen)) $ \len ->
+            forAll (vectorOf len (arbitrary :: Gen Int)) $ \vec -> monadicIO $ do
+                a <- run $ S.all (\l -> case l of
+                                          [] -> True
+                                          (ex:_) -> ex == minl intMax l) $ S.groupsBy (>) FL.toList $ S.fromList vec
```
harendra-kumar, line 1028: You can just use `ex == minimum l` here. Also, the general convention to decompose a list is `x:xs` so you can use `x` instead of `ex`. Also the result value can be named as `r`, `res` or `result`.

test/Prop.hs
```diff
@@ -990,6 +1000,55 @@ monadBind constr eq t (a, b) = withMaxSuccess maxTestCount $
...
+                assert $ a == True
+
+
+-- to check if the [Int] is non-increasing
+decreasing :: Int -> [Maybe Int] -> Bool
+decreasing _ [] = True
+decreasing mx (x:xl) = case x of
+                           Just y -> if mx < y then False else decreasing y xl
+                           Nothing -> decreasing mx xl
```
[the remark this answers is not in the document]
harendra-kumar: How about something like: `all (== True) $ zipWith (<) (tail xs) xs`?

test/Prop.hs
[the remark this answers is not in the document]
wildcat26: okay. thanks :+1:

wildcat26: By result, do you mean `a` which should be `True`?

harendra-kumar: yes

test/Prop.hs
```diff
@@ -990,6 +996,49 @@ monadBind constr eq t (a, b) = withMaxSuccess maxTestCount $
         let list = a >>= \x -> (+ x) <$> b
         listEquals eq stream list
 
+-------------------------------------------------------------------------------
+-- tests for S.groupsBy
+-------------------------------------------------------------------------------
+
+
+-- if list is empty, returns Nothing
```
harendra-kumar, line 1004: This does not need a doc. If doc should be written start with haddock markup (`-- |`). Left justify all your doc wrapping the lines at 80 columns. You can use some editor command to do that automatically. Start sentences with capital letters and end with a full stop. This applies to all the docs in this PR.

harendra-kumar:
Please the squash the commits into a single commit.

test/Prop.hs
```diff
@@ -990,6 +993,56 @@ monadBind constr eq t (a, b) = withMaxSuccess maxTestCount $
...
+
+-- |
+-- If the list is empty, returns Nothing,
+-- else wraps the minimum value of the list in Just.
+maybeMinimum :: [Int] -> Maybe Int
+maybeMinimum [] = Nothing
+maybeMinimum ls = Just $ minimum ls
+
+
+-- |
```
harendra-kumar, line 1008: See https://github.com/composewell/haskell-dev/blob/master/coding-style.rst#haddock for how to write the haddock comments. We do not use `-- |`` on a line of its own.

wildcat26: oh but you merged this ?

harendra-kumar: That does not stop the merge but something that you should keep in mind in the next commit.

PR #601, opened 2020-07-16
Add Data.Prim.Pinned.Array and extend Data.Prim.Array

adithyaov:


.hlint.ignore
```diff
@@ -16,3 +16,6 @@ src/Streamly/Internal/Data/Prim/Array/Types.hs
 src/Streamly/Internal/Data/SmallArray/Types.hs
 src/Streamly/Internal/Data/Unicode/Stream.hs
 src/Streamly/Internal/Mutable/Prim/Var.hs
+src/Streamly/Internal/Data/Array/Prim/Types.hs
```
harendra-kumar, line 19: we can remove `src/Streamly/Internal/Data/Prim/Array/Types.hs` from this list.

PR #611, opened 2020-07-24
Break down properties test suite into smaller ones.

pranaysashank:


pranaysashank:
One potential drawback I see is that the modules `Streamly.Test.Common` and `Streamly.Test.Prelude` get built for each test suite. It has its uses in that we can use separate compilation options for each test suite, if we use a separate package with common modules refactored into a library like we did for benchmarks we lose this option.

harendra-kumar:
We will have to merge `test/Streamly/Test/Internal/Prelude.hs` as well in `test/Streamly/Test/Prelude...`

test/Prop.hs
```diff
@@ -961,88 +259,10 @@ zipAsyncMonadic constr eq t (a, b) = withMaxSuccess maxTestCount $
...
-    monadicIO $ do
-        stream <-
-            run
-                ((S.toList . t)
-                     (constr a >>= \x -> (+ x) <$> constr b))
-        let list = a >>= \x -> (+ x) <$> b
-        listEquals eq stream list
-
 -------------------------------------------------------------------------------
 -- Tests for S.groupsBy
```
adithyaov: Do you want to keep this header here?

test/Prop.hs
```diff
@@ -493,7 +433,7 @@ main = hspec
         forOps (mkOps parallely) $ concurrentOps folded "parallely folded" sortEq
 
     describe "Concurrent application" $ do
-        serialOps $ prop "serial" . concurrentApplication (==)
+--        serialOps $ prop "serial" . concurrentApplication (==)
```
adithyaov: Should this  be moved to ` test/Streamly/Test/Prelude/Serial.hs`?

test/Prop.hs
```diff
@@ -393,7 +350,7 @@ main = hspec
     describe "Concurrent application" $ do
 --        serialOps $ prop "serial" . concurrentApplication (==)
         asyncOps $ prop "async" . concurrentApplication sortEq
-        aheadOps $ prop "ahead" . concurrentApplication (==)
+--        aheadOps $ prop "ahead" . concurrentApplication (==)
```
adithyaov: Are we discarding this benchmark?

cabal.project.coverage
```diff
@@ -4,7 +4,7 @@ packages:
 source-repository-package
   type: git
   location: https://github.com/composewell/hpc-coveralls
-  tag: d05f0c68aabec6120f8a96a9b5181b21a8f364df
+  tag: 93f3dcd473b3f92bbb40b9d3db9c83f4bf085cff
```
adithyaov, line 7: Update the commit to the latest version/tag instead of using the master?

test/Prop.hs
```diff
@@ -349,7 +332,7 @@ main = hspec
 
     describe "Concurrent application" $ do
 --        serialOps $ prop "serial" . concurrentApplication (==)
-        asyncOps $ prop "async" . concurrentApplication sortEq
+--        asyncOps $ prop "async" . concurrentApplication sortEq
```
adithyaov: Discarded?

adithyaov:
LGTM. I've crossed checked all the benchmarks. All the benchmarks that have been removed seem to be added elsewhere.

cabal.project.coverage
adithyaov: Nvm, We are not releasing a new version and even if we are `cabal.project.coverage` is not a part of sdist.

pranaysashank:
> LGTM. I've crossed checked all the benchmarks. All the benchmarks that have been removed seem to be added elsewhere.

\* tests :)

cabal.project.coverage
pranaysashank: This commit is not on master, it's on a different branch. cabal occasionally runs into weird problems when I use branch or tag name instead of the commit id.

PR #615, opened 2020-07-26
Alternative end of input issue

### LINKED ISSUE #604: Bug in alternative instance of ParserD
ParserD's `<|>` is not trying the other parser in case the first parser fails in the `extract` function, this behaviour is different from ParserK's `<|>` implementation.

harendra-kumar:
Fixes #604

### DIFF
--- a/src/Streamly/Internal/Data/Parser/ParserD/Types.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserD/Types.hs
@@ -392,7 +392,7 @@ data AltParseState sl sr = AltParseL Int sl | AltParseR sr
 -- /Internal/
 --
 {-# INLINE alt #-}
-alt :: Monad m => Parser m x a -> Parser m x a -> Parser m x a
+alt :: MonadCatch m => Parser m x a -> Parser m x a -> Parser m x a
 alt (Parser stepL initialL extractL) (Parser stepR initialR extractR) =
     Parser step initial extract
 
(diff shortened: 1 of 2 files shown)


pranaysashank:
I think it should feed the entire input to the alternate parser when the first one fails at EOF like attoparsec here.

```haskell
{-# LANGUAGE OverloadedStrings #-}

import Control.Applicative
import Data.Attoparsec.ByteString as B

import qualified Streamly.Internal.Prelude as IP
import qualified Streamly.Internal.Data.Parser.ParserD as P
import qualified Streamly.Data.Fold as FL

main = do
   print (parseOnly (B.take 2 <|> B.take 1) "3")
   print =<< IP.parseD ((P.takeEQ 2 FL.toList) <|> (P.takeEQ 1 FL.toList)) (IP.fromList ([51]::[Int]))
```

```bash
$ ./PR615
Right "3"
PR615: ParseError "takeEQ: Expecting exactly 1 elements, got 0"
$
```

harendra-kumar:
This will require a bigger change. We may have to resurrect the original strategy in which we returned error from extract instead of using exceptions. extract should return the `Step` type. It will make some combinators more complicated.

pranaysashank:
Maybe we should not provide alternative instance for `ParserD`

src/Streamly/Internal/Data/Parser/ParserD/Types.hs
```diff
@@ -426,7 +426,11 @@ alt (Parser stepL initialL extractL) (Parser stepR initialR extractR) =
             Error err -> Error err
 
     extract (AltParseR sR) = extractR sR
-    extract (AltParseL _ sL) = extractL sL
+    extract (AltParseL _ sL) = do
+        r <- try $ extractL sL
+        case r of
+            Left (_ :: ParseError) -> initialR >>= extractR
```
adithyaov, line 432: This will just create more confusion in my opinion. Until we find a proper solution, we might as well use `extractL sL`  and note that `alt` does not work for parsers that fail without sufficient input. Or as @pranaysashank suggests, remove the `alt` instance for now.

adithyaov:
A pretty straightforward but not a good solution would be to provide a consecutive `alt` instead of a sequential one.

harendra-kumar:
Closing in favor of #1782 .

PR #624, opened 2020-08-07
hie.yaml: Add a hie.yaml file for Haskell Language Server

SanchayanMaity:
This is helpful for Haskell Language Server to detect project
configuration and set up GHC.

Generated using gen-hie from the implicit-hie project.

Signed-off-by: Sanchayan Maity <maitysanchayan@gmail.com>

### DIFF
--- a/hie.yaml
+++ b/hie.yaml
@@ -1,0 +1,136 @@
+cradle:
+  cabal:
+    - path: "./"
+      component: "streamly:bench:/*.hs"
+
+    - path: "src"
+      component: "lib:streamly"
+
+    - path: "test"
+      component: "streamly:test:test"
+
+    - path: "test"
+      component: "streamly:test:internal-prelude-test"
+
+    - path: "test"
+      component: "streamly:test:pure-streams-base"
+
+    - path: "test"
+      component: "streamly:test:pure-streams-streamly"
+
+    - path: "./"
+      component: "streamly:test:Prelude.Serial"
+
+    - path: "./"
+      component: "streamly:test:Prelude.WSerial"
+
+    - path: "./"
+      component: "streamly:test:Prelude.ZipSerial"
+
+    - path: "./"
+      component: "streamly:test:Prelude.ZipAsync"
+
+    - path: "./"
+      component: "streamly:test:Prelude.Ahead"
+
+    - path: "./"
+      component: "streamly:test:Prelude.Async"
+
+    - path: "./"
+      component: "streamly:test:Prelude.WAsync"
+
+    - path: "./"
+      component: "streamly:test:Prelude.Parallel"
+
+    - path: "./"
+      component: "streamly:test:Prelude.Concurrent"
+
+    - path: "test"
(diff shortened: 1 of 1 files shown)


pranaysashank:
The generated file contains wrong paths for the test components Prelude.*.  Also does the component listed for benchmark files actually work?

harendra-kumar:
@SanchayanMaity thanks for the PR. Are you planning to make the test and benchmark components work?

harendra-kumar:
I see this error when running `ghcide`:

```
ghcide: CradleError (ExitFailure 1) ["Failed to parse result of calling cabal","","cabal: Unknown target 'streamly:bench:/*.hs'.\nThe package streamly has no component 'bench'.\n\n\n",""]
```

SanchayanMaity:
@harendra-kumar Yes, I want to make the test and benchmark components work. Yet to find a solution. I have not tested with `ghcide`, was using `Haskell Language Server`. But, I guess HLS does not work because `ghcide` doesn't.

harendra-kumar:
I tried running ghcide but could not get past this error:

```
Step 4/6, Cradle 1/1: Loading GHC Session
ghcide: CradleError ExitSuccess ["Multi Cradle: No prefixes matched","pwd: /Users/vol/vemba/composewell-tech/Eng/projects/streamly/streamly1","filepath: /Users/vol/vemba/composewell-tech/Eng/projects/streamly/streamly1","prefixes:","(\"./src\",Cabal {component = Just \"streamly:lib:streamly\"})"]
```

I tried writing `hie.yaml` in the same way as in the ghcide repo.

harendra-kumar:
I cannot find documentation on how to write a working `hie.yaml`. That's basic. Maybe I did not dig carefully.

pranaysashank:
@harendra-kumar https://github.com/mpickering/hie-bios/ . I can share mine, I think I should have it from long ago

pranaysashank:
`ghcide` used to be behind the times with hie-bios, I used to just patch the latest version in. But I stopped using both the language servers long ago

pranaysashank:
This is my config from long ago, hopefully it still works.
```cabal
cradle:
  cabal:
    - path: "./src"
      component: "lib:streamly"
    - path: "./examples/ListDir.hs"
      component: "exe:ListDir"
    - path: "./test/Prop.hs"
      component: "test:properties"

dependencies:
  - streamly.cabal
```

harendra-kumar:
It did not work. ghcide gives the same error. I tried without an hie.yaml and in that case I think it fails because of CPP macros. maybe we need to enable the extensions explicitly. Does it look at the cabal file at all to get the build information?

It may not be worth spending time, it seems immature as of now.

pranaysashank:
We can wait until the integration work is complete https://github.com/haskell/haskell-language-server. But I didn't have any problems haskell-ide-engine (other than template-haskell), I don't know about ghcide though.

pranaysashank:
#732 adds a hie.yaml file.

PR #627, opened 2020-08-08
Add a cleanup action to Fold

### LINKED ISSUE #543: Add a `done` action in the Fold type
#531 added exception handling to folds, however it has an issue. The `after`, `finally` and `bracket` functions run the release action on `extract`. But extract can be called over and over again during a scan. Its not right if we release a resource during `extract`.

One possible improvement to this would be when we have terminating folds, we can run the release action on `Stop`. However when the stream ends there is no way to communicate to the fold to do any cleanup. To be able to do that we need a `done` action in the fold which will always be called by the fold driver when the stream ends. This can help the fold perform any final cleanup.

Also, see the `scanlMAfter'` combinator introduced by #542 . It also introduces a similar `done` action for the scan.

adithyaov:
Closes #543

### DIFF
--- a/.packcheck.ignore
+++ b/.packcheck.ignore
@@ -1,2 +1,3 @@
 .packcheck.ignore
 cabal.project.coverage
+src/Streamly/Internal/Data/Prim/Array.hs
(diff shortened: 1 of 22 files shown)


adithyaov:
I don't know how I can continue with this.

PR #629, opened 2020-08-11
Drop support for GHC 7.10.3.

pranaysashank:


.travis.yml
```diff
@@ -101,10 +101,10 @@ matrix:
     os: linux
     arch: amd64
 
-  - env: BUILD=cabal-v2 GHCVER=7.10.3 GHC_OPTIONS="" CABAL_BUILD_OPTIONS="--flags examples-sdl"
-    addons: {apt: {packages: [cabal-install-2.4,ghc-7.10.3,libsdl1.2-dev], sources: [hvr-ghc]}}
-    os: linux
-    arch: amd64
+  #- env: BUILD=cabal-v2 GHCVER=7.10.3 GHC_OPTIONS="" CABAL_BUILD_OPTIONS="--flags examples-sdl"
```
harendra-kumar, line 104: why not completely remove it?

streamly.cabal
```diff
@@ -374,6 +373,7 @@ library
 
                     -- IO devices
                      , Streamly.Memory.Array
+                     , Streamly.Data.Prim.Array
```
harendra-kumar, line 376: This should go on `other-modules`.

streamly.cabal
```diff
@@ -1072,7 +1059,7 @@ executable Split
     buildable: True
     build-Depends:
         streamly
-      , base         >= 4.8   && < 5
+      , base         >= 4.9   && < 5
       , mtl          >= 2.2   && < 3
     if impl(ghc < 8.0)
```
harendra-kumar, line 1064: This can be removed now.

adithyaov:
This https://github.com/composewell/streamly/pull/565/files can be reverted?
Maybe grep for code that distingushes ghc version? There might be a few more cases.

pranaysashank:
@adithyaov I don't see why `NFData` instance should be removed?

adithyaov:
I think I just confused Ghc's version and Primitive's version. My bad. I'd like to retract that comment.

.travis.yml
pranaysashank: In other CI configurations, these are commented. So I just decided to comment it here as well, no particular reason other than that.

PR #633, opened 2020-08-14
Move fromAddr# and fromString# to the immutable Memory.Array module

### LINKED ISSUE #631: Move fromAddr#, fromString# to immutable Array module
These have been moved to `src/Streamly/Internal/Memory/Mutable/Array/Types.hs` which not correct, these are immutable arrays. Also export them from the Internal.Memory.Array module. I need this in the inotify work.

adithyaov:
Closes #631

### DIFF
--- a/src/Streamly/Internal/Memory/Array/Types.hs
+++ b/src/Streamly/Internal/Memory/Array/Types.hs
@@ -21,6 +21,8 @@ module Streamly.Internal.Memory.Array.Types
     -- * Construction
     , spliceTwo
 
+    , fromAddr#
+    , fromString#
     , fromList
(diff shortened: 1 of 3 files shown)


src/Streamly/Internal/Memory/Array/Types.hs
```diff
@@ -21,6 +21,8 @@ module Streamly.Internal.Memory.Array.Types
     -- * Construction
     , spliceTwo
 
+    , fromAddr#
+    , fromString#
```
harendra-kumar, line 25: Export these from `Internal.Memory.Array` as well.

PR #639, opened 2020-08-18
Make the finalizer based exception handling routines the default.

pranaysashank:
I made the routines that end with `IO` the default ones and renamed the default exception handling routines to have a `_` at the end. Let me know if this naming seems fine or any other suggestions.


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3392,22 +3392,22 @@ _bracket bef aft bet = Stream step' BracketInit
...
 -- finally action xs = after action $ onException action xs
-finally action xs = bracket (return ()) (\_ -> action) (const xs)
+finally_ action xs = bracket_ (return ()) (\_ -> action) (const xs)
 
-{-# INLINE finallyIO #-}
-finallyIO :: (MonadAsync m, MonadCatch m) => m b -> Stream m a -> Stream m a
-finallyIO action xs = bracketIO (return ()) (\_ -> action) (const xs)
+{-# INLINE finally #-}
+finally :: (MonadAsync m, MonadCatch m) => m b -> Stream m a -> Stream m a
+finally action xs = bracket (return ()) (\_ -> action) (const xs)
```
pranaysashank, line 3402: Do we need to use `MonadAsync`? `MonadCatch` implies `MonadThrow` but its included in `MonadAsync` again, probably fine. This happens in quite a few places btw.

PR #643, opened 2020-08-19
Improve error handling in decodeUtf8

harendra-kumar:


pranaysashank (approved):
Did you test if this makes sense on a failing input?

harendra-kumar:
It does not make much sense to someone who does not know the internals of the decoder. But I could not find a way to make it sensible to normal users. So at least they will know the failing byte and if one knows the internals we can make a bit more sense out of state and codepoint built till now.

PR #650, opened 2020-08-22
Fix exception handling atomicity wrt async exceptions

harendra-kumar:


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -3169,38 +3169,72 @@ gbracket_ bef exc aft fexc fnormal =
...
+            -- finalizing action will never be run. We need to do this
+            -- atomically wrt async exceptions.
+            mask_ $ do
+                writeIORef ref Nothing
+                action
+
+-- | Run an action clearing the finalizer IORef atomically wrt async
+-- exceptions. The action is run with async exceptions masked.
+withIORefFinalizer :: MonadBaseControl IO m
+    => IORef (Maybe (IO ())) -> m a -> m a
```
pranaysashank, line 3230: Is it not possible to do a `withIORefFinalizer :: IORef (Maybe (IO ())) -> Stream m a -> Stream m a`?

pranaysashank (approved):

src/Streamly/Internal/Data/Stream/StreamD.hs
harendra-kumar: Not sure.

PR #654, opened 2020-08-23
Pass the failed Stream as an argument to ghandle, gbracket 

pranaysashank:
@harendra-kumar I am unsure what the type signature of `gbracket` for `Unfold`s should be.


harendra-kumar:
Can you rebase this? Its showing changes from the other merged PR as well.

pranaysashank:
Rebased.

harendra-kumar (approved):
Let's commit this and keep the issue open for Unfold case.

For unfolds I guess we can make an Unfold from the current state at the time when the exception occurred and pass that Unfold as an argument to the handler unfold. The handler unfold can choose to use the unfold we passed.

PR #657, opened 2020-08-23
Deprecate Streamly module.

pranaysashank:


harendra-kumar:
We can create wrappers and deprecate the wrappers. I did this for foldWith etc. earlier. You can see this commented out code in the Streamly module:

```
-- | Same as "Streamly.Prelude.foldWith".
--
{-# DEPRECATED foldWith "Please use Streamly.Prelude.foldWith instead." #-}
{-# INLINABLE foldWith #-}
foldWith :: (IsStream t, Foldable f)
    => (t m a -> t m a -> t m a) -> f (t m a) -> t m a
foldWith = P.foldWith
```

This can be uncommented now.

We can create a C pre-processor macro to generate this kind of code for all the deprecated functions. We can omit the signatures and disable the warning.

src/Streamly/Prelude.hs
```diff
@@ -477,6 +498,12 @@ module Streamly.Prelude
     -- -------Stream m a------|
     -- @
     --
+    , serial
+    , wSerial
+    , ahead
+    , async
+    , wAsync
+    , parallel
```
harendra-kumar, line 506: `serial` and `ahead` should go in the appending section above. `wSerial` should go in a separate `Interleaving` section - see `IsStream` module, organize in the same way.

src/Streamly/Prelude.hs
```diff
@@ -802,3 +840,74 @@ import Streamly.Internal.Data.Stream.IsStream
...
+-- application with or without the '$' operator, or with reverse function
+-- application operator '&'. Streamly provides concurrent versions of these
+-- operators applying stream processing functions such that each stage of the
+-- stream can run in parallel. The operators start with a @|@; we can read '|$'
+-- as "@parallel dollar@" to remember that @|@ comes before '$'.
+--
+-- Imports for the code snippets below:
+--
+-- @
+--  import Streamly
```
harendra-kumar, line 856: remove this, and use `S.drain` in the examples.

src/Streamly/Prelude.hs
```diff
@@ -802,3 +840,74 @@ import Streamly.Internal.Data.Stream.IsStream
...
+--  import qualified Streamly.Prelude as S
+--  import Control.Concurrent
+-- @
+
+-- $sum
```
harendra-kumar, line 861: This is not referenced anywhere.

src/Streamly.hs
```diff
@@ -115,45 +117,55 @@ module Streamly
...
-    -- * Concurrency Control
-    -- $concurrency
+    -- ** Concurrency Control
     , maxThreads
     , maxBuffer
 
-    -- * Rate Limiting
-    , Rate (..)
+    -- ** Rate Limiting
+    , Rate
```
harendra-kumar, line 160: We are not exposing the constructors anymore? This can break existing code.

src/Streamly.hs
```diff
@@ -115,45 +117,55 @@ module Streamly
...
     , avgRate
     , minRate
     , maxRate
     , constRate
 
-    -- * Stream Type Adapters
-    -- $adapters
-    , IsStream ()
+    -- ** Stream Type Adapters
+    , IsStream
```
harendra-kumar, line 168: We do not want to expose the type class methods.

harendra-kumar (changes requested):
1) Need to replace the names (foldWith etc) in the Tutorial module as well.
2) Need to update the names in the README as well
3) All the operations that appeared in the Streamly.Prelude module for the first time need to have a `since 0.8.0` annotation even though they were present in other modules earlier. We can keep the previous since notation as well with a annotation saying "in module xyz". You can see previous of examples of similar cases and how they are annotated.

src/Streamly.hs
pranaysashank: No idea how to deprecate constructors

pranaysashank: Existing code will break even without this change.

harendra-kumar: Yes, but least breakage is preferred. Not a big deal though.

src/Streamly.hs
pranaysashank: In a small example I tested, it doesn't expose the typeclass methods.

src/Streamly.hs
```diff
@@ -218,20 +211,32 @@ module Streamly
...
 
 import qualified Streamly.Prelude as P
-import qualified Streamly.Internal.Data.Stream.IsStream as IP
 import qualified Streamly.Internal.Data.Stream.StreamK as K
 import qualified Streamly.Internal.Data.Stream.Async as Async
 
+#define GG x=3
+
+#define DEPRECATE_FUNCTION(fn)                                     \
+{-# DEPRECATED fn "Please use 'Streamly.Prelude.fn' instead." #-}; \
```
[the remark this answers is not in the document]
pranaysashank: I don't understand what you're saying here, haddock renders them correctly. Do you mean the case where I pass `fn` also in double quotes?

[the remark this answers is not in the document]
harendra-kumar: This is how haddock shows this:

> Deprecated: Please use fn instead.

harendra-kumar (commented):
Can you also make sure that there is no inconsistency in the haddock in the Streamly module after the deprecations, and the code examples are working correctly?

src/Streamly.hs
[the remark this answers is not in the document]
pranaysashank: Must be clang, I'll use a workaround.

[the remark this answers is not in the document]
harendra-kumar: Its not clang, `Language CPP` is GHC's c pre-processor. You can use external cpp program and use stringizing operator (`#`) but that may also not solve the problem, because after stringizing you will have to concatenate. I do not think it is easy to do, may be impossible.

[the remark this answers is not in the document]
pranaysashank: I don't there's any c pre-processor shipped with ghc that it uses. It just picks up the system's cpp AFAICT. There is a package called `cpphs`, and it looks like both cpphs and clang fail in presence of quoting.

src/Streamly.hs
```diff
@@ -218,20 +211,32 @@ module Streamly
...
 import Streamly.Internal.Data.Stream.Zip
+       ( ZipAsync, ZipAsyncM, ZipSerial, ZipSerialM, ZipStream
+       , zipping, zippingAsync)
 
 import qualified Streamly.Prelude as P
-import qualified Streamly.Internal.Data.Stream.IsStream as IP
 import qualified Streamly.Internal.Data.Stream.StreamK as K
 import qualified Streamly.Internal.Data.Stream.Async as Async
 
+#define GG x=3
```
pranaysashank, line 232: Extraneous line, should be deleted.

pranaysashank:
Streamly module is deprecated and the functions `foldWith`, `foldMapWith`, `forEachWith` have been deprecated and renamed.

src/Streamly/Internal/Data/SVar.hs
```diff
@@ -229,6 +229,8 @@ data WorkerInfo = WorkerInfo
     }
 
 
+-- /Since: 0.8.0 ("Streamly.Prelude")/
+--
```
harendra-kumar, line 233: We need to move this down in the haddock to make it visible in the docs. Also, we should have 0.8.0 as the primary since notation and use `since 0.5.0 (Streamly)` as a historical note. This needs to be updated for all rate operations and all other operations that have been moved.

src/Streamly/Tutorial.hs
```diff
@@ -1556,12 +1539,12 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 --  Interop with @vector@:
 --
 -- @
--- import Streamly
--- import qualified Streamly.Prelude as S
+-- import "Streamly" (SerialT)
```
harendra-kumar, line 1542: Streamly.Prelude?

src/Streamly/Tutorial.hs
```diff
@@ -1584,13 +1567,13 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 --  Interop with @pipes@:
 --
 -- @
--- import "Streamly"
+-- import "Streamly" (SerialT)
```
harendra-kumar, line 1570: Streamly.Prelude?

src/Streamly/Tutorial.hs
```diff
@@ -1611,7 +1594,7 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- Interop with @streaming@:
 --
 -- @
--- import "Streamly"
+-- import "Streamly" (SerialT, MonadAsync)
```
harendra-kumar, line 1597: Streamly.Prelude?

src/Streamly/Tutorial.hs
```diff
@@ -1635,7 +1618,7 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- Interop with @conduit@:
 --
 -- @
--- import "Streamly"
+-- import "Streamly" (SerialT)
```
harendra-kumar, line 1621: Streamly.Prelude?

harendra-kumar (changes requested):
In README we have to prefix all occurrences of `asyncly` etc with `S.` or import them explicitly.

src/Streamly/Internal/Data/SVar.hs
pranaysashank: Should historical note be in the haddock?

harendra-kumar: Yes, it should be something like this:

```
@since 0.8.0

/Since 0.1.0 (in Streamly module)/
```

so that users know that the API has been there in the library but was moved in 0.8.0.

src/Streamly/Internal/Data/SVar.hs
```diff
@@ -897,11 +898,15 @@ withDiagMVar sv label action =
 -- Spawning threads
 ------------------------------------------------------------------------------
 
+-- /Since: 0.8.0 ("Streamly.Prelude")/
```
harendra-kumar, line 901: remove this.

src/Streamly/Internal/Data/Stream/Async.hs
```diff
@@ -513,6 +513,8 @@ mkAsyncD m = D.Stream step Nothing
             D.Skip s    -> D.Skip (Just $ D.Stream step1 s)
             D.Stop      -> D.Stop
 
+-- /Since: 0.8.0 ("Streamly.Prelude")/
```
harendra-kumar, line 516: remove this.

harendra-kumar (changes requested):
Need a consistent ordering when we have multiple since notations. In some cases we have higher release first (pre-existing cases e.g. foldrM/iterateM), in the newer ones we have mostly lower version first but in some cases there is no ordering as well (e.g. ParallelT).

harendra-kumar (approved):

PR #663, opened 2020-08-24
Rename Internal.Mutable.Prim.Var to Internal.Data.IORef.Prim.

### LINKED ISSUE #619: Rename Streamly.Internal.Mutable.Prim.Var
Move `Streamly/Internal/Mutable/Prim/Var.hs` => Streamly.Internal.Data.IORef.Prim

We can make the functions in this module as IO functions and name them after IORef functions. This would make it easier to understand and use for those who already know IORef, so this just becomes an IORef specialized to Prim.

pranaysashank:
- Rename function `fromPrimVar` to `fromPrimIORef`.

Closes #619


harendra-kumar (commented):
1. Why isn't Prim.hs file showing as moved? Did you use `git mv` or copied it?

harendra-kumar:
Need to change the filename in `.hlint.ignore` as well.

pranaysashank:
It doesn't matter whether we use `git mv` or not. It works sometimes and doesn't work other times. But no I didn't use `git mv`.

PR #673, opened 2020-08-27
Provide a build flag to use C malloc for pinned arrays instead of using GHC malloc

### LINKED ISSUE #644: Provide a build flag to use C malloc for pinned arrays instead of using GHC malloc
src/Streamly/Memory/Malloc.hs has a C preprocessor macro to control whether to use GHC malloc or C malloc. We should expose this via a build flag so that users can try this out in case fragmentation becomes an issue due to GHC malloc.

Also, we should move src/Streamly/Memory/Malloc.hs to Streamly/Internal.

adithyaov:
Closes #644

### DIFF
--- a/streamly.cabal
+++ b/streamly.cabal
@@ -360,7 +360,6 @@ library
                      , Streamly.Data.SmallArray
 
                     -- Memory storage
-                       Streamly.Memory.Malloc
                      , Streamly.Memory.Ring
 
                      , Streamly.FileSystem.IOVec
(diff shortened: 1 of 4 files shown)


streamly.cabal
```diff
@@ -366,7 +373,10 @@ library
                      , Streamly.Data.Unfold
                      , Streamly.Data.Unicode.Stream
 
-                    -- IO devices
+                     -- Memory storage
+                     ,  Streamly.Internal.Memory.Malloc
```
harendra-kumar, line 377: Should we call the `Streamly.Internal.Foreign.Malloc`, it exports only Foreign ptr malloc. The `Memory` hierarchy anyway goes away with the Array module renaming. Also this would be consistent with the Foreign module in base.

harendra-kumar (changes requested):

harendra-kumar (approved):

Issue #653, opened 2020-08-22
Move Memory.Array to Data.Array.Storable.Pinned


harendra-kumar:
As per this plan: https://github.com/composewell/streamly/blob/module-org/design/module-organization.md .

Let's deprecate the old module and start using the new namespace.

harendra-kumar:
Or should we call it `Data.Array.Foreign` instead? The fact it has the ability to use `Foreign` memory/`ForeignPtr` is perhaps more important than `Storable`. Foreign also automatically implies `Storable`, it also implies that the memory is pinned so we do not need the `Pinned` suffix.

PR #674, opened 2020-08-27
Rename Unicode and Memory modules according the defined module structure

### LINKED ISSUE #652: Move Streamly.Data.Unicode to Streamly.Unicode
Like FileSystem/Network, `Unicode` can be a separate namespace directly under Streamly instead of under Data. This is also proposed in this: https://github.com/composewell/streamly/blob/module-org/design/module-organization.md .

We can deprecate the current Streamly.Data.Unicode and move it under streamly.

adithyaov:
CONTAINS BREAKING CHANGES

Closes #652 and #653


adithyaov:
@harendra-kumar Please review & merge this once the CI passes.

adithyaov:
We can remove `ArrayStream`.

harendra-kumar (commented):
Looks good except that we may want to change the array module to `Data.Array.Foreign` instead.

harendra-kumar (approved):
For unicode arrays (or unicode strings) we perhaps do not care about Storable/Foreign in the name.  Chars are anyway Storable, and we do not care if the array is foreign or not.

PR #687, opened 2020-09-06
Add Windows file system event handling

rnjtranjan:


src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,172 @@
+{-# LANGUAGE ForeignFunctionInterface #-}
+{-# LANGUAGE CPP #-}
+
+module Streamly.Internal.FileSystem.Event.Windows(watchTrees, Event, showEvent)
```
adithyaov, line 4: Maybe rewrite this like,
```
module Streamly.Internal.Data.Stream
    ( watchTrees
    , Event
    , showEvent
    )
where
```

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,172 @@
+{-# LANGUAGE ForeignFunctionInterface #-}
+{-# LANGUAGE CPP #-}
```
adithyaov, line 2: `CPP` is enabled by default I believe. You can probably remove this.

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,172 @@
...
+import Streamly.Internal.Data.Parser (Parser)
+import Streamly.Internal.Data.Array.Storable.Foreign.Types (Array(..))
+
+import qualified Streamly.Internal.Data.Parser as PR
+import qualified Streamly.Internal.Data.Array.Storable.Foreign as A
+import qualified Streamly.Internal.Data.Stream.IsStream as S
+
+type Handle = HANDLE
+
+watchTrees :: [FilePath] -> SerialT IO Event
```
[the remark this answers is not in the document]
harendra-kumar: Windows provides a flag to watch either the entire subtree or just the directory. We can use that flag to implement the `watchPaths` API. See https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-readdirectorychangesw .

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,172 @@
...
+
+foreign import ccall safe "windows.h ReadDirectoryChangesW"
+  c_ReadDirectoryChangesW :: Handle -> LPVOID -> DWORD -> BOOL -> DWORD -> LPDWORD -> LPOVERLAPPED -> LPOVERLAPPED_COMPLETION_ROUTINE -> IO BOOL
+
+readOneEvent :: String -> Parser IO Word8 [Event]
+readOneEvent root = do
+  let headerLen = 8*1024
+  arr <- PR.take headerLen (A.writeN headerLen)  
+  fnis <- PR.yieldM $ A.asPtr arr $ \buf -> readChangeEvents buf root
+  return $ fmap maptoEvent fnis
```
harendra-kumar, line 118: We get an Array from `readDirectoryChanges`, we convert that into a stream and then we use the parser to convert it back into an Array. Instead doing this we can directly use the Array returned by readDirectoryChanges and apply `readChangeEvents` on that.

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,172 @@
...
+                         eventRootPath = snd fni
+                         }                                            
+
+readChangeEvents :: Ptr FILE_NOTIFY_INFORMATION -> String -> IO [(FILE_NOTIFY_INFORMATION, String)]
+readChangeEvents pfni root = do  
+  fni <- peekFNI pfni
+  let entry = (fni, root)
+      nioff = fromEnum $ fniNextEntryOffset fni   
+  entries <- if nioff == 0 then return [] else readChangeEvents  (pfni `plusPtr` nioff)  root
+  return $ entry:entries
```
harendra-kumar, line 132: We need to handle the case when an overflow occurred. See the Remarks section in the windows API doc.

streamly.cabal
```diff
@@ -223,6 +223,9 @@ common compile-options
     if os(linux)
       cpp-options:    -DCABAL_OS_LINUX
 
+    if os(mingw32)
```
harendra-kumar, line 226: Can't we use `os(windows)` here as we are using down below.

harendra-kumar (changes requested):
Take a look at this https://github.com/composewell/haskell-dev/blob/master/coding-style.rst for style guidelines.

test/Streamly/Test/FileSystem/Event.hs
```diff
@@ -8,54 +8,63 @@
...
 #endif
 
+#if defined(CABAL_OS_WINDOWS)
+main :: IO ()
+main = do
+    args <- getArgs   
+    watchPaths (NonEmpty.fromList args) 
+        & Stream.mapM_ (putStrLn . Event.showEvent)
+
+#else
```
harendra-kumar, line 63: Instead of using the whole code under `CABAL_OS_WINDOWS` we can just put the conditional around the `paths <- mapM toUtf8 args` in the code below.

harendra-kumar (commented):
This is looking good. I have mostly formatting related comments.

Thanks and great job for first Haskell PR!

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,432 @@
...
+    fni <- peekFNI pfni
+    let entry = Event{  eventFlags = fniAction fni 
+                      , eventRelPath = fniFileName fni  
+                      , eventRootPath = root
+                      , totalBytes = bytesRet
+                      }  
+        nioff = fromEnum $ fniNextEntryOffset fni   
+    entries <- if nioff == 0 
+                    then return [] 
+                    else readChangeEvents (pfni `plusPtr` nioff) root bytesRet
```
harendra-kumar, line 263: The formatting of `Event` struct and `if` clause is missed. I gave examples about how to format it in the previous comments.

test/Streamly/Test/FileSystem/Event.hs
```diff
@@ -8,54 +8,48 @@
...
-
-#if defined(CABAL_OS_DARWIN)
-watchPaths = Event.watchTrees
-#elif defined(CABAL_OS_LINUX)
+#if defined(CABAL_OS_LINUX)
+watchPaths :: NonEmpty (Array Word8) -> SerialT IO Event.Event    
 watchPaths = Event.watchPaths
 #else
-#error "Unsupported platform
+watchPaths :: NonEmpty (Array Word8) -> SerialT IO Event.Event
```
harendra-kumar, line 46: The signature can be factored outside the conditional macro.

src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -0,0 +1,447 @@
...
+--
+watchPathsWith :: 
+       (Config -> Config) 
+    -> NonEmpty (Array Word8) 
+    -> SerialT IO Event
+watchPathsWith f paths = do
+    let cfg = f $ setRecursiveMode False defaultConfig
+        sth = pathsToHandles (utf8ToStringList paths) cfg        
+    S.after (closePathHandleStream sth) 
+        $ S.concatMapWith parallel eventStreamAggr sth
```
harendra-kumar, line 358: BUG: the stream `sth` is being evaluated multiple times. `after` and `concatMapWith` will independently evaluate the stream. We are closing different handles in `after` rather than the ones being used in `concatMapWith`.

Instead, we can use a stream of paths as input to `concatMapWith`, in eventStreamAggr generate a stream from the path it receives, it can use a `bracket` around it to open the path before and close it after the stream is done.

We have the same issue in watchTrees as well.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #691, opened 2020-09-14
New encoding/decoding routines.

### LINKED ISSUE #658: Use invalid character replacement as the default in encode/decodeUtf8
Our default encoding/decoding routines are partial, they throw errors when an invalid condition is encountered in the stream. This is a good idea to always let the user know about the error and let them use a lax version of the API which replaces the invalid chars instead of throwing errors.

More often than not, over a period of time users run into encoding/decoding errors and replace these with the lax versions. Also, the problem is that in casual testing they may not run into any crash due to this, it is relatively rare and therefore likely to come as a surprise at a later time. 

Therefore, I propose that we use the lax behavior for the default APIs i.e. `decodeUtf8` and `encodeUtf8` and use stricter versions `decodeUtf8'` and `encodeUtf8'` to enforce strict error behavior. Also we can use `decodeUtf8_` and `encodeUtf8_` for dropping the invalid chars instead of replacing them.

pranaysashank:
Add `encodeUtf8_`, `encodeLatin1_`, `encodeUtf8'`, `encodeLatin1'`, `decodeUtf8_`, `decodeUtf8'`, `decodeUtf8Arrays'`, `decodeUtf8Arrays_`. 

See #658. 

Closes #658

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 ## Unreleased
 
+### Enhancements
+
+* New encoding/decoding routines, `encodeUtf8'`, `encodeLatin1'`, `decodeUtf8'`,
+  are added, these routines fail when they encounter any invalid characters.
+
 ### Breaking changes
 
 * Deprecate `Streamly.Memory.Array` in favor of `Streamly.Data.Array.Storable.Foreign`
@@ -14,6 +19,12 @@
   constraint.
 * Change the associativity of combinators `serial`, `wSerial`,
   `ahead`, `async`, `wAsync`, `parallel` to be the same as `<>`.
+* `encodeUtf8`, `decodeUtf8` now replace any invalid character encountered
+  during encoding/decoding with the Unicode replacement character. Use
+  `encodeUtf8'`, `decodeUtf8'` to recover the previous functionality.
+* `encodeLatin1` now silently truncates any character beyond 255 to incorrect
+  characters in the input stream. Use `encodeLatin1'` to recover previous
+  functionality.
 * Drop support for GHC 7.10.3.
 
 ### Bug Fixes
@@ -33,9 +44,13 @@
 * The `Streamly` module is now deprecated, its functionality is subsumed
   by `Streamly.Prelude`.
 * Some functions from `Streamly` module have been renamed in `Streamly.Prelude` module:
-    * `foldWith` has been replaced by `concatFoldableWith`
-    * `foldMapWith` has been replaced by `concatMapFoldableWith`
-    * `forEachWith` has been replaced by `concatForFoldableWith`
+    * `foldWith` to `concatFoldableWith`
+    * `foldMapWith` to `concatMapFoldableWith`
+    * `forEachWith` to `concatForFoldableWith`
+* The following encoding/decoding routines have been renamed:
+    * `encodeUtf8Lax` to `encodeUtf8`
+    * `encodeLatin1Lax` to `encodeLatin1`
+    * `decodeUtf8Lenient` to `decodeUtf8`
 
 ## 0.7.2
 
--- a/src/Streamly/Data/Unicode/Stream.hs
+++ b/src/Streamly/Data/Unicode/Stream.hs
@@ -33,7 +33,7 @@
 -- stream as @Array Char@ using the array 'Streamly.Data.Array.Storable.Foreign.write' fold.
 -- The 'Array' type provides a more compact representation and pinned memory
 -- reducing GC overhead. If space efficiency is a concern you can use
--- 'encodeUtf8' on the 'Char' stream before writing it to an 'Array' providing
+-- 'encodeUtf8'' on the 'Char' stream before writing it to an 'Array' providing
 -- an even more compact representation.
 --
 -- = String Literals
@@ -65,18 +65,15 @@
 -- lived strings in memory.
 --
 module Streamly.Data.Unicode.Stream
-    {-# DEPRECATED "Use Streamly.Unicode.Stream instead" #-}
+    {-# DEPRECATED "Use \"Streamly.Unicode.Stream\" instead" #-}
     (
     -- * Construction (Decoding)
       decodeLatin1
     , decodeUtf8
-    , decodeUtf8Lax
 
     -- * Elimination (Encoding)
     , encodeLatin1
-    , encodeLatin1Lax
     , encodeUtf8
-    , encodeUtf8Lax
     {-
     -- * Operations on character strings
     , strip -- (dropAround isSpace)
@@ -90,6 +87,11 @@ module Streamly.Data.Unicode.Stream
     -- , words
     -- , unlines
     -- , unwords
+
+    -- * Deprecations
+    , decodeUtf8Lax
+    , encodeLatin1Lax
+    , encodeUtf8Lax
     )
 where
 
--- a/src/Streamly/Internal/Unicode/Stream.hs
+++ b/src/Streamly/Internal/Unicode/Stream.hs
@@ -16,20 +16,24 @@ module Streamly.Internal.Unicode.Stream
     -- * Construction (Decoding)
       decodeLatin1
     , decodeUtf8
-    , decodeUtf8Lax
+    , decodeUtf8'
+    , decodeUtf8_
     , DecodeError(..)
     , DecodeState
     , CodePoint
     , decodeUtf8Either
     , resumeDecodeUtf8Either
     , decodeUtf8Arrays
-    , decodeUtf8ArraysLenient
+    , decodeUtf8Arrays'
+    , decodeUtf8Arrays_
 
     -- * Elimination (Encoding)
     , encodeLatin1
-    , encodeLatin1Lax
+    , encodeLatin1'
+    , encodeLatin1_
     , encodeUtf8
-    , encodeUtf8Lax
+    , encodeUtf8'
+    , encodeUtf8_
     {-
     -- * Operations on character strings
     , strip -- (dropAround isSpace)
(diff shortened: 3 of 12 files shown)


src/Streamly/Data/Unicode/Stream.hs
```diff
@@ -70,13 +70,13 @@ module Streamly.Data.Unicode.Stream
     -- * Construction (Decoding)
       decodeLatin1
     , decodeUtf8
-    , decodeUtf8Lax
```
harendra-kumar, line 73: This module is deprecated, do not change this. We only need to deprecate it. The same applies to `encodeLatin1Lax` and `encodeUtf8Lax` as well.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +630,57 @@ encodeLatin1 = S.map convert
...
 
--- | Like 'encodeLatin1' but silently truncates and maps input characters beyond
+-- | Like 'encodeLatin1'' but silently truncates and maps input characters beyond
 -- 255 to (incorrect) chars in 0-255 range. No error or exception is thrown
 -- when such truncation occurs.
 --
--- /Since: 0.7.0/
-{-# INLINE encodeLatin1Lax #-}
-encodeLatin1Lax :: (IsStream t, Monad m) => t m Char -> t m Word8
-encodeLatin1Lax = S.map (fromIntegral . ord)
```
harendra-kumar, line 616: We need to keep the old ones for deprecation.

src/Streamly/Data/Unicode/Stream.hs
harendra-kumar: Also, the behavior other routines in this module should perhaps remain the same as old.

harendra-kumar: If we do this then there won't be any breaking changes, only deprecations.

src/Streamly/Unicode/Stream.hs
```diff
@@ -69,13 +69,16 @@ module Streamly.Unicode.Stream
     -- * Construction (Decoding)
       decodeLatin1
```
harendra-kumar, line 70: Since this is a new module in this release we should add "since 0.8.0" for all routines here. We forgot this when renaming this.

harendra-kumar (changes requested):

src/Streamly/Data/Unicode/Stream.hs
pranaysashank: > Also, the behavior other routines in this module should perhaps remain the same as old.
Nah, can't do that it'd be completely unnecessary. 

Will deprecate the lax/lenient variants.

Changelog.md
```diff
@@ -14,6 +19,11 @@
   constraint.
 * Change the associativity of combinators `serial`, `wSerial`,
   `ahead`, `async`, `wAsync`, `parallel` to be the same as `<>`.
+* `encodeUtf8`, `decodeUtf8` replaces any invalid character encountered during
+  encoding/decoding with the Unicode replacement character. Use `encodeUtf8'`,
+  `decodeUtf8'` to recover the previous functionality.
+* `encodeLatin1` truncates any invalid codepoints in the input stream.
```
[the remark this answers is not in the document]
pranaysashank: It doesn't drop, it truncates the input characters to chars in the 0-255 range. I'll do "truncates" => "now truncates"

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +637,67 @@ encodeLatin1 = S.map convert
...
 
--- | Like 'encodeLatin1' but silently truncates and maps input characters beyond
+-- | Like 'encodeLatin1'' but silently truncates and maps input characters beyond
 -- 255 to (incorrect) chars in 0-255 range. No error or exception is thrown
 -- when such truncation occurs.
 --
--- /Since: 0.7.0/
+-- /Since: 0.7.0 ("Streamly.Data.Unicode.Stream")/
+--
+-- @since 0.8.0
```
harendra-kumar, line 646: The API in 0.7.0 was not the same, so we can mention that:

```
/Since: 0.8.0 (Lenient behavior)/
```

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +637,67 @@ encodeLatin1 = S.map convert
...
+encodeLatin1 = S.map (fromIntegral . ord)
+
+-- | Like 'encodeLatin1' but drops the input characters beyond 255.
+--
+-- @since 0.8.0
+{-# INLINE encodeLatin1_ #-}
+encodeLatin1_ :: (IsStream t, Monad m) => t m Char -> t m Word8
+encodeLatin1_ = S.map (fromIntegral . ord) . S.filter (<= chr (255))
+
+{-# DEPRECATED encodeLatin1Lax "Please use 'encodeLatin1' instead" #-}
```
harendra-kumar, line 658: No need to deprecate this as it is exposed only through a deprecated module.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +637,67 @@ encodeLatin1 = S.map convert
...
 {-# INLINE encodeLatin1Lax #-}
 encodeLatin1Lax :: (IsStream t, Monad m) => t m Char -> t m Word8
-encodeLatin1Lax = S.map (fromIntegral . ord)
+encodeLatin1Lax = encodeLatin1
 
 -- | Decode a UTF-8 encoded bytestream to a stream of Unicode characters.
 -- The incoming stream is truncated if an invalid codepoint is encountered.
```
harendra-kumar, line 664: "truncated" is not the right word. We should say "throws an exception".

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +637,67 @@ encodeLatin1 = S.map convert
...
--- /Since: 0.7.0/
+-- /Since: 0.7.0 ("Streamly.Data.Unicode.Stream")/
+--
+-- @since 0.8.0
+{-# INLINE decodeUtf8 #-}
+decodeUtf8 :: (Monad m, IsStream t) => t m Word8 -> t m Char
+decodeUtf8 = D.fromStreamD . decodeUtf8D . D.toStreamD
+
+-- | Decode a UTF-8 encoded bytestream to a stream of Unicode characters.
+-- Any invalid codepoint encountered dropped.
```
harendra-kumar, line 690: "dropped" => "is dropped"

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -606,38 +637,67 @@ encodeLatin1 = S.map convert
...
+decodeUtf8Arrays' = D.fromStreamD . decodeUtf8ArraysD' . D.toStreamD
 
 -- | Decode a UTF-8 encoded bytestream to a stream of Unicode characters.
 -- Any invalid codepoint encountered is replaced with the unicode replacement
 -- character.
 --
--- /Since: 0.7.0/
+-- /Since: 0.7.0 ("Streamly.Data.Unicode.Stream")/
+--
+-- @since 0.8.0
```
harendra-kumar, line 684: annotate behavior change in 0.8.0.

harendra-kumar (changes requested):

src/Streamly/Unicode/Stream.hs
```diff
@@ -89,6 +92,11 @@ module Streamly.Unicode.Stream
     -- , words
     -- , unlines
     -- , unwords
+
+    -- * Deprecations
+    , decodeUtf8Lax
+    , encodeLatin1Lax
+    , encodeUtf8Lax
```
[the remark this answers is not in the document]
pranaysashank: huh?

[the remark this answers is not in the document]
harendra-kumar: read it as "we should NOT have deprecated functions in this."

harendra-kumar:
I think we can keep the `xxx_` versions internal for now and expose them later if they find good adoption.

src/Streamly/Internal/Unicode/Stream.hs
pranaysashank: It is not obvious what happened to this function or what to use instead.

src/Streamly/Data/Unicode/Stream.hs
```diff
@@ -90,6 +87,11 @@ module Streamly.Data.Unicode.Stream
     -- , words
     -- , unlines
     -- , unwords
+
+    -- * Deprecations
+    , decodeUtf8Lax
+    , encodeLatin1Lax
+    , encodeUtf8Lax
```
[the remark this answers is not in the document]
pranaysashank: Well, it's not much work and also makes sense for me.

Changelog.md
```diff
@@ -14,6 +19,12 @@
   constraint.
 * Change the associativity of combinators `serial`, `wSerial`,
   `ahead`, `async`, `wAsync`, `parallel` to be the same as `<>`.
+* `encodeUtf8`, `decodeUtf8` now replace any invalid character encountered
+  during encoding/decoding with the Unicode replacement character. Use
+  `encodeUtf8'`, `decodeUtf8'` to recover the previous functionality.
+* `encodeLatin1` now silently truncates any character beyond 255 to incorrect
+  characters in the input stream. Use `encodeLatin1'` to recover previous
+  functionality.
```
harendra-kumar, line 27: Since these routines are in a new module, they are not technically breaking changes in themselves. We can write in the `Streamly.Data.Unicode.Stream` deprecation comment above that the routines in the new module have different behavior.

Changelog.md
```diff
@@ -33,9 +44,13 @@
...
-    * `foldWith` has been replaced by `concatFoldableWith`
-    * `foldMapWith` has been replaced by `concatMapFoldableWith`
-    * `forEachWith` has been replaced by `concatForFoldableWith`
+    * `foldWith` to `concatFoldableWith`
+    * `foldMapWith` to `concatMapFoldableWith`
+    * `forEachWith` to `concatForFoldableWith`
+* The following encoding/decoding routines have been renamed:
+    * `encodeUtf8Lax` to `encodeUtf8`
+    * `encodeLatin1Lax` to `encodeLatin1`
+    * `decodeUtf8Lenient` to `decodeUtf8`
```
harendra-kumar, line 53: They are not renamed but deprecated and renamed in the new module.

harendra-kumar (approved):

Issue #696, opened 2020-09-22
Add tests for fs events module


harendra-kumar:
We can create a temporary directory tree structure, perform some operations on it and watch/verify the events.

rnjtranjan:
Duplicate of #724

Issue #637, opened 2020-08-17
Add filesystem event handling module for Windows


harendra-kumar:
#632 and #628 added event notification for Linux and macOS. Similarly we can add a Win32 module for fs event streams.

harendra-kumar:
Fixed by #687 .

PR #697, opened 2020-09-22
Add support for recursive directory tree events in Linux

rnjtranjan:


src/Streamly/Internal/FileSystem/Event/Linux.hs
```diff
@@ -216,10 +225,170 @@ setFlag mask status cfg@Config{..} =
...
+
+-- | Traverse a dir recusively and return the list of subdirs.
+--
+-- /Internal/
+--
+traverseDirRec :: 
+       FilePath -- ^ subdir path
+    -> FilePath -- ^ root dir path
+    -> [FilePath] -- ^ roots passed on commnd line
+    -> StateT ([FilePath], RootMap) IO [FilePath] -- ^ Mapping with root dir
```
harendra-kumar, line 250: Use the code from the `ListDir.hs` example to return a stream of sudirs here. Then use `S.foldlM'` on the resulting stream to apply the function that adds each dir to the watch handle and return the accumulated result. The accumulator can contain the map or any other aggregated value that you may want to return. We do not need a `StateT`.

harendra-kumar (changes requested):
We can change the design a bit. Let's say we define the `Watch` type like this:

```
data Watch = Watch Handle (IORef (IntMap (Array Word8, Array Word8)))
```

The first Array is the original watch root directory supplied by the user. The second array is a subdirectory under that path.

Initially, we traverse the user supplied paths recursively and add all the subdirectories to the watch handle updating the map in this IORef, mapping the wd (watch decriptor) to the (original watch root, subdir). 

When we receive a dir create event we can add that dir to this map and also scan its contents to update this map.

This way we need to maintain only one map, its simpler and we do not do multiple lookups. When an event arrives we anyway perform a lookup of the wd map to map the wd to the path.

Let me know if this works or if there are any issues with this approach.

test/Streamly/Test/Internal/FileSystem/Event.hs
```diff
@@ -0,0 +1,314 @@
...
+import qualified Data.Set as Set
+#if !defined(CABAL_OS_WINDOWS)
+import qualified Streamly.Internal.Unicode.Stream as U
+import Data.Functor.Identity (runIdentity)
+#endif
+
+main :: IO ()
+main = hspec $ do
+    prop "Recursive directory events validator" validator
+    prop "Staging directories cleanUp" cleanUp
```
harendra-kumar, line 36: Divide the testing into many small test cases. One test case would be a small sequence of file system operations and the resulting events that are expected from these operations. Now let's write a single test case first:

1. Let's say the function to perform fs ops is called `fsOps`, this would be an IO function. It would perform a sequence of ops under our `watchRootDir`.

```
fsOps synch = do
    takeMVar synch
     ....
```

2. Write a parser `parseEvents` for this test case that would parse the stream of events from `watchRootDir`. The parser would expect a certain sequence of events. Run the parser on the event stream e.g. `checkEvents synch = S.parse parseEvents (S.before (putMVar synch) inputStream)`. Instead of a `Parser`, we could even use `S.take n` to take `n` events from the stream. We already know how much `n` should be based on the test case. And then we can match the list of these events with expected events.

3. Now compose `fsOps` and `checkEvents` to run in parallel using

```
timeout = threadDelay 5000000 >> return (Just Failure)
checker synch = S.yieldM (checkEvents synch) `S.parallel` S.yieldM timeout

 S.head $ checker synch `Par.parallelFst` S.yieldM (fsOps synch)
```

`checkEvents` can return a `Just Success` or `Just Failure` depending on whether the parse succeeded or failed. You can test the value returned by `S.head` to check if the test passed or failed. If we are not able to parse in a given time then we will receive a failure from the timeout thread.

You can replicate this for each test case and write as many tests as you want.

harendra-kumar (changes requested):

harendra-kumar (changes requested):

harendra-kumar:
Can you also send a PR to the https://github.com/composewell/streamly-examples repo with the old event testing program added as an example there?

streamly.cabal
```diff
@@ -726,8 +730,11 @@ test-suite version-bounds
...
   main-is: version-bounds.hs
 
-executable FileSystem.Event
+test-suite FileSystem.Event
   import: test-options
-  main-is: Streamly/Test/FileSystem/Event.hs
-  if !flag(manual-tests)
-    buildable: False
+  type: exitcode-stdio-1.0
+  main-is: Streamly/Test/Internal/FileSystem/Event.hs
```
harendra-kumar, line 736: We should move `Streamly/Test/Internal/FileSystem/Event.hs` to `Streamly/Test/FileSystem/Event.hs` and use the latter here, and remove the former. The CI tests are failing because the file `Streamly/Test/FileSystem/Event.hs` is present in the git tree but is not being used anywhere in cabal file.

harendra-kumar: Also, we should move the old file `Streamly/Test/FileSystem/Event.hs` to `streamly-examples` repo, it can act as a simple example instead of a test.

test/Streamly/Test/Internal/FileSystem/Event.hs
```diff
@@ -0,0 +1,210 @@
+import Control.Concurrent
+import Control.Monad.IO.Class
```
harendra-kumar, line 2: import the symbols explicitly.

test/Streamly/Test/Internal/FileSystem/Event.hs
```diff
@@ -0,0 +1,210 @@
...
+toUtf8 :: MonadIO m => String -> m (Array Word8)
+toUtf8 = Array.fromStream . Unicode.encodeUtf8' . S.fromList
+
+watchPaths :: NonEmpty (Array Word8) -> SerialT IO Event.Event
+watchPaths = Event.watchTrees
+
+timeout :: IO String
+timeout = threadDelay 5000000 >> return "Timeout"    
+
+data Sync = Sync (MVar ())
```
harendra-kumar, line 43: Do we need this type? We can just use the MVar type directly.

harendra-kumar (changes requested):

rnjtranjan:
"fixes #696"

rnjtranjan:
Fixes #724

harendra-kumar:
@rnjtranjan 

1. take a look at the last two commits and see if everything makes sense or some changes are needed
2. there are too many commits in the PR, we can squash the previous commits into a few logical commits

There are several TODOs that I have noted in the comments. We can fix the comments in the test file right away, it will reduce a lot of code. We can fix the some of the other ones after merging this PR.

rnjtranjan:
Fixes #637

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+type Config = Event.Config
+type Event = Event.Event
+
+setAccessed :: Toggle -> Config -> Config
+#if defined(CABAL_OS_WINDOWS)
+setAccessed = Event.setModifiedLastWrite
+#elif defined(CABAL_OS_LINUX)
+setAccessed = Event.setAccessed
+#elif defined(CABAL_OS_DARWIN)
+setAccessed = undefined
```
harendra-kumar, line 90: We cannot leave it `undefined`. This means this module won't work on darwin. If an API cannot be implemented on any platform then we cannot have it in the common module.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#endif
+
+setMetadataChanged :: Toggle -> Config -> Config
+#if defined(CABAL_OS_WINDOWS)
+setMetadataChanged = Event.setModifiedAttribute
+#elif defined(CABAL_OS_LINUX)
+setMetadataChanged = Event.setMetadataChanged
```
harendra-kumar, line 97: If these two APIs have the same functionality on windows and linux then we should use the same name for them on both platforms in the platform specific module. Same holds for any other such APIs.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+setAllEvents = defaultConfig
+#else
+setAllEvents = Event.setAllEvents
+#endif
+
+getAbsPath :: Event -> String
+#if defined(CABAL_OS_DARWIN)
+getAbsPath = utf8ToString Event.getAbsPath
+#else
+getAbsPath ev = getRootPath ev </> getRelativePath ev
```
harendra-kumar, line 164: We should implement it in the platform specific module itself. Also see an issue about this that I raised earlier.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#endif
+
+getRelativePath :: Event -> String
+#if defined(CABAL_OS_WINDOWS)
+getRelativePath = Event.getRelPath
+#elif defined(CABAL_OS_DARWIN)
+getRelativePath = getAbsPath
```
harendra-kumar, line 171: This is not relative path.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#elif defined(CABAL_OS_LINUX)
+getRelativePath = utf8ToString . Event.getRelPath
+#endif
+
+getRootPath :: Event -> String
+#if defined(CABAL_OS_WINDOWS)
+getRootPath = Event.getRoot
+#elif defined(CABAL_OS_DARWIN)
+getRootPath = getAbsPath
```
harendra-kumar, line 180: This is not root path.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#else
+isMoved ev = Event.isMovedFrom ev || Event.isMovedTo ev
+#endif
+
+isMovedFrom :: Event -> Bool
+#if defined(CABAL_OS_DARWIN)
+isMovedFrom = Event.isMoved
+#else
+isMovedFrom = Event.isMovedFrom
+#endif
```
harendra-kumar, line 203: Moved on darwin may not have the same semantics as movedFrom on Linux/Windows. Same holds for MovedTo.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+
+watch :: Bool -> Config -> NonEmpty (Array Word8) -> SerialT IO Event.Event
+#if defined(CABAL_OS_DARWIN)
+watch rec cfg paths =  Event.watchTreesWith (\_ -> cfg) paths
+#else
+watch rec cfg paths =  case rec of
+    True -> Event.watchTreesWith (\_ -> cfg) paths
+    False -> Event.watchPathsWith (\_ -> cfg) paths
+#endif
```
harendra-kumar, line 239: We should define a `watch` API in the platform specific modules and import it from there.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#else
+isMovedTo =  Event.isMovedTo
+#endif
+
+isModified :: Event -> Bool
+isModified = Event.isModified
+
+isOverflow :: Event -> Bool
+#if defined(CABAL_OS_DARWIN)
+isOverflow = False
```
harendra-kumar, line 217: There is a `isMustScanSubdirs` on darwin which is similar to overflow.

src/Streamly/FileSystem/Event.hs
```diff
@@ -0,0 +1,239 @@
...
+#endif
+
+showEvent :: Event -> String
+showEvent = Event.showEvent
+
+emptyConfig :: Config
+#if defined(CABAL_OS_DARWIN)
+emptyConfig = Event.Config 0.0 0
+#else
+emptyConfig = Event.Config True 0
```
harendra-kumar, line 229: This should be `defaultConfig`, and we should not export the defaultConfig from platform specific modules.

harendra-kumar (changes requested):

test/Streamly/Test/FileSystem/Event.hs
```diff
@@ -395,4 +396,4 @@ testDesc =
 main :: IO ()
 main = do
     hSetBuffering stdout NoBuffering
-    hspec $ sequence_ $ map driver testDesc
+    hspec $ sequence_ $ map driver testDesc
```
harendra-kumar, line 399: You may have to change your editor settings or be careful about removing newline at the end of file.

PR #698, opened 2020-09-22
Unicode normalization combinators

adithyaov:
Todo:
* [x] Add benchmarks


adithyaov:
Ignore `cabal.project`. We'll release `unicode-data` before merging this.

harendra-kumar:
@adithyaov can you rebase this, port it on the new unicode-data package, and add benchmarks from unicode-transforms package?

src/Streamly/Internal/Unicode/Normalize.hs
```diff
@@ -0,0 +1,273 @@
...
+    = YieldCharList [Char] (DecomposeState st)
+    | ReadInputChar ReBuf st
+    | IsHangul Char st
+    | IsDecomposable [Char] ReBuf st
+    | DecomposeStop
+
+{-# INLINE_NORMAL partialDecomposeD #-}
+partialDecomposeD ::
+       Monad m => Bool -> DecomposeMode -> Stream m Char -> Stream m Char
+partialDecomposeD decomposeHangul mode (Stream step state) =
```
harendra-kumar, line 63: Call this `decomposeD` and remove the existing `decomposeD`.

src/Streamly/Internal/Unicode/Normalize.hs
```diff
@@ -0,0 +1,273 @@
...
+    | ComposeReg Int [Char] st
+    | ComposeJamo JamoBuf st
+    | ComposeStop
+
+-- Assumes every character except hangul characters are fully decomposed and the
+-- combining characters are reordered. Hangul characters may or may not be
+-- decomposed.
+{-# INLINE_EARLY partialComposeD #-}
+partialComposeD :: Monad m => Stream m Char -> Stream m Char
+partialComposeD (Stream step state) = Stream step' (ComposeNone state)
```
harendra-kumar, line 147: call this `composeD`.

harendra-kumar (changes requested):
1. Let's add the normalization routines to the Unicode.Char module. We will use this module for operations on Char type including stream of Char to Char transformations. The Unicode.Stream module should go away. We will move the encode/decode routines in Unicode.Stream to Unicode.Utf8 and Unicode.Latin1.
2. Let's add some basic benchmarks as well so that we can start working on performance optimizations of these modules. Just take an input file and normalize it. We already have some benchmarks in the existing Unicode.Stream benchmarks that take input files using env vars.

Once the normalization tests pass and we have benchmarks in place, we can commit this and do the performance optimizations later.

src/Streamly/Internal/Unicode/Normalize.hs
adithyaov: This is kind of partially composing. This does not compose a semi-composed stream. It shouldn't be used directly.

src/Streamly/Internal/Unicode/Char.hs
```diff
@@ -63,3 +78,246 @@ toUpper = undefined
...
+    | Hangul !Char -- Hangul Syllable LV or LVT
+    | HangulLV !Char
+
+{-# INLINE fromJamoBuf #-}
+fromJamoBuf :: JamoBuf -> Char
+fromJamoBuf (Jamo ch) = ch
+fromJamoBuf (Hangul ch) = ch
+fromJamoBuf (HangulLV ch) = ch
+
+
```
harendra-kumar, line 182: remove blank line.

harendra-kumar (approved):
Looks good at a high level. I did not check the details.

PR #708, opened 2020-10-07
Use namespaced components when passing to cabal in bench.sh.

### LINKED ISSUE #676: Unable to run individual benchmarks
```
$ cabal bench Prelude.Async
cabal: The bench command is for running benchmarks, but the target
'Prelude.Async' refers to the test suite Prelude.Async from the package
streamly-0.7.2.
```

If I run this command inside the `benchmark` directory then it rebuilds streamly. Also, we need a cabal.project.local there to pick up the local streamly.

1) rename the tests using a Test prefix e.g. `Test.Prelude.Async`
2) add a `cabal.project.local` in the `benchmark` directory to pick the local streamly lib, just in case we want to use it

pranaysashank:
Closes #676.

### DIFF
--- a/bench.sh
+++ b/bench.sh
@@ -565,7 +565,11 @@ then
   then
     $BUILD_BENCH || die "build failed"
   else
-    $BUILD_BENCH $EXECUTABLES || die "build failed"
+    for c in $EXECUTABLES
+    do
+      COMPONENTS+="streamly-benchmarks:bench:$c "
+    done
+    $BUILD_BENCH $COMPONENTS || die "build failed"
   fi
   run_measurements "$BENCHMARKS"
 fi


bench.sh
```diff
@@ -565,7 +565,11 @@ then
   then
     $BUILD_BENCH || die "build failed"
   else
-    $BUILD_BENCH $EXECUTABLES || die "build failed"
+    for c in $EXECUTABLES
+    do
+      COMPONENTS+="streamly-benchmarks:bench:$c "
```
adithyaov, line 570: It would make me feel a lot better if you define `COMPONENTS` before usage.

adithyaov (approved):

harendra-kumar (approved):
This solves the issue when running through bench.sh. How about `cabal bench`? Can we write some instructions in `benchmark/README` about how to run `cabal bench`?

PR #709, opened 2020-10-11
Additional combinators

adithyaov:
* [ ] Add tests for all the combinators added
* [x] Add benchmarks for all the combinators added
* [x] Add documentation for all the combinators added
* [x] Choose between `retry` and `retry1`

Combinators added:
1. retry
2. choice
3. takeP


adithyaov:
See #1176 #1174 #1171

PR #710, opened 2020-10-11
Add default.nix for nix shell/build

harendra-kumar:
It depends on a fork of nixpkgs which adds a minor change to `shellFor` so that it honors the benchmark dependencies.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,103 @@
+# CAUTION! a spelling mistake in arg string is ignored silently.
+#
+# To include benchmark deps use:
+# nix-shell --argstr c2nix "--benchmark --flag fusion-plugin"
+#
+# To build the chart executable for running bench.sh use:
+# nix-shell --argstr c2nix "--flag dev" --run "cabal build chart --flag dev"
+#
+# To use ghc-8.6.5
+# nix-shell --argstr compiler "ghc865"
+
+{
+  nixpkgs ?
+    import (builtins.fetchTarball https://github.com/composewell/nixpkgs/archive/01dd2b4e738.tar.gz)
+        # fusion-plugin is marked as broken
+        { config.allowBroken = true;}
+, compiler ? "default"
+, c2nix ? "" # cabal2nix CLI options
+# TODO
+#, sources ? [] # e.g. [./. ./benchmark]
+#, hdeps ? [] # e.g. [time, mtl]
+#, deps ? [] # e.g. [SDL2]
+}:
+let haskellPackages =
+        if compiler == "default"
+        then nixpkgs.haskellPackages
+        else nixpkgs.haskell.packages.${compiler};
+
+    # we can possibly avoid adding our package to HaskellPackages like
+    # in the case of nix-shell for a single package?
+    mkPackage = super: pkg: path: opts: inShell:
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -0,0 +1,81 @@
...
+}:
+let haskellPackages =
+        if compiler == "default"
+        then nixpkgs.haskellPackages
+        else nixpkgs.haskell.packages.${compiler};
+
+    # we can possibly avoid adding our package to HaskellPackages like
+    # in the case of nix-shell for a single package?
+    mkPackage = self: pkg: path: opts: inShell:
+                let orig = self.callCabal2nixWithOptions pkg path opts {};
```
adithyaov, line 25: It's good practice to always use `super` for functions.

default.nix
```diff
@@ -0,0 +1,81 @@
...
+        haskellPackages.override {
+            overrides = self: super:
+                with nixpkgs.haskell.lib;
+                {
+                    streamly = mkPackage self "streamly" ./. c2nix inShell;
+                    streamly-benchmarks =
+                        mkPackage self "streamly-benchmarks"
+                            ./benchmark c2nix inShell;
+                    QuickCheck = self.QuickCheck_2_14;
+
```
adithyaov, line 42: You might need to add a `splitmix = haskellPackages.splitmix_0_1_0_1;`, a depedency of `QuickCheck_2_14`

`self` here would cause an infinite recursion as `haskellPackages.callPackage` depends on `splitmix`

harendra-kumar: Its a working/tested code. Note I am using      `import (builtins.fetchTarball https://github.com/composewell/nixpkgs/archive/f0ec167983f.tar.gz)`.

adithyaov: Ok

adithyaov:
We also need to also add build instructions for `bench-show` if we want to use `bench.sh`.

harendra-kumar:
> We also need to also add build instructions for bench-show if we want to use bench.sh.

Its there:

```
# To build the chart executable for running bench.sh use:
# nix-shell --argstr c2nix "--flag dev" --run "cabal build chart --flag dev"
```

harendra-kumar:
This default.nix can be further improved by turning this code into a project shell derivation function:

```
projectShell = paths: c2nixFlags: haskellDeps: otherDeps:
```

Where `paths` is the list of cabal file paths e.g. `[./. ./benchmark]`, `haskellDeps` is any additional haskell dependencies that we may want, `otherDeps` is any additional non-haskell dependencies.

These arguments can be taken from command line when creating the shell.

The next level improvement would be to parse the `paths` from cabal.project. But we can do that later.

adithyaov:
> The next level improvement would be to parse the paths from cabal.project. But we can do that later.

People have already done this.
Look at https://gist.github.com/codebje/000df013a2a4b7c10d6014d8bf7bccf3
This is one of the many implementations.

harendra-kumar:
yes, I have seen that. We can possibly use that for cabal.project parsing later on, though there seem to be some caveats about it. Currently I am only interested in something like the `projectShell` function above, its a low level way of specifying everything we need.

adithyaov (approved):

PR #712, opened 2020-10-13
Put fs event executable under manual-tests flag

harendra-kumar:
Otherwise it is always built when we build the library.

### DIFF
--- a/src/Streamly/Internal/FileSystem/Event/Darwin.hs
+++ b/src/Streamly/Internal/FileSystem/Event/Darwin.hs
@@ -1020,4 +1020,11 @@ showEvent ev@Event{..} =
         where showev f str = if f ev then "\n" ++ str else ""
 #else
 module Streamly.Internal.FileSystem.Event.Darwin () where
+#warning "Autoconf did not find the definition \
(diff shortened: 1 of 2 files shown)


adithyaov (approved):
LGTM

src/Streamly/Internal/FileSystem/Event/Darwin.hs
```diff
@@ -1020,4 +1020,11 @@ showEvent ev@Event{..} =
         where showev f str = if f ev then "\n" ++ str else ""
 #else
 module Streamly.Internal.FileSystem.Event.Darwin () where
+#warning "Autoconf did not find the definition \
+kFSEventStreamCreateFlagFileEvents in Darwin header files.\
+Do you have Cocoa framework header files installed?\
+Not compiling the Streamly.Internal.FileSystem.Event module. \
```
adithyaov, line 1026: Where is `Streamly.Internal.FileSystem.Event`?

harendra-kumar: Should be `Streamly.Internal.FileSystem.Event.Darwin`.

PR #714, opened 2020-10-13
Allow QuickCheck version to be >= 2.13

### LINKED ISSUE #711: Allow QuickCheck < 2.14
There are many packages that have not moved to QuickCheck 2.14 yet. It may not be a good idea to only allow >= 2.14 as yet. I found it out when trying to use streamly with nix.

adithyaov:
Closes #711 .

Should we try to relax lower bound even more?

### DIFF
--- a/streamly.cabal
+++ b/streamly.cabal
@@ -330,7 +330,7 @@ common prelude-test-options
   build-depends:
       streamly
     , base              >= 4.9   && < 5
-    , QuickCheck        >= 2.14  && < 2.15
+    , QuickCheck        >= 2.13  && < 2.15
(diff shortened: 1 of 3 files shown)


adithyaov:
`2.13` is marked as broken. We can probably start from `2.13.1`

harendra-kumar (approved):

harendra-kumar:
> Should we try to relax lower bound even more?

I saw all packages had `< 14`, if we see some important package has lower we can try then. Anyway this is test only so only those who include test dependencies would be affected by this.

adithyaov:
See #718

Issue #667, opened 2020-08-25
Add mapMaybe for folds


harendra-kumar:

adithyaov:
I'm assuming this is `lmapMaybe`?

harendra-kumar:
Fixed by #717 .

PR #717, opened 2020-10-17
add mapMaybe combinator in Fold

rnjtranjan:


src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -353,6 +354,16 @@ lmapM f (Fold step begin done) = Fold step' begin done
   where
     step' x a = f a >>= step x
 
+-- | @(mapMaybe f fold)@ maps a 'Maybe' returning function @f@ on the input of the fold.
+-- filter out the 'Nothing' elements, and return a values extracted from 'Just'
+-- >>> S.fold (FL.mapMaybe (\x -> case x of  5 -> Just x ; _ -> Nothing) FL.sum) 
+-- (S.enumerateFromTo 1 10) 
+-- 5
+-- @since 0.8.1
```
harendra-kumar, line 362: We usually keep the combinators internal for a while and then expose them later once we know they are stable and should be exposed. So here we can mark it `Internal` instead of annotating with a release. The markup would be:

```
/Internal/
```

Also, let's make it `INLINE` instead of `INLINABLE`. We have marked some functions `INLINABLE` but we really want these `INLINE` for guaranteed fusion.

harendra-kumar: Let's add a benchmark and test as well for this. See `benchmark/Streamly/Benchmark/Data/Fold.hs` for existing benchmarks and `test/Streamly/Test/Internal/Data/Fold.hs` for tests.

harendra-kumar: In the docs, need a comma instead of period:

```
on the input of the fold, filter out the 'Nothing' elements,
```

harendra-kumar (changes requested):

test/Streamly/Test/Internal/Data/Fold.hs
```diff
@@ -96,6 +96,9 @@ safeLast (_:xs) = safeLast xs
 last :: [String] -> Expectation
 last ls = S.fold FL.last (S.fromList ls) `shouldReturn` safeLast ls
 
+mapMaybe :: Expectation
+mapMaybe = S.fold (F.mapMaybe (\x -> case x of  5 -> Just x ; _ -> Nothing) F.sum)
+    (S.enumerateFromTo 1 10) `shouldReturn` (5 :: Int)
```
harendra-kumar, line 101: We can write this as `mapMaybe :: Int -> Expectation`, then generate a list `xs = [1..n]`. Then we can map `(\x -> if even x then Just x else Nothing)` on `fromList xs` and compare that the result is the same as `filter even xs`.

harendra-kumar (changes requested):

rnjtranjan:
Fixes #667

Issue #159, opened 2018-12-09
make sure there is at least one test for each user exposed API


harendra-kumar:
Sometimes we do not add tests due to lack of time and sometimes we do not add tests because an API gets tested through some other API. However, it is important to have tests, at the minimum we can add a trivial test using the symbol, this is to ensure that we do not accidentally remove an API during refactor and do not even get to know about it.

pranaysashank:
## src/Streamly/Prelude.hs
- [ ] mkAsync
- [ ] IsStream ()
- [X] serially
- [X] wSerially
- [X] asyncly
- [X] aheadly
- [X] wAsyncly
- [X] parallely
- [X] zipSerially
- [X] zipAsyncly
- [X] adapt
- [X] nil
- [ ] cons
- [ ] (.:)
- [ ] consM
- [ ] (|:)
- [ ] yield
- [ ] yieldM
- [ ] repeat
- [ ] repeatM
- [X] replicate
- [X] replicateM
- [ ] Enumerable (..)
- [ ] enumerate
- [ ] enumerateTo
- [ ] unfoldr
- [X] unfoldrM
- [X] unfold
- [X] iterate
- [X] iterateM
- [X] fromIndices
- [X] fromIndicesM
- [X] fromList
- [ ] fromListM
- [X] fromFoldable
- [X] fromFoldableM
- [ ] uncons
- [X] tail
- [X] init
- [X] foldrM
- [X] foldr
- [X] foldl'
- [X] foldl1'
- [X] foldlM'
- [X] fold
- [X] drain
- [X] last
- [X] length
- [X] sum
- [X] product
- [X] maximumBy
- [X] maximum
- [X] minimumBy
- [X] minimum
- [X] the
- [X] toList
- [X] drainN
- [ ] drainWhile
- [X] (!!)
- [X] head
- [ ] findM
- [X] find
- [X] lookup
- [X] findIndex
- [X] elemIndex
- [X] null
- [X] elem
- [X] notElem
- [X] all
- [X] any
- [X] and
- [X] or
- [ ] eqBy
- [ ] cmpBy
- [X] isPrefixOf
- [X] isSubsequenceOf
- [X] stripPrefix
- [ ] (|$)
- [X] (|&)
- [ ] (|$.)
- [ ] (|&.)
- [ ] map
- [ ] sequence
- [X] mapM
- [ ] mapM_
- [ ] trace
- [X] tap
- [X] scanl'
- [X] scanlM'
- [ ] postscanl'
[...]

PR #728, opened 2020-10-21
Implement tests for Data.Fold combinators

wildcat26:
To implement tests for remaining `Data.Fold` combinators #159


harendra-kumar:
We have most of the tests now, we can close this.

PR #737, opened 2020-10-23
Fix "Invalid byte" on file read

harendra-kumar:
We were reading input from a binary file trying to decode it as Char
leading to an invalid byte sequence error.


adithyaov (approved):

adithyaov:
Few CI's are failing.

PR #740, opened 2020-10-26
Miscellaneous fixes.

pranaysashank:


adithyaov:
Maybe you can remove `hs-source-dirs` everywhere.

harendra-kumar (approved):
1) Can we add a point in the MAINTAINERS guide to remember to do a coverage build before the release, so that we ensure that the DEVBUILD flag is actually being passed and used.

2) We should remove the coverage badge from the README because it does not give the true picture of coverage.

PR #745, opened 2020-10-27
WIP: Add a primitive CI benchmark workflow

adithyaov:


CI.hs
```diff
@@ -0,0 +1,88 @@
+-- XXX IDEALLY:
+-- Right now this is very minimal and has no external dependencies.
+-- Use streamly
+-- Move this into benchmark/
+-- Add an executable to streamly-bencharks or merge this with Chart.hs
```
harendra-kumar, line 5: Chart.hs should go away once bench-show is made generic enough to handle the cases in it.

adithyaov: This should not be reviewed.

adithyaov:
We, unfortunately, cannot share artifacts across workflows in Github CI

adithyaov:
We need to somehow persist data between workflows,
One way to do this is to create a repository for saving this data. 
For every push to master, we can push the `charts/` to it.
For every pull request, we can pull charts from that repository.
This way we need to only build charts on master for every new push to master.

adithyaov:
Check allocations will currently fail as `dd` is not configured properly for linux. This depends on #747

adithyaov:
We can possibly save the temporary benchmark files as artifacts and share them across jobs.

adithyaov:
The reason that there are 2 jobs in `check-allocations.yml` is because, in the future I'd like to move one job out to a different workflow that will run on every push to master and save the results in S3 or some other persistent backend

.github/workflows/check-allocations.yml
```diff
@@ -0,0 +1,94 @@
...
+
+  run-benchs:
+    name: Run benchmarks
+    needs: gen-artifacts
+    env:
+      BENCH_SH: ./bench.sh
+      CI_HS: benchmark/CI.hs
+      CI_EXE: ./CI
+      CHARTS_DIR: charts/
+      DEGRADATION: -20
```
adithyaov, line 54: `DEGRADATION` is `(1 - x2/x1) * 100`
`x1 = master`
`x2 = pr`

benchmark/CI.hs
```diff
@@ -0,0 +1,82 @@
...
+        fp
+
+    where
+
+    muxRes ls
+        | length ls == 0 = Nothing
+        | otherwise = Just $ unlines $ map (\(a, b) -> a ++ " " ++ show b) ls
+
+main :: IO ()
+main = do
```
adithyaov, line 77: Change to 4 space indentation.

adithyaov:
As discussed on discord,
* We can perhaps learn from GHC as well. Look at how they have automated benchmark regression tests.
* One interesting thing they have is "instruction count", that should be a stable measure like allocations, we can reliably test that in automation. CPU time/cycles are very unreliable in comparison. We should introduce an instruction count measure.
* Instead of using gauge we can have a simple module in streamly itself to measure such metrics. We anyway must have that as a diagnostic tool in streamly.

PR #746, opened 2020-10-27
Specify benchmark input file via env var

harendra-kumar:
Specifying the input file via an environment variable was broken.

### DIFF
--- a/benchmark/Streamly/Benchmark/FileSystem/Handle.hs
+++ b/benchmark/Streamly/Benchmark/FileSystem/Handle.hs
@@ -65,7 +65,7 @@ blockCount size = (size + blockSize - 1) `div` blockSize
 main :: IO ()
 main = do
     (_, cfg, benches) <- parseCLIOpts defaultStreamSize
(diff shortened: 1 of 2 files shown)


benchmark/Streamly/Benchmark/FileSystem/Handle.hs
```diff
@@ -65,7 +65,7 @@ blockCount size = (size + blockSize - 1) `div` blockSize
 main :: IO ()
 main = do
     (_, cfg, benches) <- parseCLIOpts defaultStreamSize
-    r <- lookupEnv "Benchmark.FileSystem.Handle.InputFile"
+    r <- lookupEnv "Benchmark_FileSystem_Handle_InputFile"
```
adithyaov, line 68: Where do we set this variable?

adithyaov (approved):

benchmark/Streamly/Benchmark/FileSystem/Handle.hs
harendra-kumar: In the shell when running the benchmark.

PR #747, opened 2020-10-28
Avoid using dd, pipe data directly & truncate it using head

adithyaov:
dd is confusing to work with and has it's own idiosyncracies.

### DIFF
--- a/benchmark/Streamly/Benchmark/FileSystem/Handle.hs
+++ b/benchmark/Streamly/Benchmark/FileSystem/Handle.hs
@@ -52,12 +52,6 @@ smallFileSize = 10 * 1024 * 1024
 bigFileSize :: Int
 bigFileSize = 100 * 1024 * 1024
 
-blockSize :: Int
-blockSize = 32768
-
-blockCount :: Int -> Int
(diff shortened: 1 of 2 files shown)


harendra-kumar (commented):
Looks good, but need to test it on mac as we have changed the device to urandom. We can perhaps add a simple invocation (not too expensive to run, e.g. running a single benchmark with --quick option) of bench.sh in CI so that it can be tested automatically.

Also, we may want to change the location of the input data files to "dist-newstyle" so that we do not create big files at random places within the workspace.

adithyaov:
@harendra-kumar Can we merge this? This will be implicitly tested on #745

harendra-kumar (approved):
The FileSystem.Handle benchmark run seems to be creating a directory named "benchmark-tmp/ " as well. I am not sure why, can you take a look at that too.

adithyaov:
> The FileSystem.Handle benchmark run seems to be creating a directory named "benchmark-tmp/ " as well. I am not sure why, can you take a look at that too.

We are explicitly creating it. aka `scratchDir`

harendra-kumar (approved):

adithyaov:
Merging this.

PR #749, opened 2020-10-28
Get rid of unecessary argument & extend --help with --raw

adithyaov:
build_report_progs does not need an argument


adithyaov:
Preparing the necessary things for the Benchmark CI. I wanted to add a `silent` option but turns out it already existed.

harendra-kumar (approved):

PR #751, opened 2020-10-29
Refactor splitOnSeq

harendra-kumar:
This is based on the refactor by @adithyaov done for terminating folds change. The performance after that refactor regressed by manifold. This PR fixes the performance.

* The code fuses completely now, allocations are minimal.
* Performance is same as before

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -1799,10 +1799,10 @@ data SplitOnState fs s a b w rb rh ck =
     | GO_SHORT_PAT_NEXT_WITH !fs s a !w
     | GO_SHORT_PAT_DRAIN Int !fs !w
     -- | GO_SHORT_PAT_YIELD_SEP Int !fs s w -- Only for splitSuffixOn.
-    | GO_KARP_RABIN_ACCUM Int s rb rh
-    | GO_KARP_RABIN_NEXT !fs s rb rh !ck
-    | GO_KARP_RABIN_NEXT_WITH !fs s a rb rh !ck
-    | GO_KARP_RABIN_DRAIN Int !fs rb rh
+    | GO_KARP_RABIN_ACCUM Int s rb !rh
+    | GO_KARP_RABIN_NEXT !fs s rb !rh !ck
+    | GO_KARP_RABIN_NEXT_WITH !fs s a rb !rh !ck
+    | GO_KARP_RABIN_DRAIN Int !fs rb !rh
     -- | GO_KARP_RABIN_YIELD_SEP Int !fs s rb rh -- Only for splitSuffixOn.
     | GO_YIELD b (SplitOnState fs s a b w rb rh ck)
     | GO_DONE
--- a/src/Streamly/Internal/Data/Stream/StreamD.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD.hs
@@ -1791,19 +1791,15 @@ data SplitOnState fs s a b w rb rh ck =
       GO_START
     | GO_EMPTY_PAT s
     | GO_EMPTY_PAT_WITH s a
-    -- | GO_SINGLE_PAT_BEGIN s a -- Only for splitSuffixOn
     | GO_SINGLE_PAT_NEXT !fs s a
     | GO_SINGLE_PAT_WITH !fs s a a
     | GO_SHORT_PAT_ACCUM Int s !w
     | GO_SHORT_PAT_NEXT !fs s !w
     | GO_SHORT_PAT_NEXT_WITH !fs s a !w
     | GO_SHORT_PAT_DRAIN Int !fs !w
-    -- | GO_SHORT_PAT_YIELD_SEP Int !fs s w -- Only for splitSuffixOn.
     | GO_KARP_RABIN_ACCUM Int s rb !rh
     | GO_KARP_RABIN_NEXT !fs s rb !rh !ck
-    | GO_KARP_RABIN_NEXT_WITH !fs s a rb !rh !ck
     | GO_KARP_RABIN_DRAIN Int !fs rb !rh
-    -- | GO_KARP_RABIN_YIELD_SEP Int !fs s rb rh -- Only for splitSuffixOn.
     | GO_YIELD b (SplitOnState fs s a b w rb rh ck)
     | GO_DONE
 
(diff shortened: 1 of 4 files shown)


src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1786,193 +1786,294 @@ data SplitOptions = SplitOptions
...
+                return $ Skip $ SplitOnSeqYield r SplitOnSeqDone
 
-        patWord :: Word
-        patWord = mask .&. A.foldl' addToWord 0 patArr
+    ---------------------------
+    -- Short Pattern - Shift Or
+    ---------------------------
 
-        go0 !_ !idx wrd st !acc = do
+    stepOuter _ (SplitOnSeqWordDone 0 fs _) = do
```
adithyaov, line 1914: I've tried this here and don't see much of a difference,
```
    stepOuter _ (SplitOnSeqWordDone n0 fs0 wrd0) =
        go SPEC n0 fs0 wrd0

        where

        {-# INLINE go #-}
        go !_ !n !fs !_ | n == 0 = do
            r <- done fs
            skip $ SplitOnSeqYield r SplitOnSeqDone
        go !_ !n !fs !wrd = do
            let old = elemMask .&. (wrd `shiftR` (elemBits * (n - 1)))
            fs1 <- fstep fs (toEnum $ fromIntegral old)
            go SPEC (n - 1) fs1 wrd
```

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1786,193 +1786,294 @@ data SplitOptions = SplitOptions
...
+                        skip $ SplitOnSeqYield r next
+                    else go SPEC wrd1 s fs1
+                Skip s -> go SPEC wrd s fs
+                Stop -> skip $ SplitOnSeqWordDone patLen fs wrd
+
+    -------------------------------
+    -- General Pattern - Karp Rabin
+    -------------------------------
+
+    stepOuter gst (SplitOnSeqKRInit idx st rb rh) = do
```
adithyaov, line 1979: No loopification here?

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1786,193 +1786,294 @@ data SplitOptions = SplitOptions
...
-        addCksum cksum a = cksum * k + fromIntegral (fromEnum a)
-        deltaCksum cksum old new =
-            addCksum cksum new - coeff * fromIntegral (fromEnum old)
-
-        -- XXX shall we use a random starting hash or 1 instead of 0?
-        patHash = A.foldl' addCksum 0 patArr
-
-        -- rh == ringHead
-        go0 !_ !idx !rh st !acc = do
+        go !_ !fs !st !rh !cksum = do
```
adithyaov, line 2019: `{- INLINE go -}`?

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1786,193 +1786,294 @@ data SplitOptions = SplitOptions
...
-        go0 !_ !idx wrd st !acc = do
+    stepOuter _ (SplitOnSeqWordDone 0 fs _) = do
+        r <- done fs
+        skip $ SplitOnSeqYield r SplitOnSeqDone
+    stepOuter _ (SplitOnSeqWordDone n fs wrd) = do
+        let old = elemMask .&. (wrd `shiftR` (elemBits * (n - 1)))
+        fs1 <- fstep fs (toEnum $ fromIntegral old)
+        skip $ SplitOnSeqWordDone (n - 1) fs1 wrd
+
+    stepOuter gst (SplitOnSeqWordInit idx0 wrd0 st0) =
```
adithyaov, line 1922: `idx0` is always 0
`wrd0` is always 0

src/Streamly/Internal/Data/Stream/StreamD.hs
harendra-kumar: Does not make any difference, I did it and measured.

src/Streamly/Internal/Data/Stream/StreamD.hs
harendra-kumar: Its getting inlined/specialized anyway.

src/Streamly/Internal/Data/Stream/StreamD.hs
harendra-kumar: Yeah we can remove these.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: No difference in performance, dint check fusion though.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1786,193 +1786,292 @@ data SplitOptions = SplitOptions
...
+                        -- XXX We need a direct yield here otherwise GHC is not
+                        -- able to fuse the code. This GHC issue needs to be
+                        -- investigated. We should know why.
+                        return $ Yield r next
+                    else do
+                        fs <- initial
+                        skip $ SplitOnSeqKRLoop fs s rb rh1 ringHash
+                else skip $ SplitOnSeqKRInit (idx + 1) s rb rh1
+            Skip s -> skip $ SplitOnSeqKRInit idx s rb rh
+            Stop -> do
```
adithyaov, line 1999: What is `idx == maxIndex` here and the pattern is matched?

adithyaov (changes requested):
The is a bug in the case of `SplitOnSeqKRInit:Stop`

src/Streamly/Internal/Data/Stream/StreamD.hs
harendra-kumar: `idx` represents the next index that we are expecting. When `idx` == `maxIndex` and we got a `Stop` we have not got the last element, so it is not a match. It would be a match only in the `Yield` case.

adithyaov:
The last commit namely the bug fix degrades the performance. Having any kind of ring folds in the mini loops degrades the performance. Comparing the arrays in any of the mini loops degrades the performance.

adithyaov:
I've continued dev on `HEAD~1` `Optimize (loopify) ...`. The fix is on #760

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: `What if`* was my question. Typo.

adithyaov: Would be pretty stupid of me if I refactored `splitOnSeq` without knowing what `idx` was.

harendra-kumar:
Looks good, are you fixing the `Word8` case as well?

adithyaov:
> Looks good, are you fixing the Word8 case as well?

Yeah, Not in this though.

PR #756, opened 2020-11-03
Add basic Karp-Rabin tests for splitOnSeq and splitOnSuffixSeq

adithyaov:


test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,12 +106,34 @@ splitOnSuffixSeq = do
...
+
 groupSplitOps :: String -> Spec
 groupSplitOps desc = do
     -- splitting
     splitOnSeq
     splitOnSuffixSeq
-    -- XXX add tests with multichar separators too
+
+    -- XXX add more elaborate tests with multichar separators
+    karpRabinTests
```
harendra-kumar, line 136: Aren't the property tests below more general than this, and already covering it? You can write the same for splitOnSuffixSeq. They actually test that splitOnSeq and intercalate are inverses. 

Currently the separator is a single char you can add a test with a multichar separator too. The `del` in above test can be used in the `intercalate [x] . splitOn (== x) == id` property test below.

harendra-kumar: `splitOnSuffixSeq` should probably be an inverse of `intercalateSuffix`.

adithyaov:
I'm closing this then. This is a very specific/primitive test. I'll probably write more elaborate property tests later. I'm only using this for testing some stuff while developing, this primitive test is all I need for now..

PR #757, opened 2020-11-03
More elaborate property tests for SplitOnSeq & splitOnSuffixSeq combinators

adithyaov:


test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,151 @@ splitOnSuffixSeq = do
...
+       forall a. (Arbitrary a, Eq a, Show a, Storable a, Enum a)
+    => a
+    -> String
+    -> Spec
+seqSplitterProperties sep desc = do
+    describe (desc <> " splitOnSeq/intercalate")
+        $ do
+            -- Empty seperator
+            intercalateSplitEqId 0
+            concatSplitIntercalateEqConcat splitOnSeq_ intercalate 0
```
harendra-kumar, line 120: In the same tests we could add `Streamly.intercalate` and `Streamly.intercalateSuffix` as well. Do not remove the list intercalate, keep that too, streamly intercalate should also provide the same results as list intercalate.

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,151 @@ splitOnSuffixSeq = do
...
+        $ do
+            -- Empty seperator
+            intercalateSplitEqIdNoSepEnd 0
+            concatSplitIntercalateEqConcat splitOnSuffixSeq_ intercalateSuffix 0
+
+            -- Single element seperator
+            intercalateSplitEqIdNoSepEnd 1
+            concatSplitIntercalateEqConcat splitOnSuffixSeq_ intercalateSuffix 1
+
+            -- Shift Or
```
adithyaov, line 147: These comments only make sense if we use `Word8`.

test/Streamly/Test/Prelude/Serial.hs
adithyaov: Why would we want to test `Streamly.intercalate` here?

harendra-kumar: Ideally we should be using Streamly.intercalate instead of List.intercalate here because we want to test that Streamly.splitOnSeq and Streamly.intercalate are inverses. But its ok to use List.intercalate to test against an existing implementation of intercalate.

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,258 @@ splitOnSuffixSeq = do
...
+                "S.intercalate . splitOnSeq == id ("
+                    <> show i <> " element separator)"
+         in prop name
+                $ forAll listWithSep
+                $ \xs -> do
+                      withMaxSuccess maxTestCount
+                          $ monadicIO
+                          $ do
+                              ys <- splitOnSeq_ xs (replicate i sep)
+                              zs <-
```
harendra-kumar, line 240: Why don't we move this step to `intercalateSplitEqId` itself and remove this function which is mostly the as the former. We can just use `listEquals xs ys` and `listEquals xs zs` in the same function.

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,258 @@ splitOnSuffixSeq = do
...
+    listsWithoutSep1 :: Gen [[a]]
+    listsWithoutSep1 = listOf1 listWithoutSep
+
+    intercalateSplitEqId i =
+        let name =
+                "intercalate . splitOnSeq == id ("
+                    <> show i <> " element separator)"
+         in prop name
+                $ forAll listWithSep
+                $ \xs -> do
```
harendra-kumar, line 219: We can move this do block to a where clause to reduce the indentation. Or we can assign the `intercalate` in `listEquals` to a variable so that `listEquals` check becomes a single line.

harendra-kumar (changes requested):

test/Streamly/Test/Prelude/Serial.hs
adithyaov: Yea, we can do that in all functions.

adithyaov: I could probably do it for all of them but then it creates long lines. I've done it for the suggested

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,207 @@ splitOnSuffixSeq = do
...
+        let name =
+                "splitter . intercalater == id"
+                    <> " (exclusive separator/non-empty list)"
+         in prop name
+                $ forAll listsWithoutSep1
+                $ \xss -> do
+                      withMaxSuccess maxTestCount
+                          $ monadicIO
+                          $ do
+                              xs <-
```
harendra-kumar, line 293: Why can't we combine `splitIntercalateEqIdS` with `splitIntercalateEqId`?

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,207 @@ splitOnSuffixSeq = do
...
+                              listEquals (==) (concat ys) (concat xss)
+
+    concatSplitIntercaEqConcatS splitter intercalater i =
+        let name =
+                "concat . splitter . S.intercalater == "
+                    <> "concat ("
+                    <> show i <> " element separator/possibly empty list)"
+         in prop name
+                $ forAll listsWithoutSep
+                $ \xss -> withMaxSuccess maxTestCount $ monadicIO $ testCase xss
```
harendra-kumar, line 257: Why can't we combine `concatSplitIntercaEqConcatS` with `concatSplitIntercaEqConcat`?

harendra-kumar (changes requested):

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,192 @@ splitOnSuffixSeq = do
...
+       forall a. (Arbitrary a, Eq a, Show a, Storable a, Enum a)
+    => a
+    -> String
+    -> Spec
+seqSplitterProperties sep desc = do
+    describe (desc <> " splitOnSeq")
+        $ do
+            -- Empty seperator
+            intercalateSplitEqId 0
+            concatSplitIntercaEqConcat splitOnSeq_ intercalate IS.intercalate 0
```
harendra-kumar: In all these functions - why are we passing `intercalate` and `IS.intercalate`? We can use them directly without passing.

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,160 @@ splitOnSuffixSeq = do
...
+                      intercalateSuffix
+                      IS.intercalateSuffix
+
+            -- Exclusive case
+            splitIntercalateEqId
+                splitOnSuffixSeq_
+                intercalateSuffix
+                IS.intercalateSuffix
+
+    where
```
harendra-kumar, line 145: This where clause has become too big. We could possibly float these definitions out to the top level. We can also split the tests for `splitOnSeq` and `splitOnSuffixSeq` in separate functions.

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -106,55 +107,160 @@ splitOnSuffixSeq = do
...
+seqSplitterProperties sep desc = do
+    describe (desc <> " splitOnSeq")
+        $ do
+
+            forM_ [0, 1, 2, 4] $ intercalateSplitEqId
+            forM_ [0, 1, 2, 4]
+                $ concatSplitIntercalateEqConcat
+                      splitOnSeq_
+                      intercalate
+                      IS.intercalate
```
harendra-kumar, line 124: We do not need to pass the last two arguments, we can use them directly. In all these functions.

harendra-kumar (changes requested):

test/Streamly/Test/Prelude/Serial.hs
adithyaov: We need to pass `intercalateSuffix` in the case of `splitOnSuffixSeq_`

test/Streamly/Test/Prelude/Serial.hs
adithyaov: These functions are pretty specific to `splitOnSeq`. I don't see any need for moving them out. I did initially split the tests but that led to some code duplication earlier

test/Streamly/Test/Prelude/Serial.hs
adithyaov: We need to pass intercalateSuffix in case of `splitOnSuffixSeq_`

harendra-kumar (approved):

PR #758, opened 2020-11-06
Fix StreamD.splitOnSeq

adithyaov:
I'm still very skeptical if we tested all cases. The first time I ran this after the fix, the tests failed but every other consecutive time the test succeeded.


adithyaov:
See #760

PR #759, opened 2020-11-06
 Rename all test-suites to be similar to corresponding modules. 

### LINKED ISSUE #146: Module wise test organization
How about having a test file for each user facing module? It will be easier to trace tests for the corresponding source. Plus, the test file can be referenced in the documentation for example usage, if it is organized per module it will be easier for users to find test examples for the APIs.

pranaysashank:
Closes #146


streamly.cabal
```diff
@@ -670,10 +670,10 @@ test-suite array-test
   main-is: Streamly/Test/Array.hs
   cpp-options: -DTEST_ARRAY
 
-test-suite internal-data-fold-test
+test-suite data-fold-test
```
harendra-kumar, line 673: The name should be `Data.Fold`.

harendra-kumar (changes requested):
Can we also change the names of all other tests as well to conform to the naming based on module names?

streamly.cabal
```diff
@@ -664,58 +664,58 @@ test-suite Prelude.MaxRate
   else
     buildable: False
 
-test-suite array-test
+test-suite Data.Array.Storable.Foreign
   import: test-options
   type: exitcode-stdio-1.0
   main-is: Streamly/Test/Array.hs
```
harendra-kumar, line 670: This test module name also needs to in sync with the source module name.

harendra-kumar: We can have a common module and include that in each array module, defining the module specific cpp macro before we include it.

streamly.cabal
```diff
@@ -664,58 +664,58 @@ test-suite Prelude.MaxRate
...
+test-suite Data.Parser.ParserD
   import: test-options
   type: exitcode-stdio-1.0
   main-is: Streamly/Test/Data/Parser/ParserD.hs
 
-test-suite data-array-test
+test-suite Data.Array
   import: test-options
   type: exitcode-stdio-1.0
   main-is: Streamly/Test/Array.hs
```
harendra-kumar, line 698: Make the test and source module name consistent.

harendra-kumar (changes requested):

.circleci/config.yml
```diff
@@ -232,13 +232,37 @@ jobs:
...
+      <<: *env
+      steps:
+        - checkout
+        - *preinstall
+        - *restore
+        - run:
+            name: Build and test with inspection, coverage, Werror
+            environment:
+              BUILD: cabal-v2
+              COVERALLS_OPTIONS: "--repo-token=9ezA743hCAcmzX3Ic273LhNlIXVdl7D1l --coverage-mode=StrictlyFullLines --exclude-dir=test"
```
harendra-kumar, line 245: Why do we have multiple COVERALLS builds, 8.2.2 build already has this?

test/Streamly/Test/Data/Array.hs
```diff
@@ -0,0 +1,12 @@
+-- |
+-- Module      : Streamly.Test.Data.Array
+-- Copyright   : (c) 2019 Composewell Technologies
+--
```
harendra-kumar, line 4: We could remove this blank line from all headers.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #760, opened 2020-11-07
Fix splitOn

adithyaov:


src/Streamly/Memory/Ring.hs
```diff
@@ -58,6 +59,10 @@ data Ring a = Ring
     , ringBound :: {-# UNPACK #-} !(Ptr a)        -- first address beyond allocated memory
     }
 
+-- | Get the first address of the ring as a pointer.
+startOf :: Ring a -> Ptr a
```
adithyaov, line 63: We should have a more elaborate name here.

adithyaov: `startOf` might be confusing.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1801,6 +1801,7 @@ data SplitOnSeqState rb rh ck w fs s b x =
 
     | SplitOnSeqKRInit Int s rb !rh
     | SplitOnSeqKRLoop fs s rb !rh !ck
+    | SplitOnSeqKRCheck fs s rb !rh
```
adithyaov, line 1804: Moving this into the loops breaks fusion. Decreases performance.

harendra-kumar (approved):
Looks good, you can squash and merge.

I added two commits one to add an `unsafeFoldRingNM` and another one to use it in splitOnSeq. It avoids manually folding the ring and use an abstraction for that. You can drop the last commit if you want, it requires a separate yield point, which I am not sure if it is good or bad.

adithyaov:
The reason I did not do something like `unsafeFoldRingNM` is because when I introduce terminating folds i'll have to revert it anyway.

PR #770, opened 2020-11-16
Enhance the fold type to capture terminating folds

adithyaov:
A continuation of #488

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/Fold.hs
+++ b/benchmark/Streamly/Benchmark/Data/Fold.hs
@@ -77,9 +77,9 @@ all value = IP.fold (FL.all (<= value))
 take :: Monad m => Int -> SerialT m a -> m ()
 take value = IP.fold (FL.ltake value FL.drain)
 
-{-# INLINE takeWhile #-}
-takeWhile :: Monad m => Int -> SerialT m Int -> m ()
-takeWhile value = IP.fold (FL.ltakeWhile (<= value) FL.drain)
+{-# INLINE takeSepBy #-}
+takeSepBy :: Monad m => Int -> SerialT m Int -> m ()
+takeSepBy value = IP.fold (FL.takeSepBy (<= value) FL.drain)
 
 {-# INLINE many #-}
 many :: Monad m => SerialT m Int -> m Int
@@ -180,12 +180,12 @@ o_1_space_serial_elimination value =
             , benchIOSink value "drainN" (S.fold (FL.drainN value))
             , benchIOSink
                   value
-                  "drainWhileTrue"
-                  (S.fold (FL.drainWhile $ (<=) (value + 1)))
+                  "drainSepByTrue"
+                  (S.fold (FL.drainSepBy $ (<=) (value + 1)))
             , benchIOSink
                   value
-                  "drainWhileFalse"
-                  (S.fold (FL.drainWhile $ (>=) (value + 1)))
+                  "drainSepByFalse"
+                  (S.fold (FL.drainSepBy $ (>=) (value + 1)))
             , benchIOSink value "sink" (S.fold $ Sink.toFold Sink.drain)
             , benchIOSink value "last" (S.fold FL.last)
             , benchIOSink value "lastN.1" (S.fold (IA.lastN 1))
@@ -252,7 +252,7 @@ o_1_space_serial_elimination value =
             , benchIOSink value "all" $ all value
             , benchIOSink value "any" $ any value
             , benchIOSink value "take" $ take value
-            , benchIOSink value "takeWhile" $ takeWhile value
+            , benchIOSink value "takeSepBy" $ takeSepBy value
             , benchIOSink
                   value
                   "and"
--- a/benchmark/Streamly/Benchmark/Data/Fold.hs
+++ b/benchmark/Streamly/Benchmark/Data/Fold.hs
@@ -12,15 +12,28 @@ module Main (main) where
 import Control.DeepSeq (NFData(..))
 import Data.Monoid (Last(..), Sum(..))
 import System.Random (randomRIO)
+import Prelude (IO, Int, Double, String, (>), (<$>), (+), ($), even,
+                (<=), Monad(..), (==), Maybe(..), (.), fromIntegral,
+                compare, (>=), concat, seq, mod, fst, snd, const, Bool, Ord(..),
+                div, Num(..), odd, Either(..))
+
+import Data.Map.Strict (Map)
+
+import Streamly.Prelude (SerialT)
+import Streamly.Internal.Data.Fold (Fold(..))
+
+import qualified Data.Map.Strict as Map
+
+import qualified Streamly.Prelude  as S
+import qualified Streamly.Internal.Data.Fold as FL
+import qualified Streamly.Internal.Data.Fold.Types as FL
+import qualified Streamly.Internal.Data.Pipe as Pipe
+
+import qualified Streamly.Internal.Data.Sink as Sink
 
 import qualified Streamly.Data.Array.Storable.Foreign as A
-import qualified Streamly.Data.Fold as FL
 import qualified Streamly.Internal.Data.Array.Storable.Foreign as IA
-import qualified Streamly.Internal.Data.Fold as IFL
-import qualified Streamly.Internal.Data.Pipe as Pipe
-import qualified Streamly.Internal.Data.Sink as Sink
 import qualified Streamly.Internal.Data.Stream.IsStream as IP
-import qualified Streamly.Prelude  as S
 
 import Gauge
 import Streamly.Benchmark.Common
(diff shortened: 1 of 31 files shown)


src/Streamly/Internal/FileSystem/File.hs
```diff
@@ -360,7 +360,7 @@ writeChunks path = Fold step initial extract
         return (fld, h)
     step (fld, h) x = do
         r <- FL.runStep fld x `MC.onException` liftIO (hClose h)
-        return (r, h)
+        return $ FL.Partial (r, h)
```
adithyaov, line 363: Use a strict tuple here?

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1226,23 +1448,93 @@ distribute_ fs = Fold step initial extract
...
+        case r of
+            Left b -> do
+                res <- stepL sL b
+                return
+                  $ Partial
+                  $ case res of
+                        Partial sres -> RunBoth sres sR
+                        Partial1 sres -> RunBoth sres sR
+                        Done bres -> RunRight bres sR
+                        Done1 bres -> RunRight bres sR
```
harendra-kumar, line 1471: It is complicated to correctly determine leftover, we will have to return leftover only when none of the partitions consumed the element. We will have to use different states depending on whether there was a leftover or not.

Alternatively, we can say that we never return a leftover in this combinator.

harendra-kumar: `distribute`, `partition` type of combinators are perhaps best implemented without leftovers. Maybe we should error out if a fold with leftover is used in these, just like we did with transform? Especially, I am not sure if Partial1 can be handled correctly here.

harendra-kumar (commented):
I am submitting partial review, will continue tomorrow.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -949,45 +1074,138 @@ splitAt n (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
...
+    initial = finitial
+
+    step s a =
+        if not (predicate a)
+        then fstep s a
+        else do
+            res <- fstep s a
+            case res of
+                Partial sres -> Done <$> fextract sres
+                Partial1 sres -> Done <$> fextract sres
```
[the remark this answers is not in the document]
adithyaov: My line of thinking:
It's `sliceSepWith`. The predicate is satisfied. We should not be propagating it to the next fold. Either consume or drop it.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1163,27 +1351,61 @@ foldCons (Fold stepL beginL doneL) (Fold stepR beginR doneR) =
...
+
+    -- XXX This itself can be a recursive function but we might face inlining
+    -- problems.
+    step [] _ = return $ Done1 ()
+    step ss a = do
+        !ss1 <- go ss a
+        return
+          $ if Prelude.null ss1
+            then Done ()
+            else Partial ss1
```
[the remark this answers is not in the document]
adithyaov: Isn't 
```
-- 'Done1' and 'Done' are treated equally in the case of distribute_. 'Partial1'
-- and 'Partial' are also treated equally in the case of distribute_. This is
-- not the case for 'distribute'.
```
sufficient? Could be worded properly I guess.

src/Streamly/Internal/Data/Fold.hs
adithyaov: Yeah, implementation details bothered me a lot as well. Initially, `Partial1`  ran `step` again. Basically an infinite loop. But then I changed it to this after some thought.

adithyaov:
See #777

benchmark/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -177,40 +290,39 @@ o_1_space_serial_transformation value =
...
+                , benchIOSink value "classifyWith sum" $ classifyWith (fst . fn)
+                , benchIOSink value "many length sum" many
+                , benchIOSink value "split (all, any)" $ splitAllAny value
+                ]
+          ]
     ]
 
+    where
+
+    fn x = (x `mod` 3, x)
```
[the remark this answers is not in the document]
adithyaov: I'm using fn for all kinds of demux. I want it give an index out of bounds at times.
The benchmarks can be improved. I made an issue about it.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1199,7 +1141,50 @@ spanBy
     -> Fold m a b
     -> Fold m a c
     -> Fold m a (b, c)
-spanBy cmp fld1 fld2 = splitWith (,) (groupBy cmp fld1) fld2
+spanBy cmp (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
```
harendra-kumar: We can make this a parser operation and use the `groupBy` parser combinator to implement it.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1227,7 +1212,38 @@ span
     -> Fold m a b
     -> Fold m a c
     -> Fold m a (b, c)
-span p fld1 fld2 = splitWith (,) (ltakeWhile p fld1) fld2
+span p (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
```
harendra-kumar: Make this a parser operation and use splitWith, takeWhile parser ops to implement.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1270,7 +1286,46 @@ spanByRolling
     -> Fold m a b
     -> Fold m a c
     -> Fold m a (b, c)
-spanByRolling cmp fld1 fld2 = splitWith (,) (groupByRolling cmp fld1) fld2
+spanByRolling cmp (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
```
harendra-kumar, line 1276: Make this a parser operation.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1722,16 +1744,7 @@ demuxWithDefault_ f kv (Fold dstep dinitial dextract) =
...
-                                if Map.size mp == 1
-                                -- XXX This should just be Done?
-                                then Done1 ()
-                                else Partial
-                                       $ DemuxingWithDefault
-                                             dacc
-                                             (Set.insert k bst)
-                                             (Map.delete k mp)
-        else return $ Done1 ()
+        else return $ Done ()
```
harendra-kumar, line 1732: We are losing the input here, we should not run into this case.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1626,24 +1665,13 @@ demuxWith f kv = Fold step initial extract
...
                                     else Partial $ Tuple' bmp1 mp1
-                            Done1 bres ->
-                                let mp1 = Map.delete k mp
-                                    bmp1 = Map.insert k bres bmp
-                                 in if Map.size mp1 == 0
-                                    -- XXX This should just be Done?
-                                    then Done1 bmp1
-                                    else Partial $ Tuple' bmp1 mp1
-        else return $ Done1 bmp
+        else return $ Done bmp
```
harendra-kumar, line 1658: We are losing the input here, we should not run into this case.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1748,22 +1761,11 @@ demuxWithDefault_ f kv (Fold dstep dinitial dextract) =
...
                                 if Map.size mp == 1
                                 then Done ()
                                 else Partial $ DemuxingOnlyMap $ Map.delete k mp
-                            Done1 _ ->
-                                if Map.size mp == 1
-                                -- XXX This should just be Done?
-                                then Done1 ()
-                                else Partial $ DemuxingOnlyMap $ Map.delete k mp
-        else return $ Done1 ()
+        else return $ Done ()
```
harendra-kumar, line 1757: Losing the input.

src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -606,15 +568,8 @@ ltake n (Fold fstep finitial fextract) = Fold step initial extract
...
-                Partial1 sres -> do
-                    let i1 = i + 1
-                        s1 = Tuple' i1 sres
-                    if i1 < n
-                    then return $ Partial1 s1
-                    else Done1 <$> fextract sres
                 Done bres -> return $ Done bres
-                Done1 bres -> return $ Done1 bres
-        | otherwise = Done1 <$> fextract r
+        | otherwise = Done <$> fextract r
```
harendra-kumar, line 583: Remember to fix the `take 0 ` case.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1528,15 +1528,129 @@ groupsBy :: Monad m
     -> Fold m a b
     -> Stream m a
     -> Stream m b
-groupsBy cmp f = foldMany (FL.groupBy cmp f)
+groupsBy cmp f (Stream step state) = Stream (stepOuter f) (Just state, Nothing)
```
harendra-kumar: We can define this as:

```
groupsBy cmp f = parseMany (PR.groupBy cmp f)
```

it can accept a Parser instead of a Fold. Anyway folds can be upgraded to parsers. We only need to make sure that the performance is not affected.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1528,15 +1528,129 @@ groupsBy :: Monad m
...
+    stepOuter _ _ (Nothing,_) = return Stop
 
 {-# INLINE_NORMAL groupsRollingBy #-}
 groupsRollingBy :: Monad m
     => (a -> a -> Bool)
     -> Fold m a b
     -> Stream m a
     -> Stream m b
-groupsRollingBy cmp f = foldMany (FL.groupByRolling cmp f)
+groupsRollingBy cmp f (Stream step state) =
```
harendra-kumar, line 1594: We can define it using parseMany.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1549,7 +1663,44 @@ splitSuffixBy predicate f = foldMany1 (FL.sliceSepBy predicate f)
 
 {-# INLINE_NORMAL wordsBy #-}
 wordsBy :: Monad m => (a -> Bool) -> Fold m a b -> Stream m a -> Stream m b
-wordsBy predicate f = foldMany1 (FL.wordBy predicate f)
+wordsBy predicate f (Stream step state) = Stream (stepOuter f) (Just state)
```
harendra-kumar: We can define it using parseMany and compare if perf remains the same.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -193,10 +193,6 @@ module Streamly.Internal.Data.Fold
 
     -- * Unzipping
     , unzip
-    -- These can be expressed using lmap/lmapM and unzip
```
harendra-kumar, line 187: This is an important fact to remember. Or better we can have an unzip based definition of these in the haddock docs just to remind the user the relation between unzip and these.

harendra-kumar (commented):
We can drop this commit: b7287fbebddd6e7cc02ff89a37c03ee6a86d0acf

src/Streamly/Internal/Data/Fold.hs
adithyaov: Yes, you left this comment earlier (in the closed PR). I've replied by saying: This is a correct implementation and behavior so we might as well keep it.

src/Streamly/Internal/Data/Fold.hs
adithyaov: Yeah, I wanted to return an `error` here. This was on my scratchpad. Good catch.

src/Streamly/Internal/Data/Fold/Types.hs
adithyaov: Yeah, I thought I put a comment.

src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -685,18 +623,10 @@ initialize (Fold step initial extract) = do
...
-        -- XXX runStep is kind of a wierd one. No backtracking possible here.
-        Partial1 s -> return $ Fold step (return s) extract
-        Done b ->
-            return $ Fold (\_ _ -> return $ Done1 b) (return i) (\_ -> return b)
-        Done1 b ->
-            return $ Fold (\_ _ -> return $ Done1 b) (return i) (\_ -> return b)
+runStep ::
+    -- Monad m =>
+    Fold m a b -> a -> m (Fold m a b)
+runStep _ _ = undefined
```
adithyaov, line 638: Change this

src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -651,27 +606,10 @@ ltakeWhile predicate (Fold fstep finitial fextract) =
...
-                  $ Fold (\_ _ -> return $ Done1 b) (return x) (\_ -> return b)
-            Done1 b ->
-                return
-                  $ Done1
-                  $ Fold (\_ _ -> return $ Done1 b) (return x) (\_ -> return b)
-
+duplicate ::
+    -- Monad m =>
+    Fold m a b -> Fold m a (Fold m a b)
+duplicate _ = undefined
```
adithyaov, line 621: Change this

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: We don't want to maintain multiple implementations unless it gives us perf advantage over the parser version.

adithyaov: The advantage is, if the use case can be solved using folds there is no need for the users to look into parsers.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: Yeah, once they are added we can change it.

src/Streamly/Internal/FileSystem/File.hs
adithyaov: File is no longer used, we can look into it when the usecase arrives

src/Streamly/Internal/Data/Fold.hs
adithyaov: Let's leave this for now, let's not remove things that are working properly. We can always remove stuff later.

src/Streamly/Internal/Data/Fold.hs
adithyaov: Let's leave this for now, let's not remove things that are working properly. We can always remove stuff later.

adithyaov: Let's leave this for now, let's not remove things that are working properly. We can always remove stuff later.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: We will do it in a future change. The parser functions are still not defined I believe.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: We will do it in a future change. The parser functions are still not defined I believe.

src/Streamly/Internal/Data/Stream/StreamD.hs
adithyaov: We will do it in a future change. The parser functions are still not defined I believe.

src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1657,27 +1659,30 @@ splitSuffixBy :: Monad m
     => (a -> Bool) -> Fold m a b -> Stream m a -> Stream m b
 splitSuffixBy predicate f = foldMany1 (FL.sliceSepBy predicate f)
 
+data WordsByState s = WordsByJust s | WordsByNothing
+
 {-# INLINE_NORMAL wordsBy #-}
 wordsBy :: Monad m => (a -> Bool) -> Fold m a b -> Stream m a -> Stream m b
```
adithyaov, line 1665: I had to change this anyway in

adithyaov:
See #795

PR #774, opened 2020-11-19
Add scripts to generate hscope db and tags

harendra-kumar:


bin/mk-tags.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+# XXX it does not seem to include the current project anyway
+cat << EOF > $CODEX_CONFIG
+currentProjectIncluded: true
+hackagePath: $HSCOPE_DIR/packages/hackage.haskell.org/
+tagsFileHeader: true
+tagsFileName: tags
+tagsFileSorted: true
+tagsCmd: hasktags --ctags --follow-symlinks --output="\$TAGS" "\$SOURCES"
```
adithyaov, line 19: Also, add `--etags`?

harendra-kumar: You can add emacs option to it. codex has an option to set emacs format, passing that option we can generate that config file and paste it in this script.

PR #789, opened 2020-11-29
Break Benchmark.Prelude.Serial into smaller modules

harendra-kumar:
This module took a lot of time and 4GB RAM to compile. Its now broken
into smaller modules that can be compiled in less than 512 MB RAM.

The easiest way to review it compare each new file in this commit with the old Serial.hs. Each file should be the same as fold file except some code deleted from the old file and a few refactoring changes.


benchmark/Streamly/Benchmark/Prelude/Serial/Elimination.hs
```diff
@@ -0,0 +1,683 @@
...
+import Gauge
+import Streamly.Prelude (SerialT, IsStream, serially)
+import Streamly.Benchmark.Common
+import Streamly.Benchmark.Prelude
+import Prelude hiding (length, sum, or, and, any, all, notElem, elem, (!!),
+    lookup, repeat, minimum, maximum, product, last, mapM_, init)
+import qualified Prelude
+
+{-# INLINE repeat #-}
+repeat :: (Monad m, S.IsStream t) => Int -> Int -> t m Int
```
adithyaov, line 49: This should go under generation.

harendra-kumar: It is in generation as well. It is repeated here only because `the` needs it.

benchmark/Streamly/Benchmark/Prelude/Serial/Transformation2.hs
```diff
@@ -0,0 +1,425 @@
...
+groupsByRollingLT =
+    S.drain . S.groupsByRolling (<) FL.drain
+
+{-# INLINE groupsByRollingEq #-}
+groupsByRollingEq :: MonadIO m => SerialT m Int -> m ()
+groupsByRollingEq =
+    S.drain . S.groupsByRolling (==) FL.drain
+
+o_1_space_grouping :: Int -> [Benchmark]
+o_1_space_grouping value =
```
adithyaov, line 135: Are these O(1) space? How would you represent heap proportional to group size?

adithyaov: This is a mistake from my side.

harendra-kumar: It is O(1) wrt to the stream size not the group size. If something is not O(1) it will be caught in the infinite/long stream benchmarks.

benchmark/streamly-benchmarks.cabal
```diff
@@ -426,6 +435,7 @@ benchmark Memory.Array
 benchmark FileSystem.Handle
   import: bench-options
   type: exitcode-stdio-1.0
+  ghc-options: +RTS -M3000M -RTS
```
adithyaov, line 438: Weren't we specifying memory options via the scripts?

adithyaov (approved):
LGTM.

benchmark/streamly-benchmarks.cabal
harendra-kumar: This for restricting memory consumed by GHC when compiling benchmarks. The one you are talking about heap/memory when running benchmarks.

PR #795, opened 2020-12-01
Enhance the fold type to capture terminating folds 

adithyaov:
A continuation of #770 with squashed commits

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -70,7 +70,7 @@ module Streamly.Internal.Data.Fold
 
     -- ** Partial Folds
     , drainN
-    , drainWhile
+    , drainSepBy
     -- , lastN
     -- , (!!)
     -- , genericIndex
@@ -131,7 +131,7 @@ module Streamly.Internal.Data.Fold
     -- ** Trimming
     , ltake
     -- , lrunFor -- time
-    , ltakeWhile
+    , takeSepBy
     {-
     , ltakeWhileM
     , ldrop
@@ -818,9 +818,9 @@ drainN n = ltake n drain
 
 -- | A fold that drains elements of its input as long as the predicate succeeds,
 -- running the effects and discarding the results.
-{-# INLINABLE drainWhile #-}
-drainWhile :: Monad m => (a -> Bool) -> Fold m a ()
-drainWhile p = ltakeWhile p drain
+{-# INLINABLE drainSepBy #-}
+drainSepBy :: Monad m => (a -> Bool) -> Fold m a ()
+drainSepBy p = takeSepBy p drain
 
 ------------------------------------------------------------------------------
 -- To Elements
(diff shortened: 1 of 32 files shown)


harendra-kumar:
Can you add the `fusion-plugin` flag as well to the `inspection` test CI?

harendra-kumar:
Also  fix the hlint issues.

adithyaov:
@harendra-kumar Ignore the fold benchmarks. It needs to be rewritten in a better way. There is an issue open regarding that.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1178,27 +1406,53 @@ foldCons (Fold stepL beginL doneL) (Fold stepR beginR doneR) =
...
+    -- problems.
+    step [] _ = return $ Done ()
+    step ss a = do
+        !ss1 <- go ss a
+        return
+          $ if Prelude.null ss1
+            then Done ()
+            else Partial ss1
+
+    go [] _ = return []
```
adithyaov, line 1432: This could probably have a better solution. We can change this later.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1382,23 +1706,82 @@ demuxWithDefault_ f kv (Fold dstep dinitial dextract) =
...
-      = case Map.lookup k mp of
-            Nothing -> do
-                acc <- dstep dacc (k, a')
-                return (Tuple' mp acc)
-            Just (Fold step' acc _) -> do
-                _ <- acc >>= \x -> step' x a'
-                return (Tuple' mp dacc)
-    extract (Tuple' mp dacc) = do
+    -- XXX Some code can probably be abstracted
+    step (DemuxingWithDefault dacc bst mp) a =
```
adithyaov, line 1715: This looks ugly.
We should use some `let` statements. Ignore this for now. 
This is changed in `stepify-initial` anyway.

src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -583,54 +586,86 @@ take n (Stream step state) = n `seq` Stream step' (state, 0)
 ------------------------------------------------------------------------------
 
 -- s = stream state, fs = fold state
-data GroupState s fs
+{-# ANN type GroupState Fuse #-}
+data GroupState s fs b a
     = GroupStart s
-    | GroupBuffer s fs Int
-    | GroupYield fs (GroupState s fs)
+    | GroupConsume s fs a
```
adithyaov, line 592: `GroupConsume` is technically not needed but looks like the performance is affected without it. I did not check it with the plugin. But we can eliminate this later. I'd not make any more changes in this PR now.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -302,31 +299,18 @@ either parser = Parser step initial extract
...
-            res <- fstep r a
-            let i1 = i + 1
-                s1 = Tuple' i1 res
-            if i1 < n
-            then return $ Partial 0 s1
-            else Done 0 <$> fextract res
-        | otherwise = Done 1 <$> fextract r
-
-    extract (Tuple' _ r) = fextract r
+take n fld = fromFold $ FL.ltake n fld
```
adithyaov, line 311: Fixed in `stepify-initial`

PR #799, opened 2020-12-04
Add cabal.project file

specdrake:


cabal.project
```diff
@@ -0,0 +1,5 @@
+packages:   streamly.cabal
+          , benchmark/streamly-benchmarks.cabal 
+benchmarks: True
```
harendra-kumar, line 3: Why do we need this? Will this cause benchmarks to be built all the time? If so, that's not what we want. We can always use `--enable-benchmarks` CLI option.

harendra-kumar (changes requested):
You need to add `cabal.project` to `extra-source-files` in `streamly.cabal`, this is the reason for CI failure.

cabal.project
specdrake: Yeah, we don't need ``benchmarks: True`` I'll remove it. I'll add ``cabal.project`` to ``extra-source-files``

PR #810, opened 2020-12-06
Fix takeEQ

harendra-kumar:
takeEQ was not terminating promptly, it was holding up the output till
the next input even though it could have terminated earlier.


adithyaov (approved):

harendra-kumar:
Not sure what the appveyor failure is. Merging this.

PR #823, opened 2020-12-10
Update docs, add unimplemented skeletons

harendra-kumar:
And some formatting changes. There are no functionality changes in this
commit.

### DIFF
--- a/src/Streamly/Internal/Data/Fold/Types.hs
+++ b/src/Streamly/Internal/Data/Fold/Types.hs
@@ -108,7 +108,7 @@
 -- However, removing extract altogether may lead to less optimal code in some
 -- cases because the driver of the fold needs to thread around the intermediate
 -- output to return it if the stream stops before the fold could @Done@.  When
--- using this approach, the @splitParse (FL.take filesize)@ benchmark shows a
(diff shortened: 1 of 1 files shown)


src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -375,6 +437,42 @@ teeWithFst = undefined
...
+longest :: -- Monad m =>
+    Fold m x a -> Fold m x a -> Fold m x a
+longest _f1 _f2 = undefined
+
+-- | Map a 'Fold' returning function on the result of a 'Fold'.
+--
+-- Compare with 'Monad' instance method '>>='. This implementation allows
+-- stream fusion but has quadratic complexity. This can fuse with other
+-- operations and can be much faster than 'Monad' instance for small number
+-- (less than 8) of compositions.
```
adithyaov, line 467: How do you know this?

adithyaov (approved):

src/Streamly/Internal/Data/Fold/Types.hs
harendra-kumar: Magic! This is from parsers.

PR #826, opened 2020-12-10
Review/update docs, combinators in Fold module

harendra-kumar:
Data.Fold.Types:
* Remove takeSepBy:  This function does not do what the name. Also,
`takeSepBy p` is the same as (sliceSepBy (not . p)). We can just remove
this in favor of sliceSepBy.

Data.Fold:

* Reorganize exports
* Add some skeleton/unimplemented combinators.
* Add/update haddock documentation
* Add sliceSepByMax fold - It was in parsers but it can be implemented
  as a fold.
* Rename drainSepBy to drainWhile
* Rename sliceSepWith to sliceEndWith
* Remove distribute_ as it can be implemented using distribute and has
  not perf advantage.
* Remove demuxWithDefault_ and implement demuxDefaultWith instead, which
  does not discard the results. Discarding results has no perf
  advantage.
* Implement demuxWith in terms of demuxDefaultWith, it has no perf
  advantage over the latter.
* Remove demuxWith_, use demuxWith instead.
* Remove demuxDefault_, use demuxDefault instead.
* Some code/comment cleanups

Benchmarks:

* Changes corresponding to the source changes
* Some code cleanups
* Reorder benchmarks

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -6,9 +6,8 @@
 -- Maintainer  : streamly@composewell.com
 -- Stability   : experimental
 -- Portability : GHC
-
--- Also see the "Streamly.Internal.Data.Sink" module that provides specialized
--- left folds that discard the outputs.
+--
+-- See "Streamly.Data.Fold" for an overview.
 --
 -- IMPORTANT: keep the signatures consistent with the folds in Streamly.Prelude
 
@@ -18,13 +17,7 @@ module Streamly.Internal.Data.Fold
       Step (..)
     , Fold (..)
 
-    , hoist
-    , generally
-
-    -- , tail
-    -- , init
-
-    -- * Fold Creation Utilities
+    -- * Creating
     , mkAccum
     , mkAccum_
     , mkAccumM
@@ -34,19 +27,23 @@ module Streamly.Internal.Data.Fold
     , mkFoldM
     , mkFoldM_
 
-    -- ** Full Folds
+    -- * Generators
+    , yield
+    , yieldM
+
+    -- * Accumulators
+    -- ** Semigroups and Monoids
+    , sconcat
+    , mconcat
+    , foldMap
+    , foldMapM
+
+    -- ** Reducers
     , drain
     , drainBy
     , drainBy2
     , last
     , length
-    , sum
-    , product
-    , maximumBy
-    , maximum
-    , minimumBy
-    , minimum
-    -- , the
     , mean
     , variance
     , stdDev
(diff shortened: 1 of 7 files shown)


adithyaov:
> Rename drainSepBy to drainWhile

It consumes extra input here, hence was renamed.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -55,22 +52,25 @@ module Streamly.Internal.Data.Fold
...
+    -- ** Collectors
     , toList
     , toListRevF  -- experimental
+    , toParallelSVar
 
-    -- ** Partial Folds
+    -- * Terminating Folds
     , drainN
-    , drainSepBy
+    , drainWhile
```
adithyaov, line 73: I renamed this because input is lost here. Maybe we should just remove this API. `drainSepBy` is ugly and I don't see the need for a function that drains even the seperator.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -818,9 +827,9 @@ drainN n = ltake n drain
 
 -- | A fold that drains elements of its input as long as the predicate succeeds,
 -- running the effects and discarding the results.
-{-# INLINABLE drainSepBy #-}
-drainSepBy :: Monad m => (a -> Bool) -> Fold m a ()
-drainSepBy p = takeSepBy p drain
+{-# INLINABLE drainWhile #-}
```
adithyaov, line 830: `drainWhile`'s behaviour does not justify its name. An extra element is drained.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1695,164 +1714,119 @@ demux :: (Monad m, Ord k)
...
-    => (a -> (k, a')) -> Map k (Fold m a' b) -> Fold m (k, a') b -> Fold m a ()
-demuxWithDefault_ f kv (Fold dstep dinitial dextract) =
+-- | Like 'demuxWith' but uses a default catchall fold to handle inputs which
+-- do not have a specific fold in the map to handle them.
+--
+-- If any fold in the map stops, inputs meant for that fold are sent to the
+-- catchall fold. If the catchall fold stops then inputs that do not match any
+-- fold are ignored.
+--
+-- Stops when all the folds stop.
```
adithyaov, line 1729:
```suggestion
-- Stops when all the folds (including the catchall fold) stop.
```

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: I was thinking of using this when you anyway discard the rest of the stream. But that's not necessarily true. We can either remove it or move it to parsers.

harendra-kumar: `drainSepBy` is a misleading name, it seems to indicate the condition is for `sepBy` but it is the opposite.

adithyaov: We can move this to parsers.

test/Streamly/Test/Data/Fold.hs
```diff
@@ -1,13 +1,15 @@
 module Main (main) where
 
+import Streamly.Test.Common (checkListEqual)
```
adithyaov, line 3: This import causes CI to fail. Please check.

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: done

test/Streamly/Test/Data/Fold.hs
harendra-kumar: Fixed.

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: Moved to parsers.

PR #828, opened 2020-12-10
Implement Data.Parser.ParserD.wordBy

adithyaov:
- Modify Data.Parser.wordBy accordingly
- Add tests for both
- Add benchmarks for both

This requires `fusion-plugin`

```
benchmarked Data.Parser.ParserD/o-1-space/wordBy
time                 65.46 μs   (64.92 μs .. 65.92 μs)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 65.82 μs   (65.47 μs .. 66.22 μs)
std dev              1.856 μs   (1.475 μs .. 2.670 μs)
variance introduced by outliers: 22% (moderately inflated)

benchmarked Data.Parser/o-1-space/wordBy
time                 64.97 μs   (64.54 μs .. 65.34 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 66.02 μs   (65.67 μs .. 66.49 μs)
std dev              1.967 μs   (1.637 μs .. 2.780 μs)
variance introduced by outliers: 24% (moderately inflated)
```

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/Parser.hs
+++ b/benchmark/Streamly/Benchmark/Data/Parser.hs
@@ -76,6 +76,14 @@ takeWhile value = IP.parse (PR.takeWhile (<= value) FL.drain)
 groupBy :: MonadCatch m => SerialT m Int -> m ()
 groupBy = IP.parse (PR.groupBy (<=) FL.drain)
 
+{-# INLINE wordBy #-}
+wordBy :: MonadCatch m => Int -> SerialT m Int -> m ()
+wordBy value = IP.parse (PR.wordBy (>= value) FL.drain)
+
+{-# INLINE manyWordByEven #-}
+manyWordByEven :: MonadCatch m => SerialT m Int -> m ()
+manyWordByEven = IP.parse (PR.many FL.drain (PR.wordBy (Prelude.even) FL.drain))
+
 {-# INLINE many #-}
 many :: MonadCatch m => SerialT m Int -> m Int
 many = IP.parse (PR.many FL.length (PR.satisfy (> 0)))
@@ -263,11 +271,13 @@ o_1_space_serial value =
     , benchIOSink value "takeWhile" $ takeWhile value
     , benchIOSink value "drainWhile" $ drainWhile value
     , benchIOSink value "groupBy" $ groupBy
+    , benchIOSink value "wordBy" $ wordBy value
     , benchIOSink value "splitAp" $ splitAp value
     , benchIOSink value "splitApBefore" $ splitApBefore value
     , benchIOSink value "splitApAfter" $ splitApAfter value
     , benchIOSink value "splitWith" $ splitWith value
     , benchIOSink value "many" many
+    , benchIOSink value "many (wordBy even)" $ manyWordByEven
     , benchIOSink value "some" some
     , benchIOSink value "manyTill" $ manyTill value
     , benchIOSink value "tee" $ teeAllAny value
--- a/src/Streamly/Internal/Data/Parser/ParserD.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserD.hs
@@ -447,15 +447,46 @@ sliceBeginWith ::
     (a -> Bool) -> Fold m a b -> Parser m a b
 sliceBeginWith = undefined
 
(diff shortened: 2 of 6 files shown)


src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -493,13 +494,34 @@ sliceSepByMax p n = sliceSepBy p . FL.ltake n
...
+{-# INLINE wordBy #-}
+wordBy :: Monad m => (a -> Bool) -> Fold m a b -> Parser m a b
+wordBy predicate (Fold fstep finitial fextract) = Parser step initial extract
+
+    where
+
+    initial = Left' <$> finitial
+
+    step (Left' s) a =
+        if not (predicate a)
```
harendra-kumar, line 507: We should consume any leading space, e.g. "   word  " should be parsed as ["word"] and not ["", "word"], see the haddock docs of `wordBy` parser.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -493,13 +494,34 @@ sliceSepByMax p n = sliceSepBy p . FL.ltake n
...
+        then do
+            res <- fstep s a
+            return
+                $ case res of
+                      FL.Partial s1 -> Partial 0 $ Left' s1
+                      FL.Done b -> Done 0 b
+        else do
+            ex <- fextract s
+            return $ Partial 0 $ Right' ex
```
harendra-kumar, line 516: We should yield the result as early as possible instead of waiting for all the spaces after it to be over. We can declare done as soon as we see the first space.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -499,13 +500,36 @@ sliceSepByMax p n = sliceSepBy p . FL.ltake n
...
+    step (Left' s) a =
+        if not (predicate a)
+        then worder s a
+        else return $ Partial 0 $ Left' s
+    step (Right' s) a =
+        if not (predicate a)
+        then worder s a
+        else do
+            ex <- fextract s
+            return $ Done 1 $ ex
```
harendra-kumar, line 529: Since we are finishing right after the word ends, as I suggested earlier, this will result into an empty word being emitted in the end if there is white space in the end.

We need a test case that when this parser is used on a stream of a word surrounded by whitespace e.g. " hello " results in a single word being emitted i.e. ["hello"].

For short scopes, we can just use `r` for a temporary result instead of using different names every time.

benchmark/Streamly/Benchmark/Data/Parser.hs
```diff
@@ -79,6 +79,10 @@ takeEQ value = IP.parse (PR.takeEQ value FL.drain)
 takeWhile :: MonadCatch m => Int -> SerialT m Int -> m ()
 takeWhile value = IP.parse (PR.takeWhile (<= value) FL.drain)
 
+{-# INLINE wordBy #-}
+wordBy :: MonadCatch m => Int -> SerialT m Int -> m ()
+wordBy value = IP.parse (PR.wordBy (>= value) FL.drain)
```
harendra-kumar, line 84: Use `wordBy even` to stress the implementation for per word overhead by finding many short words.

harendra-kumar (changes requested):

benchmark/Streamly/Benchmark/Data/Parser.hs
adithyaov: I'll add both, The current benchmark is the minimal, check-the-fusion benchmark.

adithyaov:
```
benchmarked Data.Parser.ParserD/o-1-space/many (wordBy even)
time                 1.209 ms   (1.204 ms .. 1.214 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.209 ms   (1.205 ms .. 1.213 ms)
std dev              13.39 μs   (11.03 μs .. 18.46 μs)
```

pranaysashank (changes requested):

pranaysashank (approved):

PR #837, opened 2020-12-13
 Separate streamly tests into another package

harendra-kumar:
To be able to use an internal test library, we need to have it as a
separate package. See comments in streamly-benchmarks.cabal for details.

Use a test library for common test modules
Also, added "opt" flag to control optimization levels.

Add a README for tests.
Update benchmarks readme.


test/README.md
```diff
@@ -0,0 +1,52 @@
+# Build only
```
pranaysashank, line 1:
```suggestion
# Build only, use any of the following
```

benchmark/README.md
```diff
@@ -1,11 +1,56 @@
-## Running Benchmarks
+## Build Only
+
+Building single benchmark suite:
```
pranaysashank, line 3:
```suggestion
Building single benchmark suite: use any of the following
```

.circleci/config.yml
```diff
@@ -193,6 +194,7 @@ jobs:
             command: |
               apt-get install -y ghc-8.2.2
               apt-get install -y cabal-install-3.2
+              apt-get install libtinfo-dev
```
pranaysashank, line 197: what's this needed for?

pranaysashank (commented):
Need to check why the coverage build is failing to run any tests.

.circleci/config.yml
harendra-kumar: The linker was failing.

PR #838, opened 2020-12-14
Update INSTALL.md

specdrake:
Added cabal repl --ghc-options "-fobject-code" note

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -242,6 +242,12 @@ $ cabal repl
 Ok, modules loaded: Main.
 *Main>
 ```
+NOTE : If you get the following error:
+```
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -242,6 +242,12 @@ $ cabal repl
 Ok, modules loaded: Main.
 *Main>
 ```
+NOTE : If you get the following error:
```
harendra-kumar, line 245: Why "If"? Doesn't it come all the time, because we are using FFI? We can directly recommend `cabal repl --ghc-options "-fobject-code"` in the above example.

specdrake: It didn't come when I followed the instructions on my local machine.

PR #840, opened 2020-12-15
Add the test package to coverage/inspection tests

harendra-kumar:


benchmark/README.md
```diff
@@ -78,7 +78,7 @@ You must be in the repo root directory to run these commands.
 Run the default benchmark suites:
 
 ```
-$ ./bench.sh --quick
+$ bench.sh --quick
```
adithyaov, line 81: `bin/bench.sh`?
Is it symlinked?

default.nix
```diff
@@ -90,11 +90,9 @@ let haskellPackages =
         # XXX On macOS cabal2nix does not seem to generate a dependency on
         # Cocoa framework.
         buildInputs =
-            [nixpkgs.haskellPackages.inspection-testing] ++
-            (if builtins.currentSystem == "x86_64-darwin"
-             then [nixpkgs.darwin.apple_sdk.frameworks.Cocoa]
-             else []
-            );
+            if builtins.currentSystem == "x86_64-darwin"
```
adithyaov, line 93: Why remove this?

bin/build-lib.sh
```diff
@@ -0,0 +1,207 @@
...
+  then
+    echo $1
+  fi
+}
+
+# A group consisting of all known individual targets
+all_grp () {
+  { for i in $GROUP_TARGETS
+    do
+      eval "echo \$$i"
```
adithyaov: Why can't you echo normally here?

bin/bench.sh
```diff
@@ -413,6 +413,8 @@ run_reports() {
 # Execution starts here
 #-----------------------------------------------------------------------------
 
+cd $BENCH_SH_DIR/..
```
adithyaov: Oh, I see. We are going up a directory.

benchmark/README.md
adithyaov: The previous statement says `You must be in the repo root directory to run these commands.`
You should be in the benchmarks directory?

adithyaov (approved):
LGTM. Just went through it. I might have missed something.

default.nix
harendra-kumar: It does not work as expected. We need `inspection-testing` to be installed as a ghc package in the shell which it does not.

bin/build-lib.sh
harendra-kumar: We have to echo the values of the variable names.

bin/test.sh
```diff
@@ -0,0 +1,159 @@
...
+  echo "Usage: $0 "
+  echo "       [--targets <"target1 target2 ..." | help>]"
+  echo "       [--slow]"
+  echo "       [--quick]"
+  echo "       [--cabal-build-options <option>]"
+  echo "       -- <hspec options or test names>"
+  echo
+  echo "--targets: targets to run, use 'help' for list of targets"
+  echo "--slow: take longer but run more tests"
+  echo "--quick: faster  but run fewer tests"
```
pranaysashank, line 19: is it fewer test suites or fewer number of quickcheck iterations ?

harendra-kumar: It is a to-do, does not anything right now. Just retained this from benchmarks script.

PR #841, opened 2020-12-17
 Revert to using a separate GroupConsume state 

harendra-kumar:


src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -588,6 +588,7 @@ take n (Stream step state) = n `seq` Stream step' (state, 0)
 {-# ANN type GroupState Fuse #-}
 data GroupState s fs b a
     = GroupStart s
+    | GroupConsume s fs a
```
adithyaov, line 591: Maybe add some developer notes saying that this is an extra state needed for fusion. 

My vague guess is that it acts as a loop breaker.

Issue #456, opened 2020-03-22
A combinator for chunksOf with timeout


harendra-kumar:
We need a combinator that can chunk a stream to fixed size chunks but if the chunk is not completed within a certain time period it will still emit the chunk.

This combinator can useful e.g. to write a stream to stdout or a file but flush it within a certain time period.

I wrote this as stop gap solution, but this is not ideal. It also has an undesired behavior.

```
{-# LANGUAGE FlexibleContexts #-}
module Main (main) where

import Data.Char (ord)
import Data.Maybe (isNothing)
import System.IO (stdin, stdout)

import Streamly
import qualified Streamly.Internal.Prelude as S
import qualified Streamly.Internal.Data.Fold as FL
import qualified Streamly.Internal.Data.Parse as PR
import qualified Streamly.Internal.FileSystem.Handle as FH
import qualified Streamly.Internal.Data.Unicode.Stream as U


{-# INLINE chunksOfTimeout #-}
chunksOfTimeout :: (IsStream t, MonadAsync m)
    => Int -> Double -> FL.Fold m a b -> t m a -> t m b
chunksOfTimeout n t f =
      S.splitWithSuffix isNothing (FL.lcatMaybes f)
    . S.interjectSuffix t (return Nothing)
    . S.intersperseSuffixBySpan n (return Nothing)
    . S.map Just

main = do
      FH.putLines U.encodeLatin1
    $ chunksOfTimeout 10 5 FL.toList
    $ U.decodeLatin1
    $ S.unfold FH.read stdin
```

We also have this unimplemented comment in the FileSystem.Handle module;

```
-- | Write collecting the input in sessions of n seconds or if chunkSize
-- gets exceeded.
{-# INLINE writeByChunksOrSessionsOf #-}
writeByChunksOrSessionsOf :: MonadIO m
    => Int -> Double -> Handle -> SerialT m Word8 -> m ()
writeByChunksOrSessionsOf chunkSize sessionSize h m = undefined
```

This can potentially be implemented as a parser as well.

harendra-kumar:
We also need to think about changing the behavior of default `write` folds to include a timeout. Quite often programmers run into an issue where the data is stuck in the buffer of the `write` routine. If we add a reasonable timeout by default and provide other APIs for better control, it may help avoid such issues.

harendra-kumar:
[...]

PR #842, opened 2020-12-18
Add combinators for Timeout Buffered writer stream

rnjtranjan:
With Timeout combinators 45,000,000  Bytes written to socket and then drained out in **142.4000** seconds vs **122.0099** secs without Timeout combinators.

### DIFF
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
@@ -92,12 +92,11 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Type
     , bytesToElemCount
     , memcpy
     , memcmp
-<<<<<<< HEAD:src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
     , c_memchr
-=======
     , bytesToElemCount
     , writeNUnsafeMaybe
->>>>>>> 7e73b7af (Add combinators for Timeout Buffered writer stream):src/Streamly/Internal/Data/Array/Storable/Foreign/Mut/Types.hs
+    , writeMaybesN
+    , writeMaybes
     )
 where
 
@@ -1163,8 +1162,6 @@ toStreamD_ size Array{..} =
                     return r
         return $ D.Yield x (p `plusPtr` size)
 #endif
-<<<<<<< HEAD:src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
-=======
     Array a
 nil = Array nullForeignPtr (Ptr nullAddr#) (Ptr nullAddr#)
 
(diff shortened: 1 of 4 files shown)


src/Streamly/Internal/Network/Socket.hs
```diff
@@ -497,6 +498,17 @@ writeStrings encode h =
 fromBytesWithBufferOf :: MonadIO m => Int -> Socket -> SerialT m Word8 -> m ()
 fromBytesWithBufferOf n h m = fromChunks h $ AS.arraysOf n m
 
+-- | Transform a stream of @a@ to @(Just a)@ then intersperse a Nothing into the input 
+-- stream after every n elements then interject Nothing every @t@ seconds.
+-- @since 0.8.0
+{-# INLINE chunksOfTimeout #-}
+chunksOfTimeout :: (MonadIO m , IsStream t, MonadAsync m)
+    => Int -> Double -> t m a -> t m (Maybe a)
+chunksOfTimeout n t =
```
harendra-kumar, line 507: 1. This should go in the same module as `chunksOf`/`intervalsOf`. 
2. We could have a corresponding `chunksOf` as well (i.e. without a timeout), but we already have a `chunksOf`. So should we call these `justsOf` and `justsOfTimeout` to indicate contiguous chunks of `Just` values?

harendra-kumar (changes requested):
We also need corresponding APIs in the Handle/File modules, that could be a separate PR though. Handle module has benchmarks as well so we can see how it performs.

rnjtranjan:
a PR https://github.com/composewell/streamly/pull/1236 for FileHandle has been created.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1157,3 +1160,56 @@ toStreamD_ size Array{..} =
...
+        when (elemCount < 0) $ error "toArrayMinChunk: elemCount is negative"
+        liftIO $ newArrayAligned alignSize elemCount
+    step arr@(Array start end bound) x | end == bound = do
+        let p = unsafeForeignPtrToPtr start
+            oldSize = end `minusPtr` p
+            newSize = max (oldSize * 2) 1
+        arr1 <- liftIO $ reallocAligned alignSize newSize arr
+        insertElem arr1 x
+    step arr x = insertElem arr x
+    extract = liftIO . shrinkToFit
```
harendra-kumar, line 1215: We should not be writing new custom combinators for this job. We can use existing fold/parse combinators.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -1666,3 +1668,22 @@ applyAsync = (|$)
...
+-- | Transform a stream of @a@ to @(Just a)@ then intersperse a Nothing into the input
+-- stream after every n elements then interject Nothing every @t@ seconds.
+-- @since 0.8.0
+{-# INLINE justsOfTimeout #-}
+justsOfTimeout :: (MonadIO m, IsStream t, MonadAsync m)
+    => Int -> Double -> t m a -> t m (Maybe a)
+justsOfTimeout n t =
+      interjectSuffix t (return Nothing)
+    . intersperseSuffixBySpan n (return Nothing)
+    . justsOf
```
harendra-kumar, line 1689: As mentioned in #456 this has an undesired behavior. The timeout may occur after the span is emitted and the timeout would force eject a smaller span. Ideally we should reset the timer whenever we emit a span.

harendra-kumar (commented):
I am going to close this PR and create a new one as a replacement.

harendra-kumar:
Closing this, it should be implemented differently, I will create a new one.

PR #845, opened 2020-12-20
Implement sliceSepBy

specdrake:


src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -433,9 +433,28 @@ takeWhile1 predicate (Fold fstep finitial fextract) =
...
+
+    step s a =
+        if cond a
+        then do
+            res <- pextract s
+            return $ Done 1 res
+        else do
+            res <- pstep s a
+            case res of
+                Error _ -> do
```
pranaysashank, line 454: I don't think you should extract when an Error is encountered, you should likely just return the Error.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -433,9 +433,28 @@ takeWhile1 predicate (Fold fstep finitial fextract) =
...
+
+    where
+
+    initial = pinitial
+
+    step s a =
+        if cond a
+        then do
+            res <- pextract s
+            return $ Done 1 res
```
adithyaov, line 450: I believe, `sliceSepBy` drops the seperator.
```suggestion
            return $ Done 0 res
```

adithyaov:
@specdrake Can you make any final changes, rebase/reword commits? Please ping me once that is done.

specdrake:
> @specdrake Can you make any final changes, rebase/reword commits? Please ping me once that is done.

It can be merged, I think.

adithyaov:
In general, it is better to avoid commit messages like "Fix some code"
For small combinators, you keep everything in one commit or if the tests or benchmarks are big, spread it over 3 commits.
Instead of "Add benchmarks",  "Add benchmarks for sliceSepBy" is more informative.
"Add tests for sliceSepBy" instead of "Add tests"
I'll reword/fixup the commits

specdrake:
> In general, it is better to avoid commit messages like "Fix some code"
> For small combinators, you keep everything in one commit or if the tests or benchmarks are big, spread it over 3 commits.
> Instead of "Add benchmarks", "Add benchmarks for sliceSepBy" is more informative.
> "Add tests for sliceSepBy" instead of "Add tests"
> I'll reword/fixup the commits

Thanks, I'll keep these in mind in the future.

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -487,9 +487,22 @@ takeWhile1 predicate (Fold fstep finitial fextract) =
 --
 -- /Internal/
 --
-sliceSepBy :: -- MonadCatch m =>
+sliceSepBy :: MonadCatch m =>
     (a -> Bool) -> Parser m a b -> Parser m a b
-sliceSepBy _cond = undefined
+sliceSepBy cond (Parser pstep pinitial pextract) =
```
adithyaov, line 492: This should be named `sliceSepByP` I believe.

benchmark/Streamly/Benchmark/Data/Parser.hs
```diff
@@ -161,6 +161,12 @@ split_ value =
             (PR.drainWhile (<= value))
         )
 
+{-# INLINE sliceSepBy #-}
+sliceSepBy :: MonadCatch m
+    => Int -> SerialT m Int -> m()
+sliceSepBy value = IP.parse (PR.sliceSepBy (>= value) (PR.many FL.drain
+                    (PR.satisfy (const True))))
```
adithyaov, line 168: Indendation,
```
sliceSepBy value =
    IP.parse
        (PR.sliceSepBy (>= value) (PR.many FL.drain (PR.satisfy (const True))))
```
Why do you use `PR.many` there?
Two benchmarks of interest would be,
```
 IP.parse (PR.sliceSepBy (>= value) (PR.satisfy (const True))
```
to just see if it fuses. Should be compared to `takeWhile`
```
 IP.parse (PR.many FL.drain (PR.sliceSepBy ((== 0) . `mod` (value `div` 1000)) (PR.satisfy (const True))))
```
which is a little more complex use case.

benchmark/Streamly/Benchmark/Data/Parser/ParserD.hs
```diff
@@ -100,6 +100,10 @@ someAlt xs = do
     x <- IP.parseD (AP.some (PR.satisfy (> 0))) xs
     return $ Prelude.length x
 
+{-#INLINE sliceSepBy #-}
```
adithyaov, line 103: Same comments as above

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -487,9 +487,22 @@ takeWhile1 predicate (Fold fstep finitial fextract) =
...
+
+    Parser step initial pextract
```
adithyaov, line 494: You can put it on the same line.
```suggestion
sliceSepBy cond (Parser pstep pinitial pextract) = Parser step initial pextract
```

test/Streamly/Test/Data/Parser.hs
```diff
@@ -311,6 +311,17 @@ takeProperties =
 --                         where
 --                             list_length = Prelude.length ls
 
+sliceSepBy :: Property
+sliceSepBy =
+    forAll (listOf (chooseInt (min_value, max_value )))  $ \ls ->
+        case S.parse (P.sliceSepBy predicate prsr) (S.fromList ls) of
+            Right parsed_list -> checkListEqual parsed_list (Prelude.takeWhile
+                        (not. predicate) ls)
```
adithyaov, line 319:
```suggestion
            Right parsed_list -> 
                checkListEqual parsed_list (Prelude.takeWhile (not. predicate) ls)
```

src/Streamly/Internal/Data/Parser/ParserD.hs
harendra-kumar: It could be `sliceSepByP` though there is no `sliceSepBy` so we do not necessarily need the distinction. In some other combinators we use the P suffix where we collect the output using a Parser instead of Fold. If we want that convention to be consistent we can add `P` suffix.

benchmark/Streamly/Benchmark/Data/Parser.hs
adithyaov: Maybe use,
```
PR.fromFold FL.drain
```
instead of,
```
PR.many FL.drain (PR.satisfy (const True))
```

src/Streamly/Internal/Data/Parser/ParserD.hs
specdrake: Should I rename it, then?

adithyaov: Yes, we should rename it.

src/Streamly/Internal/Data/Parser/ParserD.hs
specdrake: It reaches column 80 because of renaming sliceSepBy to sliceSepByP

test/Streamly/Test/Data/Parser.hs
specdrake: It goes over the 79 columns limit.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -476,23 +476,24 @@ takeWhile1 cond = K.toParserK . D.takeWhile1 cond
...
+-- | @sliceSepByP cond parser@ parses a slice of the input using @parser@ until
 -- @cond@ succeeds or the parser stops.
 --
 -- This is a generalized slicing parser which can be used to implement other
 -- parsers e.g.:
 --
 -- @
--- sliceSepByMax cond n p = sliceBy cond (take n p)
--- sliceSepByBetween cond m n p = sliceBy cond (takeBetween m n p)
+-- sliceSepByMax cond n p = sliceSepByP cond (take n p)
```
harendra-kumar, line 486: `sliceSepByP cond (fromFold $ take n p)`

harendra-kumar (approved):

PR #846, opened 2020-12-21
Break IsStream and StreamD into smaller modules.

pranaysashank:


src/Streamly/Internal/Data/Stream/IsStream.hs
```diff
@@ -599,55 +599,42 @@ module Streamly.Internal.Data.Stream.IsStream
...
        ( AbsTime, MilliSecond64(..), addToAbsTime, toRelTime
-       , toAbsTime, TimeUnit64, RelTime64, addToAbsTime64)
-import Streamly.Internal.Data.IORef.Prim (Prim, IORef)
+       , toAbsTime, RelTime64)
 import Streamly.Internal.Data.Tuple.Strict (Tuple'(..))
 
+import Streamly.Internal.Data.Stream.IsStream.Generate
+import Streamly.Internal.Data.Stream.IsStream.Eliminate
+import Streamly.Internal.Data.Stream.IsStream.Transform
+import Streamly.Internal.Data.Stream.IsStream.Exceptions
```
adithyaov, line 636: Maybe you could move these after the qualified imports. Not a big deal, you can ignore this.

src/Streamly/Internal/Data/Stream/IsStream.hs
```diff
@@ -676,3678 +654,250 @@ import Prelude hiding
...
--- | Unlike @groupsBy@ this function performs a rolling comparison of two
--- successive elements in the input stream. @groupsByRolling cmp f $ S.fromList
--- [a,b,c,...]@ assigns the element @a@ to the first group, if @a \`cmp` b@ is
--- 'True' then @b@ is also assigned to the same group.  If @b \`cmp` c@ is
--- 'True' then @c@ is also assigned to the same group and so on. When the
--- comparison fails a new group is started. Each group is folded using the fold
--- @f@.
+{-# INLINE drainN #-}
+drainN :: Monad m => Int -> SerialT m a -> m ()
+drainN n = drain . take n
```
adithyaov, line 687: Shouldn't this be under `Eliminate`?

src/Streamly/Internal/Data/Stream/IsStream.hs
```diff
@@ -676,3678 +654,250 @@ import Prelude hiding
...
--- > [[1,1],[2,2]]
+-- Run a stream as long as the predicate holds true.
 --
 -- @since 0.7.0
-{-# INLINE groups #-}
-groups :: (IsStream t, Monad m, Eq a) => Fold m a b -> t m a -> t m b
-groups = groupsBy (==)
+{-# INLINE drainWhile #-}
+drainWhile :: Monad m => (a -> Bool) -> SerialT m a -> m ()
+drainWhile p = drain . takeWhile p
```
adithyaov, line 709: `Eliminate`?

src/Streamly/Internal/Data/Stream/IsStream.hs
```diff
@@ -676,3678 +654,250 @@ import Prelude hiding
...
--- [""]
---
--- >>> splitOn' (== '.') "."
--- ["",""]
+-- /Internal/
+{-# INLINE interjectSuffix #-}
+interjectSuffix
+    :: (IsStream t, MonadAsync m)
+    => Double -> m a -> t m a -> t m a
+interjectSuffix n f xs = xs `Par.parallelFst` repeatM timed
```
adithyaov, line 739: Why not put this in `Transform`?

adithyaov:
Can you comment on the way you determine, which module to put which function?

src/Streamly/Internal/Data/Stream/IsStream.hs
pranaysashank: The functions that should be in a separate module but aren't is mainly because the function depends on another independent module, like a function in Eliminate depending on a function in Generate. Since I wanted all these modules to not depend on each other, I moved them into the IsStream module. Later while refactoring StreamD I moved any such functions into a common StreamD module

pranaysashank: The classification of which function goes into which module is based on existing module structure is commented (haddocks) in the export list.

harendra-kumar (changes requested):
In the `IsStream.Eliminate` module we should keep only those functions that return the shape`m a`. We can move out the following to transformations:

* foldrS
* foldrT
* dropPrefix
* dropInfix
* dropSuffix

We can possibly push `mapM_` in IsStream as it is a mixed operation with both transformation and elimination involved. This is the only operation in `Transformation` module that does not have a `t m a` return type.

In StreamD instead of using a `Common` module we could push these to relevant modules and then push the complex operations (involving two different types of operations e.g. elimination and transformation) from those modules in `StreamD`, in the same way as we did in `IsStream`. We are taking two different approaches to the same problem in IsStream vs StreamD. Alternatively, we can merge the common module with StreamD.Type module, anwyay that module is already acting as a Common module.

StreamD.Eliminate, we can move these to transformations, that is also consistent with IsStream:

* foldrS, foldrT, foldlS, foldlT
* reverse, reverse'
* parseMany, parseIterate
* tap et al
* Conversions

StreamD.Generate, we can move fromStreamK, fromStreamD, toStreamD to transformations, probably next to wherever we put other adaptations like hoist, generally, liftInner etc. But these are also exported from StreamD.Type so we can just keep them there as well.

Merging, `zipWith` operations can be moved to `Flatten`, and we can rename the module to `Combine`. Even `SplitGroup` could be combined in the same module and we can rename the module to `Nesting`.

`eqBy`, `cmpBy` can be moved to `Eliminate`, they are multi-stream eliminations.
`the` can be moved to `Eliminate`.
`newCallbackStream` can be moved to `Generate`.

We should try to have the same modules in StreamD and IsStream.

Also, in a separate issue these two modules could be moved under `IsStream`:

 * Streamly.Internal.Data.Stream.Combinators
 * Streamly.Internal.Data.Stream.Enumeration

harendra-kumar:
Proposed naming of modules:

* Generation
* Elimination
* Transformation
* Nesting (transformation)
* Splitting (transformation)
* Exception (transformation)

pranaysashank:
New changes are from this commit,  `Merge StreamD.Common and StreamD.Type modules.`

pranaysashank:
Before merging SplitGroup, Flatten module.
Flatten - 109 M
SplitGroup - 266 M

After merging both into Nesting,
Nesting - 445M

Before new changes, `cabal build` reported 934 M 
After changes - 993 M

harendra-kumar:
Can you run the benchmarks before and after and check if there are any significant differences?

harendra-kumar:
Would you like to address the module naming as well in this PR?

PR #855, opened 2020-12-30
Fix target groups not working correctly

harendra-kumar:
RUNNING_BENCHMARKS/RUNNING_TESTS variables were getting cleared up by
set_common_vars. Also, all_grp is a function and should be treated
differently than variables.


adithyaov (commented):
Dint test it but looks good. Maybe fix memory restrictions for benchmarks as well?

PR #860, opened 2021-01-07
fix some stack space overflows

specdrake:


bin/bench-exec-one.sh
```diff
@@ -72,6 +73,7 @@ bench_rts_opts_specific () {
     Prelude.WAsync/o-n-space/monad-outer-product/*) echo -n "-K4M" ;;
 
     # XXX need to investigate these, taking too much stack
+    Prelude.Rate/o-1-space/asyncly/minRate/*) echo -n "-K128K" ;;
```
harendra-kumar, line 76: The comment above is misleading. You can insert this line using alphabetic order, i.e. before Prelude.Serial.

bin/bench-exec-one.sh
```diff
@@ -45,6 +45,7 @@ bench_rts_opts_default () {
 bench_rts_opts_specific () {
   case "$1" in
     Data.Stream.StreamK/o-n-space/elimination/toList) echo -n "-K2M" ;;
+    Data.Stream.StreamD/o-n-space/elimination/toList) echo -n "-K2M" ;;
```
harendra-kumar, line 48: This line can also be move up to maintain alphabetic order.

specdrake:
This can be merged.

harendra-kumar (approved):

bin/bench-exec-one.sh
specdrake: @harendra-kumar I just discovered it should be `Prelude.Rate/o-1-space/*) echo -n "-K128K" ;;` otherwise it will stack overflow(maxRate) without ``--quick`` option.
`

Issue #767, opened 2020-11-15
FSEvents: A "Ready"/"Begin" event for synchronization


harendra-kumar:
In some cases we may want to kick off filesystem operations after we have started listening for the events. One way to synchronize this is to have a special fs event generated at the beginning marking that we are now ready to listen for the events. After receiving this event the event handler can signal the application modifying the file system to start off.

This problem is faced in the fs events tests as well where we are currently using a delay after the event listener starts, however, that is not a very reliable mechanism.

We need this for Linux/Windows/macOS all implementations.

harendra-kumar:
Updating discussion from the PR:

We need to generate an additional event in the event stream instead using an MVar. After adding the watch we can inject an event, say `isWatchReady` event, this event will also indicate the path for which the watch is ready, each path being watched would generate this event before any other events.

However to have a race-free fool-proof sync, we will need to do the following:

* when adding a watch we would need to take an MVar, add the watch, add a "ready event" to a ready-event list, put the MVar.
* Before we read the watch handle we need to take the MVar, first read any events queued to the "ready" event list, then read events from the watch handle, then put the MVar.

If this takes too much time or makes it too complex we can just drop this. Remember, that we will have to do this for all three platforms. So effort is tripled if it is different for each one.

harendra-kumar:
The use case for this may be to reliably record all events without missing anything. To be able to do that we have to establish a watch before we can start fs operations.

harendra-kumar:
Closing this for now, may not be worth the effort unless someone really needs it. Tests can do with a timeout for now.

PR #862, opened 2021-01-11
Add support for FS Events  readiness

rnjtranjan:


rnjtranjan:
Fixes #767

harendra-kumar (changes requested):
This is not what I had in mind. 

We need to generate an additional event in the event stream instead using an MVar. After adding the watch we can inject an event, say `isWatchReady` event, this event will also indicate the path for which the watch is ready, each path being watched would generate this event before any other events.

However to have a race-free fool-proof sync, we will need to do the following:

* when adding a watch we would need to take an MVar, add the watch, add a "ready event" to a ready-event list, put the MVar.
* Before we read the watch handle we need to take the MVar, first read any events queued to the "ready" event list, then read events from the watch handle, then put the MVar.

If this takes too much time or makes it too complex we can just drop this. Remember, that we will have to do this for all three platforms. So effort is tripled if it is different for each one.

PR #866, opened 2021-01-13
Make the initial value of Fold a Step type

adithyaov:


harendra-kumar (approved):
I am running the benchmarks once more.

harendra-kumar:
Benchmarks look good, let's merge it.

Issue #864, opened 2021-01-11
fromListM not able to async execution is kinda weird


eddiemundo:
I just noticed that `fromListM` doesn't have the `concurrent` tag underneath it in the documentation so I guess this is known but it's a bit of a UX issue that `fromListM` is different from `fromFoldableM` besides possible performance in the serial case.
```haskell
getResponse :: Int -> IO ()
getResponse key = do
  print ("Begin " <> show key)
  threadDelay 1000000
  print ("End " <> show key)

main :: IO ()
main = do
  [1..5] <&> getResponse & Streamly.fromListM & Streamly.aheadly & Streamly.drain
```
The above gives
```
"Begin 1"
"End 1"
"Begin 2"
"End 2"
"Begin 3"
"End 3"
"Begin 4"
"End 4"
"Begin 5"
"End 5"
```
If I change `fromListM` to `fromFoldableM` it gives
```
"Begin 1"
"Begin 2"
"Begin 3"
"Begin 4"
"Begin 5"
"End 5"
"End 4"
"End 3"
"End 2"
"End 1"
```
which is more what I expect, but I wouldn't known the difference looking at the code alone.

harendra-kumar:
You are right and your expectation is correct. It can in fact be made concurrent easily, we just did not notice it and nobody complained. We will change it in the upcoming release. We just need to change the default definition of fromListM to fromFoldable and write a rewrite rule for serial streams to fall back to the faster non-concurrent version.

PR #867, opened 2021-01-13
change fromListM to fromFoldable, add a rewrite rule for serial streams

specdrake:
fix issue #864

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
@@ -506,9 +506,14 @@ fromFoldableM = Prelude.foldr consM K.nil
 -- @since 0.4.0
 {-# INLINE_EARLY fromListM #-}
 fromListM :: (MonadAsync m, IsStream t) => [m a] -> t m a
-fromListM = fromStreamD . D.fromListM
+fromListM = fromFoldableM
 {-# RULES "fromListM fallback to StreamK" [1]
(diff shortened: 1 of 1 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
```diff
@@ -506,10 +506,15 @@ fromFoldableM = Prelude.foldr consM K.nil
 -- @since 0.4.0
 {-# INLINE_EARLY fromListM #-}
 fromListM :: (MonadAsync m, IsStream t) => [m a] -> t m a
-fromListM = fromStreamD . D.fromListM
+fromListM = fromFoldableM
 {-# RULES "fromListM fallback to StreamK" [1]
     forall a. D.toStreamK (D.fromListM a) = fromFoldableM a #-}
 
+{-# RULES "unfoldrM serial" fromListM = fromListMSerial #-}
```
harendra-kumar, line 513: 1. typo - change the rule name.
2. Did you check the serial fromListM performance? It should remain unchanged if the rule is working.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
specdrake: 1 - Done
2 - Yes, the performance is same.

harendra-kumar (approved):

PR #868, opened 2021-01-15
Remove the prefix *l* in a few functions of Fold

adithyaov:
Fold.ltake -> Fold.takeLE
Fold.lfilter -> Fold.filter
Fold.lfilterM -> Fold.filterM
Fold.lmap -> Fold.map
Fold.lchunksOf2 -> Fold.chunksOf2
Fold.lchunksOf -> Fold.chunksOf
Fold.lsessionsOf -> Fold.intervalsOf

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -214,7 +214,7 @@ module Streamly.Internal.Data.Fold
     -- * Nesting
     , many
     , lsessionsOf
-    , lchunksOf
+    , chunksOf
     , chunksBetween
 
     , concatSequence
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -137,7 +137,7 @@ module Streamly.Internal.Data.Fold
     -}
 
     -- ** Trimming
-    , ltake
+    , takeLE
     , takeByTime
     -- By elements
     , sliceSepBy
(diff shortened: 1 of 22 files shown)


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -450,6 +434,20 @@ transform (Pipe pstep1 pstep2 pinitial) (Fold fstep finitial fextract) =
...
+-- >>> f x = if even x then Just x else Nothing
+-- >>> fld = Fold.mapMaybe f Fold.toList
+-- >>> Stream.fold fld (Stream.enumerateFromTo 1 10)
+-- [2,4,6,8]
+--
+-- /Internal/
+{-# INLINE mapMaybe #-}
+mapMaybe :: (Monad m) => (a -> Maybe b) -> Fold m b r -> Fold m a r
+mapMaybe f = map f . filter isJust . map fromJust
+
```
harendra-kumar, line 450: Why was this moved here?

src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -644,15 +644,15 @@ instance (Monad m, Floating b) => Floating (Fold m a b) where
 -- Internal APIs
 ------------------------------------------------------------------------------
 
--- | @(lmap f fold)@ maps the function @f@ on the input of the fold.
+-- | @(map f fold)@ maps the function @f@ on the input of the fold.
```
harendra-kumar, line 647: We should emphatically note that this is not the same as `fmap`. `fmap` is covariant while this contravariant.

src/Streamly/Internal/Data/Fold.hs
adithyaov: This was in the wrong place, to begin with. This was grouped with `rmapM` and `rsequence`.

harendra-kumar (approved):

PR #870, opened 2021-01-15
Deprecate mapM and sequence in favor of rmapM and rsequence

adithyaov:


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -2193,3 +2197,36 @@ toParallelSVarLimited svar winfo = Fold step initial extract
...
+-- Deprecated
+------------------------------------------------------------------------------
+
+-- | Flatten the monadic output of a fold to pure output.
+--
+-- @since 0.7.0
+{-# DEPRECATED sequence "Use rsequence instead" #-}
+{-# INLINE sequence #-}
+sequence :: Monad m => Fold m a (m b) -> Fold m a b
+sequence (Fold step initial extract) = Fold step' initial1 extract'
```
harendra-kumar, line 2211: This should be defined as `sequence = rsequence`.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -2193,3 +2197,36 @@ toParallelSVarLimited svar winfo = Fold step initial extract
...
+
+    extract' = join . extract
+
+-- | Map a monadic function on the output of a fold.
+--
+-- @since 0.7.0
+{-# DEPRECATED mapM "Use rmapM instead" #-}
+{-# INLINE mapM #-}
+mapM :: Monad m => (b -> m c) -> Fold m a b -> Fold m a c
+mapM f = sequence . fmap f
```
harendra-kumar, line 2232: This should be defined as `mapM = rmapM`. Also, we do not need a separate section for deprecated functions. They can be defined next to the main definitions. Similar functions should be together deprecated or not.

harendra-kumar (changes requested):
`rmap` and `lmap` are used profunctors package for covariant and contravariant map. So prefixing `r` for covariant is consistent with that. We can also think of these as on the left side of the `->` and on the right side of it.

Another possible mnemonic is to use `i` for input and `o` for output instead of using `r` and `l`.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -367,9 +369,9 @@ generally = hoist (return . runIdentity)
 -- | Flatten the monadic output of a fold to pure output.
 --
 -- @since 0.7.0
```
harendra-kumar, line 371: Change to 0.8.0.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -384,12 +386,28 @@ sequence (Fold step initial extract) = Fold step' initial1 extract'
...
+--
+-- @since 0.7.0
+{-# DEPRECATED sequence "Use rsequence instead" #-}
+{-# INLINE sequence #-}
+sequence :: Monad m => Fold m a (m b) -> Fold m a b
+sequence = rsequence
+
+-- | Map a monadic function on the output of a fold.
+--
+-- @since 0.7.0
```
harendra-kumar, line 399: Change to 0.8.0.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #874, opened 2021-01-17
Refactor IsStream and rearrange modules in dependency order in cabal file

harendra-kumar:
IsStream now exports entire submodules.


src/Streamly/Internal/Data/Stream/IsStream/Combinators.hs
```diff
@@ -24,6 +21,8 @@ module Streamly.Internal.Data.Stream.Combinators
     )
 where
 
+#include "inline.hs"
```
pranaysashank, line 24: It is better to add any includes at the top of file (though it doesn't matter for this particular file.)

pranaysashank (approved):

src/Streamly/Internal/Data/Stream/IsStream/Combinators.hs
harendra-kumar: In your refactoring you moved the includes next to imports, and that's how I started doing that. Did you do that by mistake? Anyway, I like it next to the imports because these also serve the same purpose as imports. It should be safe to do it here, we may only need them earlier than this if we are using the definitions in exports.

PR #878, opened 2021-01-18
Add support for new line boundary split and re position the file handle  till n…

rnjtranjan:
…ew line

### DIFF
--- a/src/Streamly/Internal/FileSystem/Handle.hs
+++ b/src/Streamly/Internal/FileSystem/Handle.hs
@@ -30,6 +30,7 @@ module Streamly.Internal.FileSystem.Handle
 
     , toChunksWithBufferOf
     , toChunks
+    , toChunksTillNewLine
     , getChunks
 
     -- ** Write to Handle
@@ -103,10 +104,10 @@ import Control.Monad.IO.Class (MonadIO(..))
 import Data.Word (Word8)
 import Foreign.ForeignPtr (withForeignPtr)
 import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
-import Foreign.Ptr (minusPtr, plusPtr)
+import Foreign.Ptr (minusPtr, plusPtr, Ptr)
 import Foreign.Storable (Storable(..))
 import GHC.ForeignPtr (mallocPlainForeignPtrBytes)
-import System.IO (Handle, hGetBufSome, hPutBuf, stdin, stdout)
+import System.IO (Handle, SeekMode(RelativeSeek), hSeek, hGetBufSome, hPutBuf, stdin, stdout)
 import Prelude hiding (read)
 
 import Streamly.Prelude (MonadAsync)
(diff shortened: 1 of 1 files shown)


harendra-kumar (changes requested):
We do not need these APIs.

Please use `splitOn :: (IsStream t, MonadIO m) => Word8 -> t m (Array Word8) -> t m (Array Word8) ` from the ArrayStream module. You can use splitOn on a stream of Array Word8 to get arrays aligned on newlines.

rnjtranjan:
> We do not need these APIs.
> 
> Please use `splitOn :: (IsStream t, MonadIO m) => Word8 -> t m (Array Word8) -> t m (Array Word8) ` from the ArrayStream module. You can use splitOn on a stream of Array Word8 to get arrays aligned on newlines.

I just tried this API, the output has coalesced. 

Output without splitOn API:
("/home/ranjeetr/composewell-private/io-relay/aa3.txt","aaaaaaaaaaaaaaaa")
("/home/ranjeetr/composewell-private/io-relay/aa3.txt","bbbbbbbbbbbbbbbbbbb")
("/home/ranjeetr/composewell-private/io-relay/aa3.txt","cccccccccccDDDDDDDDDDDDDDDDDDDssssssssssssssss")

Output with splitOn API:
("/home/ranjeetr/composewell-private/io-relay/aa3.txt","aaaaaaaaaaaaaaaabbbbbbbbbbbbbbbbbbbcccccccccccDDDDDDDDDDDDDDDDDDDssssssssssssssss")

As per fluent bit documentation: each line of a log file is treated as a separate record that can be parsed by any third party.
If we combined multiple lines in a single record then parsing will be complicated.

harendra-kumar:
If you just want to get lines from a stream that's even simpler. See `splitWithSuffix` in Streamly.Internal.Prelude and `lines` in Streamly.Internal.Unicode.Stream. On a Word8 stream you can simply use `splitOnSuffix (== 10)`.

rnjtranjan:
Kool, the existing API worked as expected:

 Stream.finally go (Handle.toChunks fh)
    & Unicode.decodeUtf8Arrays'           -- Stream IO Char
    & Unicode.lines FL.toList                   -- Stream IO [Char]
    & Stream.map (\a -> (fp, a))              -- map with file path

    where

        go = do
            off <- liftIO $ hTell fh
            -- Stream finished persist handle offset
            liftIO $ putMVar mvar fh
            liftIO $ synchPersistentMap dbdir fp (show off)

rnjtranjan:
Existing API does not solve the following use cases:
1. A file having data not ending with new line (means pending data to be written/saved). In this case incomplete line is fetched.

For example data at time : 

(initial)        t1-> 123456\n789
(next save) t2-> 123456\n789101112\n
						  
Ideally output should have two lines : 
L1 = 123456
L2 = 789101112

But current API  produces three lines:
L1 = 123456
L2 = 789
L3 = 101112

rnjtranjan:
The use case mentioned is not likely to happen.

PR #887, opened 2021-01-23
Array refactor

harendra-kumar:


src/Streamly/Internal/Data/Array/Mut/Foreign/Stream.hs
```diff
@@ -0,0 +1,310 @@
...
+-- Module      : Streamly.Internal.Data.Array.Mut.Foreign.Stream
+-- Copyright   : (c) 2020 Composewell Technologies
+-- License     : BSD3-3-Clause
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+-- Combinators to efficiently manipulate streams of immutable arrays.
+--
+module Streamly.Internal.Data.Array.Mut.Foreign.Stream
```
adithyaov, line 11: `Streamly.Internal.Data.Array.Foreign.Mut`?

harendra-kumar: Something like this:

* Data.Array.Foreign
* Data.Array.Foreign.Stream
* Data.Array.Foreign.Mut
* Data.Array.Foreign.Mut.Stream

adithyaov: Yes, something like that. I think you've mistyped it as `Data.Array.Mut.Foreign.Stream`. It should be `Data.Array.Foreign.Mut.Stream`.
`Mut.Foreign` -> `Foreign.Mut`

harendra-kumar: For boxed arrays we are going to have:

* Data.Array
* Data.Array.Stream

For mutable boxed arrays we just add the Mut suffix to these:

* Data.Array.Mut
* Data.Array.Mut.Stream

If we just add the Foreign suffix to these then we get:

* Data.Array.Foreign
* Data.Array.Stream.Foreign
* Data.Array.Mut.Foreign
* Data.Array.Mut.Stream.Foreign

PR #889, opened 2021-01-24
Add Parser(D).groupByRolling

adithyaov:


src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -624,6 +625,38 @@ groupBy cmp (Fold fstep finitial fextract) = Parser step initial extract
     extract (GroupByInit s) = fextract s
     extract (GroupByGrouping _ s) = fextract s
 
+-- | See 'Streamly.Internal.Data.Parser.groupByRolling'.
+--
+{-# INLINE groupByRolling #-}
+groupByRolling :: Monad m => (a -> a -> Bool) -> Fold m a b -> Parser m a b
+groupByRolling cmp (Fold fstep finitial fextract) = Parser step initial extract
```
harendra-kumar, line 632: Instead of `cmp` we should call it `eq`.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #892, opened 2021-01-26
Benchmark misc

harendra-kumar:
Some small changed related to benchmarking.


hie.yaml
```diff
@@ -12,6 +12,10 @@ cradle:
               component: "bench:Data.Parser.ParserD"
             - path: "./benchmark/Streamly/Benchmark/Data/Parser/ParserK.hs"
               component: "bench:Data.Parser.ParserK"
+            - path: "./benchmark/Streamly/Benchmark/FileSystem/Handle/Read.hs"
```
pranaysashank, line 15: you can just leave the path at the directory, all the haskell files inside them should use the same component

pranaysashank (approved):

PR #893, opened 2021-01-26
Add reverse' test

harendra-kumar:
Add more diagnostic information to debug better when tests fail.

Print the lists being compared in the fromListN test.

Added list length printing to know the length when the lists printed are large.

### DIFF
--- a/test/lib/Streamly/Test/Prelude/Common.hs
+++ b/test/lib/Streamly/Test/Prelude/Common.hs
@@ -133,6 +133,7 @@ import Streamly.Prelude (avgRate, rate, maxBuffer, maxThreads)
 #endif
 import qualified Streamly.Prelude as S
 import qualified Streamly.Data.Fold as FL
+import qualified Streamly.Internal.Data.Stream.IsStream as S
 import qualified Streamly.Internal.Data.Fold as FL
 import qualified Streamly.Internal.Data.Unfold as UF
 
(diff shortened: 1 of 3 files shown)


test/lib/Streamly/Test/Prelude/Common.hs
```diff
@@ -1180,7 +1181,7 @@ transformCombineOpsCommon constr desc eq t = do
 
     -- reordering
     prop (desc <> " reverse") $ transform reverse t S.reverse
-    -- prop (desc <> " reverse'") $ transform reverse t S.reverse'
+    prop (desc <> " reverse'") $ transform reverse t S.reverse'
```
pranaysashank, line 1184: This only works on ordered streams, so I don't think it should be in `transformCombineOpsCommon`

pranaysashank (changes requested):

test/lib/Streamly/Test/Prelude/Common.hs
harendra-kumar: How is it different than `reverse`? They both should be moved?

pranaysashank: Yes both of them should be moved to the Ordered ops I think

harendra-kumar: I have too much stuff to do, I will just add a comment and merge it for now.

PR #895, opened 2021-01-29
Make Fold.head terminating.

pranaysashank:


harendra-kumar (approved):

harendra-kumar:
How about using `Fold1_` and `Maybe` type instead of using Strict Maybe?

PR #901, opened 2021-02-02
Implement readRev to read an array in reverse order

harendra-kumar:
Just added the API but have not used it yet.

### DIFF
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
@@ -77,6 +77,7 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Types
     -- * Unfolds
     , ReadUState
     , read
+    , readRev
     , flattenArrays
     , flattenArraysRev
 
@@ -523,6 +524,33 @@ read = Unfold step inject
             return $ D.Yield x
                 (ReadUState fp (p `plusPtr` sizeOf (undefined :: a)))
 
+-- | Unfold an array into a stream in reverse order.
+--
+-- /Internal/
+{-# INLINE_NORMAL readRev #-}
+readRev :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
+readRev = Unfold step inject
+    where
+
+    inject (Array fp end _) =
+        let p = end `plusPtr` negate (sizeOf (undefined :: a))
+         in return $ ReadUState fp p
+
+    {-# INLINE_LATE step #-}
+    step (ReadUState fp@(ForeignPtr start _) p) | p < Ptr start =
+        let x = unsafeInlineIO $ touchForeignPtr fp
+        in x `seq` return D.Stop
+    step (ReadUState fp p) = do
+            -- unsafeInlineIO allows us to run this in Identity monad for pure
+            -- toList/foldr case which makes them much faster due to not
+            -- accumulating the list and fusing better with the pure consumers.
+            --
+            -- This should be safe as the array contents are guaranteed to be
+            -- evaluated/written to before we peek at them.
+            let !x = unsafeInlineIO $ peek p
+            return $ D.Yield x
+                (ReadUState fp (p `plusPtr` negate (sizeOf (undefined :: a))))
+
 data FlattenState s a =
       OuterLoop s
     | InnerLoop s !(ForeignPtr a) !(Ptr a) !(Ptr a)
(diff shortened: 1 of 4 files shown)


src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
```diff
@@ -523,6 +524,33 @@ read = Unfold step inject
...
+-- | Unfold an array into a stream in reverse order.
+--
+-- /Internal/
+{-# INLINE_NORMAL readRev #-}
+readRev :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
+readRev = Unfold step inject
+    where
+
+    inject (Array fp end _) =
+        let p = end `plusPtr` negate (sizeOf (undefined :: a))
```
adithyaov, line 536: `minusPtr`?

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
```diff
@@ -523,6 +524,33 @@ read = Unfold step inject
...
+         in return $ ReadUState fp p
+
+    {-# INLINE_LATE step #-}
+    step (ReadUState fp@(ForeignPtr start _) p) | p < Ptr start =
+        let x = unsafeInlineIO $ touchForeignPtr fp
+        in x `seq` return D.Stop
```
adithyaov, line 542:
```suggestion
        let !x = unsafeInlineIO $ touchForeignPtr fp
        in return D.Stop
```

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
```diff
@@ -523,6 +524,33 @@ read = Unfold step inject
...
+    step (ReadUState fp p) = do
+            -- unsafeInlineIO allows us to run this in Identity monad for pure
+            -- toList/foldr case which makes them much faster due to not
+            -- accumulating the list and fusing better with the pure consumers.
+            --
+            -- This should be safe as the array contents are guaranteed to be
+            -- evaluated/written to before we peek at them.
+            let !x = unsafeInlineIO $ peek p
+            return $ D.Yield x
+                (ReadUState fp (p `plusPtr` negate (sizeOf (undefined :: a))))
```
adithyaov, line 552:
```suggestion
            let next = ReadUState fp (p `minusPtr` sizeOf (undefined :: a))
            return $ D.Yield x next
```

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
```diff
@@ -523,6 +524,33 @@ read = Unfold step inject
...
+readRev = Unfold step inject
+    where
+
+    inject (Array fp end _) =
+        let p = end `plusPtr` negate (sizeOf (undefined :: a))
+         in return $ ReadUState fp p
+
+    {-# INLINE_LATE step #-}
+    step (ReadUState fp@(ForeignPtr start _) p) | p < Ptr start =
+        let x = unsafeInlineIO $ touchForeignPtr fp
```
adithyaov, line 541: How would this work? The `touchForeignPtr` is no longer sequential. It can be evaluated anytime according to what the compiler decides. We probably shouldn't `unsafeInlineIO` here.

adithyaov: `touchForeignPtr fp >> return D.Stop` should work?

harendra-kumar: 1. There is a `seq` in the next line.
2. touchForeignPtr fp >> return D.Stop would require m to be IO Monad/MonadIO. It cannot work in Identity.

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
harendra-kumar: `minusPtr :: Ptr a -> Ptr b -> Int `

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
harendra-kumar: `minusPtr` is not what you think. `minusPtr :: Ptr a -> Ptr b -> Int `

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
adithyaov: `seq` does not enforce an ordering? `seq a b` says both `a` and `b` will be evaluated before `seq` returns anything.

harendra-kumar: Our intention is not to enforce any ordering here. Our intention is to keep the pointer alive up to this point. It does not matter whether x is evaluated before or after. The only point is to not let the compiler optimize away x thinking that its unused.

adithyaov (approved):

Issue #100, opened 2018-08-27
Add mergeWith and mergeAsyncWith combinators


harendra-kumar:
The way `zipWith` and `zipAsyncWith` zip two streams using a zip function, the same way `mergeWith` and `mergeAsyncWith` would merge two streams using a merge function. The merge operation just orders the elements from two streams i.e. who goes first in the resultant stream.

They should look like:

```
mergeWith :: (IsStream t, Monad m) => (a -> a -> Ordering) -> t m a -> t m a -> t m a
mergeAsyncWith :: (IsStream t, MonadAsync m) => (a -> a -> Ordering) -> t m a -> t m a -> t m a
```

`mergeAsyncWith` is just like `mergeWith` except that it generates the two streams concurrently. These operations have been implemented in [this example](https://github.com/composewell/streamly/blob/master/examples/MergeSort.hs) , but they may not be  very efficient as they use `uncons`. Moreover, it makes sense to have these in the prelude as general functions just like zip.

naushadh:
👍 for the 2-way merge offered by `mergeWith` and `mergeAsyncWith`.

I was wondering if there could be additional combinators for N-way merge?

```haskell
mergeNBy :: (IsStream t, Monad m) => (a -> a -> Ordering) -> [t m a] -> t m a
mergeNAsyncBy :: (IsStream t, Monad m) => (a -> a -> Ordering) -> [t m a] -> t m a
```

harendra-kumar:
How about:

```haskell
mergeNBy cmp = Prelude.foldr (mergeWith cmp) nil
```

naushadh:
Based on [this](https://en.wikipedia.org/wiki/K-way_merge_algorithm#k-way_merge) and some of my own experimentation, 2-way merges are a lot slower than say a 16-way merge.

<details><summary>2 vs 16 way merge</summary>
<p>

```diff
--- .scratch/out/x-2018-12-06T18:14:36-default.parallel:4-1M.txt	2018-12-06 18:15:01.000000000 -0500
+++ .scratch/out/x-2018-12-06T18:14:17-default.parallel:4-1M.txt	2018-12-06 18:14:28.000000000 -0500
@@ -1,56 +1,49 @@
-2.179617s
-1.857034s
-2.331545s
-2.775779s
-2.18057s
-2.520403s
-2.17421s
-2.412253s
-2.171273s
-  57,385,772,832 bytes allocated in the heap
-  11,928,198,640 bytes copied during GC
[...]

PR #908, opened 2021-02-04
Implement concatPairswith/sortBy

harendra-kumar:
sortBy is 10x slower than lists, but it could be concurrent. It would be interesting to see how the StreamD implementation of this performs.

### DIFF
--- a/benchmark/Streamly/Benchmark/Prelude/Serial/Transformation2.hs
+++ b/benchmark/Streamly/Benchmark/Prelude/Serial/Transformation2.hs
@@ -20,6 +20,7 @@ import Control.Monad.IO.Class (MonadIO(..))
 import Data.Monoid (Sum(..))
 import GHC.Generics (Generic)
 
+import qualified Data.List as List
 import qualified Streamly.Internal.Data.Fold as FL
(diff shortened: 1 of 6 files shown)


harendra-kumar:
This is 2-way pairwise merge for #100.

adithyaov (approved):
Ok, I've looked at it long enough and I don't see any problems. Maybe you can add some docs to `concatPairsWith`. Can you also add a test-case?

rnjtranjan (approved):

harendra-kumar:
> Can you also add a test-case?

Wanted to add only skeletons for TODO stuff but ended up implementing some of those, this is one of those. I do not have time right now. Need help on this.

adithyaov:
> Wanted to add only skeletons for TODO stuff but ended up implementing some of those, this is one of those. I do not have time right now. Need help on this.

Can you add minimal documentation then (Maybe an example)? I'll add a test case later.

harendra-kumar:
> Can you add minimal documentation then (Maybe an example)? I'll add a test case later.

There is documentation along with the definition in IsStream module. Also, `sortBy` implementation is an example of using concatPairsWith. I am adding the same in the doc of concatPairsWith.

PR #910, opened 2021-02-05
Add parse and fold functions that return rest of the stream.

pranaysashank:
Naming should be decided.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
@@ -17,6 +17,7 @@ module Streamly.Internal.Data.Stream.StreamD.Eliminate
 
     -- -- * Running a 'Parser'
     , parse
+    , parse_
 
     -- * Stream Deconstruction
     , uncons
@@ -133,12 +134,27 @@ newtype List a = List {getList :: [a]}
 
 -- | Run a 'Parse' over a stream.
 {-# INLINE_NORMAL parse #-}
-parse :: MonadThrow m => PRD.Parser m a b -> Stream m a -> m b
-parse (PRD.Parser pstep initial extract) (Stream step state) = do
+parse
+    :: MonadThrow m
+    => PRD.Parser m a b
+    -> Stream m a
+    -> m b
+parse parser strm = do
+    (b, _) <- parse_ parser strm
+    return b
+
+-- | Run a 'Parse' over a stream and return rest of the Stream.
+{-# INLINE_NORMAL parse_ #-}
+parse_
+    :: MonadThrow m
+    => PRD.Parser m a b
+    -> Stream m a
+    -> m (b, Stream m a)
+parse_ (PRD.Parser pstep initial extract) (Stream step state) = do
     res <- initial
     case res of
         PRD.IPartial s -> go SPEC state (List []) s
-        PRD.IDone b -> return b
+        PRD.IDone b -> return (b, Stream (\_ _ -> return Stop) ())
         PRD.IError err -> throwM $ ParseError err
 
     where
@@ -165,10 +181,12 @@ parse (PRD.Parser pstep initial extract) (Stream step state) = do
                         let (src0, buf1) = splitAt n (x:getList buf)
                             src  = Prelude.reverse src0
                         gobuf SPEC s (List buf1) (List src) pst1
-                    PR.Done _ b -> return b
+                    PR.Done _ b -> return (b, Stream step_rest (Left (s, buf)))
                     PR.Error err -> throwM $ ParseError err
             Skip s -> go SPEC s buf pst
-            Stop   -> extract pst
+            Stop   -> do
+                b <- extract pst
+                return (b, let List buffer = buf in fromList buffer)
 
     gobuf !_ s buf (List []) !pst = go SPEC s buf pst
     gobuf !_ s buf (List (x:xs)) !pst = do
@@ -188,8 +206,19 @@ parse (PRD.Parser pstep initial extract) (Stream step state) = do
                 let (src0, buf1) = splitAt n (x:getList buf)
                     src  = Prelude.reverse src0 ++ xs
                 gobuf SPEC s (List buf1) (List src) pst1
-            PR.Done _ b -> return b
+            PR.Done _ b -> return (b, Stream step_rest (Left (s, buf)))
             PR.Error err -> throwM $ ParseError err
+
+    {-# INLINE_LATE step_rest #-}
+    step_rest _ (Left (st, List [])) = return $ Skip (Right st)
+    step_rest _ (Left (st, List (x:[]))) = return $ Yield x (Right st)
+    step_rest _ (Left (st, List (x:xs))) = return $ Yield x (Left (st, List xs))
+    step_rest gst (Right st) = do
+        r <- step gst st
+        return $ case r of
+            Yield x s -> Yield x (Right s)
+            Skip s -> Skip (Right s)
+            Stop -> Stop
 
 ------------------------------------------------------------------------------
 -- Specialized Folds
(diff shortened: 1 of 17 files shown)


harendra-kumar:
Need a benchmark that consumes 1 element, return the stream and then again consumes 1 element from the returned stream and so on.

src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -289,6 +290,30 @@ foldOnce (Fold fstep begin done) (Stream step state) = do
             Skip s -> go SPEC fs s
             Stop -> done fs
 
+{-# INLINE_NORMAL foldAndReturn #-}
+foldAndReturn :: Monad m => Fold m a b -> Stream m a -> m (b, Stream m a)
```
adithyaov, line 294: If this is as efficient as `fold` (which I believe it will be), you can implement `fold` interms of `foldAndReturn`

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -193,6 +194,70 @@ parselMx' pstep initial extract (Stream step state) = do
             PR.Done _ b -> return b
             PR.Error err -> throwM $ ParseError err
 
+-- | Run a 'Parse' over a stream and return rest of the Stream.
+{-# INLINE_NORMAL parseAndReturnD #-}
+parseAndReturnD
```
adithyaov, line 199: If this is as efficient as `parse`, you should implement `parse` interms of this.

adithyaov: You can probably omit the suffix `D`. This is the `StreamD` module.

src/Streamly/Internal/Data/Stream/StreamD/Type.hs
harendra-kumar: We should also change the name foldOnce to fold.

harendra-kumar (changes requested):
Add a benchmark using the `head` fold repeatedly on the remaining stream. It should have performance similar to uncons.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -193,6 +194,70 @@ parselMx' pstep initial extract (Stream step state) = do
...
+                    src  = Prelude.reverse src0 ++ xs
+                gobuf SPEC s (List []) (List src) pst1
+            PR.Continue 0 pst1 ->
+                gobuf SPEC s (List (x:getList buf)) (List xs) pst1
+            PR.Continue n pst1 -> do
+                assert (n <= length (x:getList buf)) (return ())
+                let (src0, buf1) = splitAt n (x:getList buf)
+                    src  = Prelude.reverse src0 ++ xs
+                gobuf SPEC s (List buf1) (List src) pst1
+            PR.Done _ b -> return (b, Stream step s)
```
harendra-kumar, line 258: Bug: We are losing the buffered data here.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -13,7 +13,7 @@
 module Streamly.Internal.Data.Stream.StreamD.Eliminate
     (
     -- * Running a 'Fold'
-      foldOnce -- XXX rename to "fold"
+      fold -- XXX rename to "fold"
```
harendra-kumar, line 16: Remove the comment.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -140,7 +141,20 @@ parselMx'
     -> (s -> m b)
     -> Stream m a
     -> m b
-parselMx' pstep initial extract (Stream step state) = do
+parselMx' pstep initial extract strm = do
+    (b, _) <- parseD_ pstep initial extract strm
+    return b
```
harendra-kumar, line 146: Move this definition after `parseD_`.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -167,10 +181,12 @@ parselMx' pstep initial extract (Stream step state) = do
...
                             src  = Prelude.reverse src0
                         gobuf SPEC s (List buf1) (List src) pst1
-                    PR.Done _ b -> return b
+                    PR.Done _ b -> return (b, Stream step_rest (Left (s, buf)))
                     PR.Error err -> throwM $ ParseError err
             Skip s -> go SPEC s buf pst
-            Stop   -> extract pst
+            Stop   -> do
+                b <- extract pst
+                return (b, Stream step_from_list buf)
```
harendra-kumar, line 189: We can use `(b, D.fromList buf)` here.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -167,10 +181,12 @@ parselMx' pstep initial extract (Stream step state) = do
                         let (src0, buf1) = splitAt n (x:getList buf)
                             src  = Prelude.reverse src0
                         gobuf SPEC s (List buf1) (List src) pst1
-                    PR.Done _ b -> return b
+                    PR.Done _ b -> return (b, Stream step_rest (Left (s, buf)))
```
harendra-kumar, line 184: We can use ``D.fromList buf `D.cons` Stream step s`` here, it should simplify to the same thing.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
pranaysashank: Need to move append from Nesting to Types module, so haven't done this for now.

test/Streamly/Test/Data/Fold.hs
```diff
@@ -60,10 +61,9 @@ rollingHashFirstN =
 head :: [Int] -> Expectation
 head ls = S.fold FL.head (S.fromList ls) `shouldReturn` headl ls
 
-    where
-
-    headl [] = Nothing
-    headl (x:_) = Just x
+headl :: [a] -> Maybe a
+headl [] = Nothing
+headl (x:_) = Just x
```
adithyaov, line 66: Where is the test for `fold_`?

test/Streamly/Test/Data/Fold.hs
```diff
@@ -431,6 +431,19 @@ many =
...
+    rests <- run $ S.toList rest
+    assert (mbh == headl ls)
+    listEquals (==) rests (taill ls)
+
+    where
+
+    taill :: [a] -> [a]
+    taill [] = []
+    taill (_:xs) = xs
+
```
adithyaov, line 446: I'm assuming this was meant to be in the last commit.

test/Streamly/Test/FileSystem/Handle.hs
```diff
@@ -144,7 +144,7 @@ testWriteWithChunk =
                 hr <- openFile fpathRead ReadMode
                 hw <- openFile fpathWrite ReadWriteMode
                 hSeek hw AbsoluteSeek 0
-                Stream.fold (Handle.writeChunks hw)
+                _ <- Stream.fold (Handle.writeChunks hw)
```
adithyaov, line 147: Indentation below this line

adithyaov (commented):
The last commit can be fixed up to where the test is (first).

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
harendra-kumar: Can you raise an issue for it then?

PR #916, opened 2021-02-07
Remove sliceSepByMax

harendra-kumar:
It can be expressed as:

sliceSepByMax p n = sliceSepBy p . takeLE n

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -145,7 +145,6 @@ module Streamly.Internal.Data.Fold
     -- By elements
     , sliceSepBy
     -- , breakOn
-    , sliceSepByMax
     , sliceEndWith
     -- , breakAfter
     {-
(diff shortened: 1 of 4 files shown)


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1037,6 +1036,13 @@ splitAt n fld = splitWith (,) (takeLE n fld)
 -- >>> splitOn' (== '.') "a..b"
 -- > ["a","","b"]
 --
+-- To terminates a fold even before the separator is encountered if its size
```
adithyaov, line 1039: Typo, `terminates -> terminate`. Also, this example is not really needed because,
1. It is very obvious
2. It is very obvious

adithyaov: Explaining such things via elaborate text is pretty hard too. This statement is not very comfortable to read. I recommend removing this example or rewording this into multiple simple sentences.

adithyaov (changes requested):
1 comment. Everything else looks good.

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: Sometimes you need to state the obvious. Examples act as hints for users to see what possibilities exist. I can try to reword it or remove it. But I wanted to keep it since there was a combinator for this, this is a common task.

adithyaov (approved):

PR #917, opened 2021-02-07
Update docs of splitting combinators

harendra-kumar:
Add a template/signature for splitOnPrefix function.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Nesting.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Nesting.hs
@@ -173,7 +173,7 @@ module Streamly.Internal.Data.Stream.IsStream.Nesting
     -- -- *** Using Element Separators
     , splitOn
     , splitOnSuffix
-    -- , splitOnPrefix
+    , splitOnPrefix
 
     -- , splitBy
     , splitWithSuffix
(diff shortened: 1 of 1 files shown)


harendra-kumar:
This needs some rework, `intersperse '.'` should be `intercalate "."` instead.

src/Streamly/Internal/Data/Stream/IsStream/Nesting.hs
```diff
@@ -903,19 +903,16 @@ iterateMapLeftsWith combine f = iterateMapWith combine (either f (const K.nil))
 --
 -- XXX We need takeGE/takeBetween to implement "some" using "many".
 
--- | Apply a 'Fold' repeatedly on a stream and emit the parsed values in the
--- output stream.
---
--- This is the streaming dual of the 'Streamly.Internal.Data.Fold.many'
--- parse combinator.
+-- | Like 'foldMany1' but appends empty fold output if the fold and stream
+-- termination aligns:
```
adithyaov, line 907: Appends an empty fold if the fold does not consume an element and the stream ends. That is probably what you meant by termination aligns? Maybe something like,

"Always emits the accumulated fold when the stream ends. If the fold does not consume any element, it emits the initial value of fold (empty fold)."

harendra-kumar: It is very hard to explain properly, this combinator should not exist. I am anyway zapping it in the next PR. "Always emits the accumulated fold" is also not clear, isn't it done in the other cominator as well? "If the fold does not consume ...", a question may be asked - why was the fold even started when we know that the stream terminated. You are explaining it like that because you know the internal implementation.

harendra-kumar: Let's merge this and we will postpone this to later if it still remains.

adithyaov:
While you're at itc an you fix the examples in `splitBySeq`?

adithyaov (approved):

PR #925, opened 2021-02-09
Add doctests

pranaysashank:


src/Streamly/Internal/Data/Parser.hs
```diff
@@ -560,17 +573,17 @@ escapedSliceSepBy _cond _esc = undefined
 --
 -- For example,
 --
--- >>> escapedFrameBy (== '{') (== '}') (== '\\') S.toList $ S.fromList "{hello}"
--- > "hello"
+-- > Stream.parse (Parser.escapedFrameBy (== '{') (== '}') (== '\\') Fold.toList) $ Stream.fromList "{hello}"
```
pranaysashank, line 576: Examples that do not work or use unimplemented functions had their markup changed from `>>>` to `>` but this results in  haddock that looks like this
image

It's probably better to surround them in `@` instead.

harendra-kumar (commented):
Looks good.

PR #928, opened 2021-02-09
Add a ghc wrapper to invoke ghc on individual files

harendra-kumar:
With all the extensions used in streamly, optimization options and
ddump-simpl options.

This is useful for compiling standalone files, when inspecting core for
perf issues.  I wish cabal could invoke ghc with environment from the
cabal file. Can it?

### DIFF
--- a/bin/ghc.sh
+++ b/bin/ghc.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+
+ghc -Wall                            \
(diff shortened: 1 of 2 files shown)


adithyaov:
@harendra-kumar Can you fix the CI. Probably need to add this to `extra-src-files`.

adithyaov (approved):

PR #932, opened 2021-02-10
Use fusion plugin by default

harendra-kumar:


bin/bench.sh
```diff
@@ -295,8 +297,8 @@ do
     --fields) shift; FIELDS=$1; shift ;;
     --base) shift; BASE=$1; shift ;;
     --candidate) shift; CANDIDATE=$1; shift ;;
-    --cabal-build-flags) shift; CABAL_BUILD_OPTIONS=$1; shift ;;
-    --cabal-build-options) shift; CABAL_BUILD_OPTIONS=$1; shift ;;
+    --cabal-build-flags) shift; CABAL_BUILD_OPTIONS+=$1; shift ;;
```
adithyaov, line 300: Do you plan to remove this in the future?

adithyaov (approved):

bin/bench.sh
harendra-kumar: Yes.

Issue #913, opened 2021-02-06
Add a CI for GHC 9.0.1


harendra-kumar:

harendra-kumar:
The CI has been added #933, but needs to enabled when github actions is able to install cabal-3.4.

harendra-kumar:
This is done.

PR #933, opened 2021-02-10
Fixes for GHC 9.0 support

harendra-kumar:
* Still needs `--allow-newer`
* For benchmarks needs - `basement` to be fixed (for gauge), fusion-plugin fixes

### DIFF
--- a/src/Streamly/Internal/FileSystem/FD.hs
+++ b/src/Streamly/Internal/FileSystem/FD.hs
@@ -217,7 +217,11 @@ readArrayUpto size (Handle fd) = do
     -- ptr <- mallocPlainForeignPtrAlignedBytes size (alignment (undefined :: Word8))
     withForeignPtr ptr $ \p -> do
         -- n <- hGetBufSome h p size
+#if MIN_VERSION_base(4,15,0)
+        n <- RawIO.read fd p 0 size
+#else
         n <- RawIO.read fd p size
+#endif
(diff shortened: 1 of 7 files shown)


hasufell:
This would be useful to move forward.

src/Streamly/Internal/FileSystem/FD.hs
```diff
@@ -217,7 +217,11 @@ readArrayUpto size (Handle fd) = do
     -- ptr <- mallocPlainForeignPtrAlignedBytes size (alignment (undefined :: Word8))
     withForeignPtr ptr $ \p -> do
         -- n <- hGetBufSome h p size
+#if __GLASGOW_HASKELL__ < 900
```
pranaysashank, line 220: constraint on base ?

harendra-kumar: This is imported from GHC `import qualified GHC.IO.Device as RawIO`

pranaysashank: it is part of base https://hackage.haskell.org/package/base-4.14.1.0/docs/GHC-IO-Device.html#t:RawIO

harendra-kumar: ok, will change.

harendra-kumar:
fixes #913

PR #934, opened 2021-02-10
Add support for --with-compiler

harendra-kumar:
We could use `--with-compiler` in `--cabal-build-options` but we need to find the benchmark/test executable under dist-newstyle based on the compiler version, so we need to know the executable in the shell script.

### DIFF
--- a/bin/build-lib.sh
+++ b/bin/build-lib.sh
@@ -137,8 +137,6 @@ set_common_vars () {
   TARGET_EXE_ARGS=
   RTS_OPTIONS=
 
-  GHC_VERSION=$(ghc --numeric-version)
-
   CABAL_BUILD_OPTIONS=""
   CABAL_EXECUTABLE=cabal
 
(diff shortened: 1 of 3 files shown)


bin/build-lib.sh
```diff
@@ -155,6 +153,18 @@ set_common_vars () {
   fi
 }
 
+# To be called after parsing CLI arguments
+set_derived_vars () {
+  if test -n "$CABAL_WITH_COMPILER"
+  then
+    CABAL_BUILD_OPTIONS+=" --with-compiler $CABAL_WITH_COMPILER"
```
adithyaov, line 160: `$CABAL_WITH_COMPILER` can be a path. Will this work if the path has spaces? Maybe quote this?

adithyaov (commented):
1 comment, rest LGTM.

bin/build-lib.sh
harendra-kumar: We do not usually use paths with spaces, in fact I would not use it with a path. We can fix the issue if and when it arises.

PR #936, opened 2021-02-13
Move parseMany benchamrks from FileSystem.Handle to Data.Parser

adithyaov:


benchmark/Streamly/Benchmark/Data/Parser.hs
```diff
@@ -18,21 +19,30 @@ import Data.Foldable (asum)
 import Data.Functor (($>))
 import Data.Maybe (fromMaybe)
 import Data.Monoid (Sum(..))
+import GHC.Magic (inline)
+#if __GLASGOW_HASKELL__ >= 802
+import GHC.Magic (noinline)
+#else
+#define noinline
```
adithyaov, line 26: This was there in `FileSystem.Read`, what does this do?

adithyaov, line 26: What is the purpose of this?

harendra-kumar: it defines "noinline" to empty value. Wherever "noinline" occurs in the file it will be erased.

harendra-kumar (approved):

benchmark/Streamly/Benchmark/Data/Parser.hs
adithyaov: Ah. Very sneaky.

PR #939, opened 2021-02-16
Improving time-ish stream combinators

adithyaov:
- [x] Improve dropByTime
- [x] Fix takeByTime

`dropByTime` improved from `63.02 ms` to `609.8 μs`

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
@@ -1072,7 +1072,7 @@ dropWhileM p m = fromStreamD $ D.dropWhileM p $ toStreamD m
 -- /Pre-release/
 --
 {-# INLINE dropInterval #-}
-dropInterval ::(MonadIO m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
+dropInterval ::(MonadAsync m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
 dropInterval d = fromStreamD . D.dropByTime d . toStreamD
 
 -- | Drop @n@ elements at the end of the stream.
--- a/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
@@ -1082,8 +1082,9 @@ dropWhileM p m = fromStreamD $ D.dropWhileM p $ toStreamD m
 -- /Pre-release/
 --
 {-# INLINE dropInterval #-}
-dropInterval ::(MonadAsync m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
-dropInterval d = fromStreamD . D.dropByTime d . toStreamD
+dropInterval ::
+       (MonadAsync m, IsStream t, TimeUnit64 d) => Double -> d -> t m a -> t m a
+dropInterval g d = fromStreamD . D.dropByTime g d . toStreamD
 
 -- | Drop @n@ elements at the end of the stream.
 --
--- a/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
@@ -235,6 +235,7 @@ import Streamly.Internal.Data.Stream.IsStream.Common
     , takeWhile
     , interjectSuffix
     , intersperseM
+    , relTimesWith
     )
 import Streamly.Internal.Data.Stream.Prelude (fromStreamS, toStreamS)
 import Streamly.Internal.Data.Stream.Serial (SerialT)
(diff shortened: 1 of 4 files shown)


adithyaov:
With, `Serial.wSerialMin`,
```
benchmarked Prelude.Serial/o-1-space/filtering/takeByTime-all
time                 59.33 ms   (58.43 ms .. 59.78 ms)
                     1.000 R²   (0.999 R² .. 1.000 R²)
mean                 58.92 ms   (58.60 ms .. 59.18 ms)
std dev              502.1 μs   (395.5 μs .. 630.4 μs)
```

With, `Par.parallelMin`,
```
benchmarked Prelude.Serial/o-1-space/filtering/takeByTime-all
time                 99.94 ms   (74.56 ms .. 116.0 ms)
                     0.929 R²   (0.799 R² .. 0.987 R²)
mean                 117.3 ms   (107.0 ms .. 133.7 ms)
std dev              23.09 ms   (15.32 ms .. 30.99 ms)
variance introduced by outliers: 68% (severely inflated)
```

adithyaov:
On `origin/master`, `takeByTime` (incorrect implementation):
```
benchmarked Prelude.Serial/o-1-space/filtering/takeByTime-all
time                 62.74 ms   (62.59 ms .. 62.91 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 62.73 ms   (62.67 ms .. 62.81 ms)
std dev              119.1 μs   (86.63 μs .. 163.1 μs)
```

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -991,8 +993,16 @@ takeWhileAround = undefined -- fromStreamD $ D.takeWhileAround n $ toStreamD m
 -- /Internal/
 --
 {-# INLINE takeByTime #-}
-takeByTime ::(MonadIO m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
-takeByTime d = fromStreamD . D.takeByTime d . toStreamD
+takeByTime ::
+       (MonadAsync m, IsStream t, TimeUnit64 d, Functor (t m))
+    => d -> t m a -> t m a
+takeByTime duration = catMaybes . Par.parallelMin timeStream . fmap Just
```
harendra-kumar, line 999: we can flip the arguments and use `parallelFst` since we know one of the streams is infinite, parallelFst may be a bit more efficient as it only checks one stream, it may be negligible though.

harendra-kumar: Behavior change: The earlier implementation would terminate immediately when the time is over. This one would depend on the granularity of the relTimes clock. So we need to let the user specify the granularity as well to control it.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -1024,7 +1034,7 @@ dropWhileM p m = fromStreamD $ D.dropWhileM p $ toStreamD m
 -- /Internal/
 --
 {-# INLINE dropByTime #-}
-dropByTime ::(MonadIO m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
+dropByTime ::(MonadAsync m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
```
harendra-kumar, line 1037: Behavior Change: same as in takeByTime.

We got the perf improvement due to the behavior change. Check the performance at lower timer granularity.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -1072,8 +1082,9 @@ dropWhileM p m = fromStreamD $ D.dropWhileM p $ toStreamD m
 -- /Pre-release/
 --
 {-# INLINE dropInterval #-}
-dropInterval ::(MonadIO m, IsStream t, TimeUnit64 d) => d -> t m a -> t m a
-dropInterval d = fromStreamD . D.dropByTime d . toStreamD
+dropInterval ::
+       (MonadAsync m, IsStream t, TimeUnit64 d) => Double -> d -> t m a -> t m a
+dropInterval g d = fromStreamD . D.dropInterval g d . toStreamD
```
harendra-kumar, line 1087: Need to fix the documentation.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -1039,8 +1041,16 @@ takeWhileAround = undefined -- fromStreamD $ D.takeWhileAround n $ toStreamD m
...
+takeInterval ::
+       (MonadAsync m, IsStream t, TimeUnit64 d, Functor (t m))
+    => Double -> d -> t m a -> t m a
+takeInterval g duration = catMaybes . Par.parallelMin timeStream . fmap Just
+
+    where
+
+    duration64 = toRelTime64 duration
+    timeStream =
+        fmap (const Nothing) $ takeWhile (< duration64) (relTimesWith g)
```
harendra-kumar, line 1053: Can't we implement this in the same way as `dropInterval`? that is by using zipWith?

harendra-kumar (changes requested):

harendra-kumar:
These are already fixed in the current code.

PR #941, opened 2021-02-16
Bug fix: Serial.wSerialMin

adithyaov:


harendra-kumar (approved):
Add a test if possible.

src/Streamly/Internal/Data/Stream/Serial.hs
```diff
@@ -421,10 +421,10 @@ wSerialFst m1 m2 = mkStream $ \st yld sng stp -> do
 -- @since 0.7.0
 {-# INLINE wSerialMin #-}
 wSerialMin :: IsStream t => t m a -> t m a -> t m a
-wSerialMin m1 m2 = mkStream $ \st yld sng stp -> do
+wSerialMin m1 m2 = mkStream $ \st yld _ stp -> do
     let stop       = stp
-        single a   = sng a
-        yieldk a r = yld a (wSerial m2 r)
+        single a   = yld a K.nil
```
harendra-kumar, line 426: It will be nice to have a comment here saying why its deliberately not implemented as `sng a`.

harendra-kumar (approved):

src/Streamly/Internal/Data/Stream/Serial.hs
adithyaov: This was incorrect. Corrected it now.

harendra-kumar (approved):

PR #942, opened 2021-02-16
Efficient parsing using generalized unfolds

harendra-kumar:
Instead of using a closed loop we can stop an unfold and then resume it
later by using an explicit loop state. This allows us to break the loop and restart it at some other
point. For example, we can parse a block from an input stream from a
file handle or a socket and then return the socket/handle plus nay
buffered data (due to backtracking) so that we can resume reading from
it later on after doing some processing.

### DIFF
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
@@ -78,6 +78,7 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Types
     , ReadUState
     , read
     , readRev
+    , readResumable
     , flattenArrays
     , flattenArraysRev
 
@@ -146,6 +147,7 @@ import qualified GHC.Exts as Exts
 import qualified Streamly.Internal.Data.Fold.Types as FL
 import qualified Streamly.Internal.Data.Stream.StreamD.Type as D
 import qualified Streamly.Internal.Data.Stream.StreamK.Type as K
+import qualified Streamly.Internal.Data.Unfold.Resume as UnfoldR
 import qualified Streamly.Internal.Foreign.Malloc as Malloc
 
 import Prelude hiding (length, foldr, read, unlines, splitAt)
@@ -498,12 +500,12 @@ data ReadUState a = ReadUState
     {-# UNPACK #-} !(ForeignPtr a)  -- foreign ptr with end of array pointer
     {-# UNPACK #-} !(Ptr a)         -- current pointer
 
--- | Unfold an array into a stream.
+-- | Resumable unfold of an array.
 --
--- @since 0.7.0
-{-# INLINE_NORMAL read #-}
-read :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
-read = Unfold step inject
+{-# INLINE_NORMAL readResumable #-}
+readResumable :: forall m a. (Monad m, Storable a) =>
+    UnfoldR.Unfold m (Array a) a
+readResumable = UnfoldR.Unfold step inject extract
     where
 
     inject (Array (ForeignPtr start contents) (Ptr end) _) =
@@ -523,6 +525,16 @@ read = Unfold step inject
             let !x = unsafeInlineIO $ peek p
             return $ D.Yield x
                 (ReadUState fp (p `plusPtr` sizeOf (undefined :: a)))
+
+    extract (ReadUState (ForeignPtr end contents) (Ptr p)) =
+        return $ Just $ Array (ForeignPtr p contents) (Ptr end) (Ptr end)
+
+-- | Unfold an array into a stream.
+--
+-- @since 0.7.0
+{-# INLINE_NORMAL read #-}
+read :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
+read = UnfoldR.simplify readResumable
 
 -- | Unfold an array into a stream in reverse order.
 --
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
@@ -78,7 +78,7 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Types
     , ReadUState
     , read
     , readRev
-    , readResumable
+    , producer
     , flattenArrays
     , flattenArraysRev
 
@@ -136,8 +136,9 @@ import GHC.ForeignPtr (ForeignPtr(..))
 import GHC.IO (IO(IO), unsafePerformIO)
 import GHC.Ptr (Ptr(..))
 import Streamly.Internal.Data.Fold.Types (Fold(..))
-import Streamly.Internal.Data.Unfold.Types (Unfold(..))
+import Streamly.Internal.Data.Producer.Type (Producer (..))
 import Streamly.Internal.Data.SVar (adaptState)
+import Streamly.Internal.Data.Unfold.Types (Unfold(..))
 import Text.Read (readPrec, readListPrec, readListPrecDefault)
 
 #ifdef DEVBUILD
@@ -145,9 +146,9 @@ import qualified Data.Foldable as F
 #endif
 import qualified GHC.Exts as Exts
 import qualified Streamly.Internal.Data.Fold.Types as FL
+import qualified Streamly.Internal.Data.Producer as Producer
 import qualified Streamly.Internal.Data.Stream.StreamD.Type as D
 import qualified Streamly.Internal.Data.Stream.StreamK.Type as K
-import qualified Streamly.Internal.Data.Unfold.Resume as UnfoldR
 import qualified Streamly.Internal.Foreign.Malloc as Malloc
 
 import Prelude hiding (length, foldr, read, unlines, splitAt)
@@ -502,10 +503,9 @@ data ReadUState a = ReadUState
 
 -- | Resumable unfold of an array.
 --
-{-# INLINE_NORMAL readResumable #-}
-readResumable :: forall m a. (Monad m, Storable a) =>
-    UnfoldR.Unfold m (Array a) a
-readResumable = UnfoldR.Unfold step inject extract
+{-# INLINE_NORMAL producer #-}
+producer :: forall m a. (Monad m, Storable a) => Producer m (Array a) a
+producer = Producer step inject extract
     where
 
     inject (Array (ForeignPtr start contents) (Ptr end) _) =
@@ -534,7 +534,7 @@ readResumable = UnfoldR.Unfold step inject extract
 -- @since 0.7.0
 {-# INLINE_NORMAL read #-}
 read :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
-read = UnfoldR.simplify readResumable
+read = Producer.simplify producer
 
 -- | Unfold an array into a stream in reverse order.
 --
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
@@ -527,7 +527,7 @@ readResumable = UnfoldR.Unfold step inject extract
                 (ReadUState fp (p `plusPtr` sizeOf (undefined :: a)))
 
     extract (ReadUState (ForeignPtr end contents) (Ptr p)) =
-        return $ Just $ Array (ForeignPtr p contents) (Ptr end) (Ptr end)
+        return $ Array (ForeignPtr p contents) (Ptr end) (Ptr end)
 
 -- | Unfold an array into a stream.
 --
--- a/src/Streamly/Internal/Data/Producer.hs
+++ b/src/Streamly/Internal/Data/Producer.hs
@@ -1,31 +1,29 @@
 -- |
--- Module      : Streamly.Internal.Data.Unfold.Resume
+-- Module      : Streamly.Internal.Data.Producer
 -- Copyright   : (c) 2021 Composewell Technologies
 -- License     : BSD-3-Clause
 -- Maintainer  : streamly@composewell.com
 -- Stability   : experimental
 -- Portability : GHC
 --
--- Unfolds in "Streamly.Internal.Data.Unfold" module run to completion, they
--- cannot be stopped and then resumed. In contrast, the unfolds in this module
--- allow extraction of the intermediate state which can be used to resume the
--- unfold later on.
+-- A 'Producer' is an 'Unfold' with an 'extract' function added to to extract
+-- the state. It is more powerful but less general than an Unfold.
 --
--- For simple cases the seed itself can be used to represent the intermediate
--- state as well. For example, if we are iterating through a list or an array
--- we can return the remaining list or the remaining array slice as the
--- state and resume from it later.
+-- A 'Producer' represents steps of a loop generating a sequence of elements.
+-- While unfolds are closed representation of imperative loops with some opaque
+-- internal state, producers are open loops with the state being accesible to
+-- the user.
 --
--- In general, the intermediate state could be different than the seed. Also,
--- when unfolds are nested, for example using 'concat', then each nesting layer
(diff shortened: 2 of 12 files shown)


harendra-kumar:
Allocations and bytesCopied are 0.

```
Data.Parser(cpuTime)
Benchmark                                                               default(μs)
----------------------------------------------------------------------- -----------
Data.Parser/o-1-space/parseMany/Unfold/1000 arrays/take 1                    356.82
Data.Parser/o-1-space/parseMany/Unfold/1000 arrays/take all                  122.11
Data.Parser/o-1-space/parseMany (take all)                                    66.04
Data.Parser/o-1-space/parseMany (take 1)                                      48.27
```

The type can be further simplified, we can remove the `inject` and `extract`. It should simply be a Step runner function.

src/Streamly/Internal/Data/Unfold/Resume/Type.hs
```diff
@@ -0,0 +1,168 @@
...
+data Unfold m a b =
+    -- | @Unfold step inject extract@
+    forall s. Unfold (s -> m (Step s b)) (a -> m s) (s -> m (Maybe a))
+
+------------------------------------------------------------------------------
+-- Unfolds
+------------------------------------------------------------------------------
+
+-- XXX we do not need a Maybe, we can just use undefined or error. There is no
+-- state to extract in this case, and it should not be used.
```
adithyaov: Is this is specific to `nilM`?

> we do not need a Maybe, we can just use undefined or error

What does this mean? We have a `Maybe` so might as well use it?

src/Streamly/Internal/Data/Unfold/Resume/Type.hs
```diff
@@ -0,0 +1,168 @@
...
+--
+-- However, in certain cases we may not be able to extract, e.g. if we are
+-- unfolding a Map, we may not be able to resume reading a Map. In such cases
+-- it may be useful to return a Nothing. Or maybe we do not allow unfolding
+-- such structures?
```
adithyaov: Is this all specific to `nilM`?

src/Streamly/Internal/Data/Unfold/Resume/Type.hs
```diff
@@ -0,0 +1,168 @@
...
+-- Type
+------------------------------------------------------------------------------
+
+-- | An @Unfold m a b@ is a generator of a stream of values of type @b@ from a
+-- seed of type 'a' in 'Monad' @m@.
+--
+-- /Internal/
+
+data Unfold m a b =
+    -- | @Unfold step inject extract@
```
adithyaov, line 60: I like `eject`. `inject` vs `eject`. `extract` in a way works too I guess.

src/Streamly/Internal/Data/Unfold/Source.hs
```diff
@@ -0,0 +1,88 @@
...
+source = Source []
+
+
+-- | Return some unused data back to the source. The data is prepended (or
+-- consed) to the source.
+--
+-- /Internal/
+unread :: [b] -> Source a b -> Source a b
+unread xs (Source ys seed) = Source (xs ++ ys) seed
+isEmpty :: Source a b -> Bool
```
adithyaov, line 54: newline and some documentation before `isEmpty`

src/Streamly/Internal/Data/Unfold/Resume.hs
```diff
@@ -0,0 +1,231 @@
...
+                        let src0 = Prelude.take n (x:getList buf)
+                            src  = Prelude.reverse src0
+                        s1 <- uextract s
+                        return (b, Source.unread src s1)
+                    Error err -> throwM $ ParseError err
+            Skip s -> go SPEC s buf pst
+            Stop   -> do
+                b <- extract pst
+                -- XXX we should return the remaining buffer
+                return (b, Source.source Nothing)
```
adithyaov, line 164: Why don't you?

adithyaov: Also, I'm only looking at `Done` in `parseD`. I'm assuming everything else is the same.

adithyaov (commented):
I've taken a look. I think I can use something like this in `streamly-lz4`.

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+-- stopped in the middle, its state can be extracted, examined, changed, and
+-- then it can be resumed later from the stopped state.
+--
+-- A producer can be used in places where a CPS stream would otherwise be
+-- needed, because the state of the loop can be passed around. However, it can
+-- be much more efficient than CPS because it allows stream fusion and
+-- unecessary function calls can be avoided.
+
+module Streamly.Internal.Data.Producer
+    ( Unfold (..)
```
adithyaov, line 27:
```suggestion
    ( Producer (..)
```

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -0,0 +1,163 @@
...
+    step x = f x >> return Stop
+
+{-# INLINE nil #-}
+nil :: Monad m => Producer m a b
+nil = nilM (\_ -> return ())
+
+{-# INLINE unfoldrM #-}
+unfoldrM :: Monad m => (a -> m (Maybe (b, a))) -> Producer m a b
+unfoldrM next = Producer step return return
+  where
```
adithyaov, line 72: Indentation

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -0,0 +1,163 @@
...
+    step st = do
+        r <- next st
+        return $ case r of
+            Just (x, s) -> Yield x s
+            Nothing -> Stop
+
+-- | Convert a list of pure values to a 'Stream'
+--
+-- /Internal/
+{-# INLINE_LATE fromList #-}
```
adithyaov, line 83: Is this the right INLINE phase?

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -0,0 +1,163 @@
...
+translate :: Functor m =>
+    (a -> c) -> (c -> a) -> Producer m c b -> Producer m a b
+translate f g (Producer step inject extract) =
+    Producer step (inject . f) (fmap g . extract)
+
+-- | Map the producer input to another value of the same type.
+--
+-- /Internal/
+{-# INLINE_NORMAL lmap #-}
+lmap :: (a -> a) -> Producer m a b -> Producer m a b
```
adithyaov, line 110: We can probably remove this or rename this. I don't think this justifies `lmap`.

harendra-kumar: A search for `f :: a -> a -> g a -> g a` on hoogle gave only fmap/liftA/liftM. Its a map even though it is restricted to an endomorphism. Any alternative names? If we remove it can we achieve the same functionality in some other way?

src/Streamly/Internal/Data/Producer/Source.hs
```diff
@@ -0,0 +1,91 @@
...
+#include "inline.hs"
+
+import Streamly.Internal.Data.Stream.StreamD.Step (Step(..))
+import Streamly.Internal.Data.Producer.Type (Producer(..))
+import Prelude hiding (read)
+
+-- XXX instead of a seed should we wrap a Producer in a source? Then we won't
+-- have to supply a Producer for read.
+--
+-- | An unfold seed with some buffered data. It allows us to 'unread' or return
```
adithyaov, line 38: unfold seed? producer seed?

src/Streamly/Internal/Data/Producer/Source.hs
```diff
@@ -0,0 +1,91 @@
...
+--
+-- /Internal/
+unread :: [b] -> Source a b -> Source a b
+unread xs (Source ys seed) = Source (xs ++ ys) seed
+
+isEmpty :: Source a b -> Bool
+isEmpty (Source [] Nothing) = True
+isEmpty _ = False
+
+-- | An unfold to read from a 'Source'. If the buffer has data then it is read
```
adithyaov, line 61:
```suggestion
-- | An producer to read from a 'Source'. If the buffer has data then it is read
```

src/Streamly/Internal/Data/Producer/Source.hs
```diff
@@ -0,0 +1,91 @@
...
+-- first and then the seed is unfolded using the supplied unfold.
```
adithyaov, line 62:
```suggestion
-- first and then the seed is unfolded using the supplied producer.
```

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
+-- |
+-- Module      : Streamly.Internal.Data.Producer
+-- Copyright   : (c) 2021 Composewell Technologies
+-- License     : BSD-3-Clause
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+-- A 'Producer' is an 'Unfold' with an 'extract' function added to to extract
+-- the state. It is more powerful but less general than an Unfold.
```
adithyaov, line 10: Ah, so we agree that `Producer` is more powerful.

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+--
+-- A 'Producer' represents steps of a loop generating a sequence of elements.
+-- While unfolds are closed representation of imperative loops with some opaque
+-- internal state, producers are open loops with the state being accesible to
```
adithyaov, line 14:
```suggestion
-- internal state, producers are open loops with the state being accessible to
```

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: Yes.

src/Streamly/Internal/Data/Producer/Type.hs
harendra-kumar: This is copied from Unfold, there may have been a reason for this.

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+import Streamly.Internal.Data.Unfold.Types (Unfold(..))
+
+import qualified Streamly.Internal.Data.Parser.ParserD as ParserD
+import qualified Streamly.Internal.Data.Parser.ParserK.Types as ParserK
+import qualified Streamly.Internal.Data.Producer.Source as Source
+
+import Streamly.Internal.Data.Producer.Type
+import Prelude hiding (concat)
+
+-- XXX We should write unfolds as producers where possible and define
```
adithyaov, line 68: I don't think we need to define unfolds using `simplify`. The users can do that themselves if ever need (They should ideally use producers where they can)

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+        return $ case r of
+            Yield x s -> Yield x (Stream step1 s)
+            Skip s    -> Skip (Stream step1 s)
+            Stop      -> Stop
+
+-------------------------------------------------------------------------------
+-- Parsing
+-------------------------------------------------------------------------------
+
+-- XXX This should probably be moved to the Source module?
```
adithyaov, line 107: Probably can. It's a low level module anyway. This type is used in a few other places also.

harendra-kumar: `Source` is going through some overhaul in the next commit/PR.

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+                        assert (n <= length (x:getList buf)) (return ())
+                        let src0 = Prelude.take n (x:getList buf)
+                            src  = Prelude.reverse src0
+                        s1 <- uextract s
+                        return (b, Source.unread src s1)
+                    Error err -> throwM $ ParseError err
+            Skip s -> go SPEC s buf pst
+            Stop   -> do
+                b <- extract pst
+                -- XXX we should return the remaining buffer
```
adithyaov, line 163: Why aren't you returning the remaining buffer here?
`Source buf Nothing`?

src/Streamly/Internal/Data/Producer.hs
```diff
@@ -0,0 +1,235 @@
...
+
+    {-# INLINE_LATE step #-}
+    step src = do
+        if Source.isEmpty src
+        then return Stop
+        else do
+            (b, s1) <- parseD parser reader src
+            return $ Yield b s1
+
+-- | Apply a parser repeatedly on an unfold to generate an unfold of parsed
```
adithyaov, line 226: rename unfold to producer?

adithyaov, line 226:
```suggestion
-- | Apply a parser repeatedly on a producer to generate a producer of parsed
```

src/Streamly/Internal/Data/Array/Foreign/Mut/Types.hs
```diff
@@ -524,6 +526,16 @@ read = Unfold step inject
             return $ D.Yield x
                 (ReadUState fp (p `plusPtr` sizeOf (undefined :: a)))
 
+    extract (ReadUState (ForeignPtr end contents) (Ptr p)) =
+        return $ Array (ForeignPtr p contents) (Ptr end) (Ptr end)
```
adithyaov, line 530: The type of `extact :: s -> m a` puts a constraint on a few things. I might want to return the entire array along with the index (ptr) till where we generated. No use-case but...

I'm sure we'll encounter a use case that forces us to change the type of `extract` to `s -> m c`.

adithyaov (changes requested):
Please address the comments.
I haven't looked at the tests too throughly.

src/Streamly/Internal/Data/Unfold/Resume/Type.hs
harendra-kumar: keeping extract for now.

src/Streamly/Internal/Data/Unfold/Source.hs
harendra-kumar: done.

src/Streamly/Internal/Data/Unfold/Resume.hs
harendra-kumar: It was before we had a Maybe in the source, now we can return the buffer safely I guess. Done.

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: done.

src/Streamly/Internal/Data/Producer/Type.hs
harendra-kumar: Keeping it as it is.

src/Streamly/Internal/Data/Producer/Source.hs
harendra-kumar: Chnaged all such occurrences in Source.hs.

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: Done.

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: Moved.

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: fixed.

src/Streamly/Internal/Data/Producer/Type.hs
harendra-kumar: done.

src/Streamly/Internal/Data/Producer.hs
harendra-kumar: done.

adithyaov (approved):

PR #943, opened 2021-02-17
Add Category and Arrow instances for unfolds

harendra-kumar:
Just moved the functions around in the lower level module, to add Category and Arrow instances. All the functionality was already there, just used the existing functions in the instances.


src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -8,6 +8,13 @@
 -- Stability   : experimental
 -- Portability : GHC
 --
+-- The 'Unfold' abstraction is less powerful but more flexible than the Stream
+-- abstraction.  Stream is an unfold with an embedded state. Therefore, Stream
+-- is like an object whereas unfold is like a function.  Unfold is a 'Category'
+-- and an 'Arrow' but not a 'Monad'. Whereas Stream is not a Category or Arrow
```
adithyaov, line 14: Unfold is a Monad now. Do you want to change the wording?

harendra-kumar: Doing that right now.

adithyaov (commented):
Did you check if the laws for the instances hold?

harendra-kumar:
> Did you check if the laws for the instances hold?

Did not do a rigorous proof, but I thought about it quickly and it looked ok.

adithyaov (approved):

PR #947, opened 2021-02-17
Add Applicative instance for Unfold

harendra-kumar:


harendra-kumar:
```
Data.Unfold(cpuTime)
Benchmark                                                    default(μs)
------------------------------------------------------------ -----------
Data.Unfold/o-1-space/outer-product/toNullAp                       47.87
Data.Unfold/o-1-space/outer-product/toNull                         42.40
```

harendra-kumar:
There was a bug in `Data.Unfold.const` which has been fixed in this PR.

src/Streamly/Internal/Data/Unfold/Types.hs
```diff
@@ -63,3 +67,86 @@ map f (Unfold ustep uinject) = Unfold step uinject
...
+{-# INLINE const #-}
+const :: Applicative m => m b -> Unfold m a b
+const m = Unfold step inject
+
+    where
+
+    inject _ = pure False
+
+    step False = (`Yield` True) <$> m
+    step True = pure Stop
```
adithyaov, line 84: This does not feel right to me. Discarding the `inject` value should be signified in the type.

```
const :: Applicative m => m b -> Unfold m Void b
```

pranaysashank: `const` has to ignore it, you get that from the regular `const :: a -> b -> a`

harendra-kumar: We want to accept an input and discard it. I am not sure how that would be possible with `Void` input type.

src/Streamly/Internal/Data/Unfold/Types.hs
```diff
@@ -63,3 +67,86 @@ map f (Unfold ustep uinject) = Unfold step uinject
...
+        s1 <- inject1 x
+        return $ CrossOuter s1 x
+
+    {-# INLINE_LATE step #-}
+    step (CrossOuter st1 sy) = do
+        r <- step1 st1
+        case r of
+            Yield x s -> do
+                s2 <- inject2 sy
+                return $ Skip (CrossInner s sy s2 x)
```
adithyaov, line 125: You can probably inline `CrossInner`. Saves 1 indirection I guess.

src/Streamly/Internal/Data/Unfold/Types.hs
```diff
@@ -63,3 +67,86 @@ map f (Unfold ustep uinject) = Unfold step uinject
...
+-- | Create a cross product (vector product or cartesian product) of the
+-- output streams of two unfolds.
+--
+{-# INLINE_NORMAL cross #-}
+cross :: Monad m => Unfold m a b -> Unfold m a c -> Unfold m a (b, c)
+cross (Unfold step1 inject1) (Unfold step2 inject2) = Unfold step inject
+
+    where
+
+    inject x = do
```
adithyaov, line 115: Can you rename `x` and `sy` to `seed`?

src/Streamly/Internal/Data/Unfold/Types.hs
```diff
@@ -63,3 +67,86 @@ map f (Unfold ustep uinject) = Unfold step uinject
...
+apDiscardSnd (Unfold _step1 _inject1) (Unfold _step2 _inject2) = undefined
+
+data Cross s1 s2 sy x y = CrossOuter s1 y | CrossInner s1 sy s2 x
+
+-- | Create a cross product (vector product or cartesian product) of the
+-- output streams of two unfolds.
+--
+{-# INLINE_NORMAL cross #-}
+cross :: Monad m => Unfold m a b -> Unfold m a c -> Unfold m a (b, c)
+cross (Unfold step1 inject1) (Unfold step2 inject2) = Unfold step inject
```
adithyaov, line 111: You can maybe rename this to `crossWith` like how we have `teeWith`.

src/Streamly/Internal/Data/Unfold/Types.hs
harendra-kumar: `CrossInner` is a loop, its not a one shot thing.

adithyaov (commented):
Can you fix the CI. Can you also add a benchmark for `cross` or open an issue to do so?

src/Streamly/Internal/Data/Unfold/Types.hs
harendra-kumar: This does not accept a function so its more like `tee` than `teeWith`.

harendra-kumar:
> Can you fix the CI. Can you also add a benchmark for `cross` or open an issue to do so?

The `toNullAp` benchmark measures the performance of `cross`.

src/Streamly/Internal/Data/Unfold/Types.hs
harendra-kumar: ok.

src/Streamly/Internal/Data/Unfold/Types.hs
adithyaov: Yes, what I meant was to make it accept a function and name it `crossWith`

src/Streamly/Internal/Data/Unfold/Types.hs
```diff
@@ -63,3 +67,86 @@ map f (Unfold ustep uinject) = Unfold step uinject
...
+
+    where
+
+    inject a = do
```
adithyaov, line 115: I like `seed` here as it is shared and is the same in the entire scope of the function.

adithyaov (approved):

harendra-kumar:
merged offline, commit 3947fee6

PR #843, opened 2020-12-18
Implement Data.Parser.ParserD.deintercalate

specdrake:

PR #950, opened 2021-02-18
Add a hack to find "chart" executable

harendra-kumar:
We can always use the chart executable built by any compiler version.
Sometimes the chart executable is not buildable by newer compiler
versions, in that case we can build it using another compiler and use
that.


specdrake:
I just tried this by placing ``chart`` executable in ``bin/chart`` but it doesn't resolve the error mentioned here https://github.com/composewell/streamly/pull/843#issuecomment-780286181

PR #953, opened 2021-02-21
Export Source.parseD and add Handle.produceChunksWithBufferOf

adithyaov:


src/Streamly/Internal/FileSystem/Handle.hs
```diff
@@ -217,6 +219,21 @@ readChunksWithBufferOf = Unfold step return
...
+produceChunksWithBufferOf = Producer step return return
+    where
+    {-# INLINE_LATE step #-}
+    step (size, h) = do
+        arr <- liftIO $ readArrayUpto size h
+        return $
+            case A.length arr of
+                0 -> D.Stop
+                _ -> D.Yield arr (size, h)
+
```
harendra-kumar, line 236: Instead of introducing this we can use `Producer.fromStream` instead to convert an array stream to a producer.  It should not hurt performance. Since the producer type is experimental as of now, we should try to reduce changes to other code as much as possible.

adithyaov: Why would using `fromStreamD` it not hurt performance?

adithyaov: You're right, we should try to keep the changes for `Producer` as minimal as possible or all in one place.
Please answer the question above and close this PR.

harendra-kumar: > Why would using fromStreamD it not hurt performance? fromStreamD wraps the stream and has 1 level of indirection. Will it not affect fusion?

Because we are reading arrays which are big chunks, the perf overhead is scaled down by 1000s of times.

PR #955, opened 2021-02-22
Fix fromStreamD performance

harendra-kumar:
Before: not finishing in reasonable time
After: 686 us (similar to fromStream)

### DIFF
--- a/src/Streamly/Internal/Data/Unfold.hs
+++ b/src/Streamly/Internal/Data/Unfold.hs
@@ -351,7 +351,7 @@ fromStreamD = Unfold step return
     where
 
     {-# INLINE_LATE step #-}
-    step (Stream step1 state1) = do
+    step (UnStream step1 state1) = do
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -351,7 +351,7 @@ fromStreamD = Unfold step return
     where
 
     {-# INLINE_LATE step #-}
-    step (Stream step1 state1) = do
+    step (UnStream step1 state1) = do
```
adithyaov, line 354: How did this fix performance?

adithyaov (approved):
I don't know why this worked the way it worked.

src/Streamly/Internal/Data/Unfold.hs
harendra-kumar: That's an issue for GHC.

PR #958, opened 2021-02-23
Add Applicative instance for Producer

harendra-kumar:


adithyaov:
I don't see a `Monad` instance

harendra-kumar:
> I don't see a Monad instance

Yes, somehow I thought I included that in the commit. Skip that for now, just applicative in this PR.

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -33,42 +35,66 @@ where
...
+--
 -- | A @Producer m a b@ is a generator of a stream of values of type @b@ from a
 -- seed of type 'a' in 'Monad' @m@.
 --
 -- /Internal/
 
 data Producer m a b =
     -- | @Producer step inject extract@
-    forall s. Producer (s -> m (Step s b)) (a -> m s) (s -> m a)
+    forall s. Producer (s -> m (Step s a b)) (a -> m s) (s -> m (Maybe a))
```
adithyaov, line 76: This is probably a required change but I don't feel so right looking at it. Until we come up with a better abstraction this works.

adithyaov: Is it possible to just modify the `Step` to represent this?

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -33,42 +35,66 @@ where
...
+    fmap _ (Skip s) = Skip s
+    fmap _ (Stop s) = Stop s
 
 ------------------------------------------------------------------------------
 -- Type
 ------------------------------------------------------------------------------
 
+-- We need the return type of extract to be "Maybe a" so that we can represent
+-- a "used up" state for types that can not inherently represent that. For
+-- example, we may not be able to implement "const" correctly without it.
```
adithyaov, line 64: I don't perfectly understand this.

src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -133,6 +168,103 @@ instance Functor m => Functor (Producer m a) where
...
+
+    where
+
+    inject a = pure (Left a)
+
+    step (Left a) = (`Yield` Right a) <$> m
+    step (Right a) = pure $ Stop (Just a)
+
+    extract (Left a) = pure $ Just a
+    extract (Right a) = pure $ Just a
```
adithyaov, line 187: Should this be `Nothing`?

harendra-kumar:
Closing this for now.

PR #965, opened 2021-03-04
Remove Applicative instance of Fold and release teeWith

### LINKED ISSUE #499: Applicative instance of Fold and the new Tee type
See also #455 .

Since we are now including terminating/splitting functionality under the 'Fold' type we need to reconsider the existing applicative instance. The current applicative instance is consistent with the `foldl` library. However, with terminating folds a splitting applicative like the `Parser` type is possible for `Fold`. 

To keep Folds consistent with Parsers we can change the applicative instance of folds to a splitting applicative. The current applicative would become a `teeWith` function like in Parsers. This would be a breaking change. To mitigate the impact of the change to existing users we can have a new `Tee` type whose applicative would behave just like the current Fold applicative. This would be similar to the way we have `ZipSerial` type in streams.

adithyaov:
Closes #499

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -36,6 +36,7 @@
 
 * New encoding/decoding routines, `encodeUtf8'`, `encodeLatin1'`, `decodeUtf8'`,
   are added, these routines fail when they encounter any invalid characters.
+* `teeWith` is released as a part of `Streamly.Data.Fold`
 
 ### Bug Fixes
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,5 @@
 ## Unreleased
 
-### Notice
-
-* In the next major release the behaviour of `<*>` for folds will be updated
-  from `teeWith` to `splitWith` to reflect it's similarity with parsers.
-
 ### Behavioral changes
 
 * `Streamly.Prelude.fold` can now terminate early without consuming the entire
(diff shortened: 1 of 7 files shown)


Changelog.md
```diff
@@ -2,8 +2,8 @@
 
 ### Notice
 
-* In the next major release the behaviour of `<*>` for folds will be updated
-  from `teeWith` to `splitWith` to reflect it's similarity with parsers.
+* In the next major release `Applicative` for folds will be re-added and the
```
harendra-kumar, line 5: We can remove the notice.

Changelog.md
```diff
@@ -70,6 +71,8 @@
     * `mapM` to `rmapM`
     * `sequence` to `rsequence`
 
+* Remove `Applicative` instance of folds.
```
harendra-kumar, line 74: This should go in breaking changes, not in deprecations.

harendra-kumar: Also, add "Please use `teeWith` or the `Tee` type as an alternative to Fold applicative.".

harendra-kumar (changes requested):
Also remove the `Semigroup`, `Monoid`, `Num`, `Fractional`, `Floating` instances. They belong to the `Tee` type. For Fold type if we want to have those then they should agree with the Fold applicative.

adithyaov:
> Also remove the Semigroup, Monoid, Num, Fractional, Floating instances. They belong to the Tee type. For Fold type if we want to have those then they should agree with the Fold applicative.

Yeah was about to ask you the behaviour.

harendra-kumar (approved):

adithyaov:
Closes #499

PR #968, opened 2021-03-07
Rename some fold combinators

harendra-kumar:
This commit has no functional change, only renaming and doc edits.

I have changed `Fold.map` back to Fold.lmap to avoid confusion with `fmap`. This is the only deviation from the names of the APIs in Stream module. It probably makes sense to distinguish names explicitly when we have the same operation as covariant as well as contravariant.

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -65,7 +65,7 @@ module Streamly.Internal.Data.Fold
 
     -- ** Collectors
     , toList
-    , toListRev  -- experimental
+    , toListRev
     -- $toListRev
     , toStream
     , toStreamRev
@@ -74,7 +74,7 @@ module Streamly.Internal.Data.Fold
     , drainN
     -- , lastN
     -- , (!!)
-    -- , genericIndex
+    , genericIndex
     , index
     , head
     -- , findM
@@ -110,6 +110,7 @@ module Streamly.Internal.Data.Fold
     -- ** Mapping
     , transform
     , map
+    , lmap
 
     --, lsequence
     , lmapM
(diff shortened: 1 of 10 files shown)


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -425,6 +419,9 @@ _Fold1 step = mkAccum step_ Nothing' toMaybe
 drainBy ::  Monad m => (a -> m b) -> Fold m a ()
 drainBy f = lmapM f drain
 
+-- |
```
adithyaov, line 422: Empty documentation? Intentional? Only for the `/Internal/` tag?

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -1042,9 +1042,9 @@ splitAt n fld = serialWith (,) (take n fld)
 -- Stops - when the predicate succeeds.
 --
 -- /Pre-release/
-{-# INLINE sliceSepBy #-}
-sliceSepBy :: Monad m => (a -> Bool) -> Fold m a b -> Fold m a b
-sliceSepBy predicate (Fold fstep finitial fextract) =
+{-# INLINE takeEndBy_ #-}
+takeEndBy_ :: Monad m => (a -> Bool) -> Fold m a b -> Fold m a b
```
adithyaov, line 1046: I am not very comfortable with `takeEndBy_` and `takeEndBy`. I'd like to have `while` somewhere is the word.
I don't have better suggestions so this should work for now.

src/Streamly/Internal/Data/Fold/Types.hs
```diff
@@ -727,27 +730,31 @@ concatMap f (Fold stepa initiala extracta) = Fold stepc initialc extractc
...
-map :: (a -> b) -> Fold m b r -> Fold m a r
-map f (Fold step begin done) = Fold step' begin done
+-- /Pre-release/
+{-# INLINABLE lmap #-}
+lmap :: (a -> b) -> Fold m b r -> Fold m a r
+lmap f (Fold step begin done) = Fold step' begin done
     where
     step' x a = step x (f a)
 
+-- XXX should be removed
```
adithyaov, line 748: Why not remove it? I see that this PR has functional changes. You will remove (If exported then deprecate) it later?

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: No time for that. This is just internal, not even pre-release.

adithyaov (approved):

PR #970, opened 2021-03-08
Change argument order in parser combinators 

### LINKED ISSUE #969: Change argument order in parser combinators
In the parser combinators where only one Fold or Parser argument is the target that receives the final input, it should always be the last argument to allow chaining during composition. Specifically the following signatures should be changed in Streamly.Internal.Data.Parser and other low level parser modules:

* count
* countBetween
* many
* manyP
* some
* manyTill
* manyTillP
* manyThen
* roundRobin
* retryMaxTotal
* retryMaxSuccessive

rnjtranjan:
Fix #969

### DIFF
--- a/src/Streamly/Internal/Data/Parser.hs
+++ b/src/Streamly/Internal/Data/Parser.hs
@@ -953,8 +953,8 @@ choice _ps = undefined
 --
 {-# INLINE manyP #-}
 manyP :: -- MonadCatch m =>
-    Parser m b c -> Parser m a b -> Parser m a c
-manyP _f _p = undefined -- K.toParserK $ D.manyP f (K.fromParserK p)
+    Parser m a b -> Parser m b c -> Parser m a c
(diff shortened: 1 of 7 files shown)


src/Streamly/Internal/Data/Parser.hs
```diff
@@ -1048,8 +1048,8 @@ manyTillP _f _p1 _p2 = undefined
 --
 {-# INLINE manyTill #-}
 manyTill :: MonadCatch m
-    => Fold m b c -> Parser m a b -> Parser m a x -> Parser m a c
-manyTill f p1 p2 =
+    => Parser m a b -> Parser m a x -> Fold m b c -> Parser m a c
```
adithyaov, line 1051: Is this the desired order here?

adithyaov (approved):

adithyaov:
I think you've missed `manyTillP`. The rest looks good.

PR #973, opened 2021-03-08
Update Fold module docs and expose some APIs

harendra-kumar:


src/Streamly/Data/Fold.hs
```diff
@@ -116,143 +173,82 @@ module Streamly.Data.Fold
...
-    , linsertBy
-    , lintersperseM
+    -- ** Filtering
+    , filter
+    , filterM
 
-    -- ** Reordering
-    , lreverse
-    -}
+    -- -- ** Mapping Filters
```
adithyaov, line 211: Extra `--`?

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -27,21 +28,21 @@ module Streamly.Internal.Data.Fold
     , mkFoldM
     , mkFoldM_
 
-    -- * Generators
+    -- * Folds
```
adithyaov, line 31: This is the `Fold` module. Don't think this heading is necessary.

src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -1119,16 +1221,21 @@ chunksOf2 n (Fold step1 initial1 extract1) (Fold2 step2 inject2 extract2) =
 -- XXX We can use asyncClock here. A parser can be used to return an input that
 -- arrives after the timeout.
 -- XXX If n is 0 return immediately in initial.
+-- XXX we should probably discard the input received after the timeout like
```
adithyaov, line 1224: Maybe keep it unreleased till we fix the behaviour?

src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
```diff
@@ -332,14 +332,17 @@ runSink = fold . toFold
 -- Running a Parser
 ------------------------------------------------------------------------------
 
--- | Parse a stream using the supplied 'Parser'.
+-- | Parse a stream using the supplied ParserD 'Parser'.
```
adithyaov, line 335:
```suggestion
-- | Parse a stream using the supplied ParserD 'PRD.Parser'.
```

src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
```diff
@@ -332,14 +332,17 @@ runSink = fold . toFold
...
 --
--- /Pre-release/
+-- /Internal/
 --
 {-# INLINE_NORMAL parseD #-}
 parseD :: MonadThrow m => PRD.Parser m a b -> SerialT m a -> m b
 parseD p = D.parse p . toStreamD
 
+-- | Parse a stream using the supplied ParserK 'Parser'.
```
adithyaov, line 343:
```suggestion
-- | Parse a stream using the supplied ParserK 'PRK.Parser'.
```

adithyaov (commented):
I did not tally the functions. Looks good. I might have missed
something. I **might** do another review later.

src/Streamly/Data/Fold.hs
harendra-kumar: That is intentional to comment out the haddock markup.

src/Streamly/Internal/Data/Fold.hs
harendra-kumar: The module is divided into `Folds` and `Fold Combinators`. This section needs to have a name, since it consists of folds that is an appropriate name for it.

src/Streamly/Internal/Data/Fold/Type.hs
harendra-kumar: We can do that.

adithyaov (approved):
I still did not tally the functions but I think we can trust the compiler for that.

PR #975, opened 2021-03-09
Rename Stream.yield & Stream.yieldM

pranaysashank:


harendra-kumar (commented):
Need to deprecate the old APIs instead of removing them.

harendra-kumar:
Let's rename these as follows:

* `yield` => `fromPure`
* `yieldM` => `fromEffect`

PR #977, opened 2021-03-10
Replace mkAccum* fold constructors with mkFoldl*

harendra-kumar:
Ran the benchmarks. There is no noticeable difference.


adithyaov (commented):
Looks good. Please fix the CI.
Also, update the changelog where it says `sequence to rsequence` under `Deprecations`.

PR #980, opened 2021-03-12
Make splitMany behaviour similar to parseMany

adithyaov:


adithyaov:
The last commit also has the tests. I'll post the benchmarks after running them. Don't think the benchmarks would differ much though (Just a 1 line change).

adithyaov:
No difference in the benchmarks.

harendra-kumar (approved):

PR #982, opened 2021-03-12
Split IsStream.Nesting into Expand and Shrink

harendra-kumar:
Sorts out the mess of combinators in the Nested module. Separated the Fold/Unfold based combinators in separate modules, also stream Fused vs CPS based combinators in separate modules. Later we can possibly rename the combinators with the same functionality to same names because they are in separate modules.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream.hs
@@ -15,7 +15,8 @@ module Streamly.Internal.Data.Stream.IsStream
     , module Streamly.Internal.Data.Stream.IsStream.Generate
     , module Streamly.Internal.Data.Stream.IsStream.Eliminate
     , module Streamly.Internal.Data.Stream.IsStream.Transform
-    , module Streamly.Internal.Data.Stream.IsStream.Nesting
+    , module Streamly.Internal.Data.Stream.IsStream.Expand
+    , module Streamly.Internal.Data.Stream.IsStream.Reduce
     , module Streamly.Internal.Data.Stream.IsStream.Exception
     , module Streamly.Internal.Data.Stream.IsStream.Lift
     , module Streamly.Internal.Data.Stream.IsStream.Top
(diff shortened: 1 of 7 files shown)


adithyaov (commented):
`IsStream.Expand` and `IsStream.Shrink` are not very clear.
There is no indication of nesting in any of these names.

I wonder if suck elaborate division is required.

I don't seems to have the big picture but if you think think
it'll help in the future then looks good.

I'm assuming `Shrink.Fold` has nothing to do with the type?
It has `parseMany` and `parseManyD` as well.

The `Fused` suffix is bugging me a little too.

`parseMany` should go into the unfused module?

When we later have `FoldK` and `FoldD` what then?
`Shrink.Fold` and `Shrink.Fold.Fused`?

The only 2 divisions I'm comfortable with are,
`Nesting.Shrink` and `Nesting.Expand`

harendra-kumar:
We have two versions of many combinators, one direct style based (with stream fusion) and the other CPS based (no stream fusion but works with large number of compositions which is absolutely essential). The may be useful in different situations. Now, these versions have to be named differently and have to be in the same module, how do we name them? Currently we `serial` and `append` for the same combinator, `wSerial` and `interleave` for the same combinator. There many more such cases and there will be more in future. This does not scale, we need different modules with the combinators having the same names to avoid confusion. That is the basic issue, breaking the Nested module in different modules is a peripheral issue. I can undo that or do that independently.

adithyaov:
This is how I see the problem and one possible solution.

DISCLAIMER: Some unstructured rambling below,

`IsStream` should not denote the fact that there are 2 types of combinators. It should contain the best solution for scaling (maybe this depends on the problem and there is no absolute).  We need to somehow abstract the idea of `StreamD` and `StreamK` from the end user.

For any other requirement users should manually use `fromStream*` and `toStream*`.  We can provide some helpers namely, `usingStreamD` & `usingStreamK`, etc.

If this was not the goal and distinction should be clearly made then we need to figure out what the majority is. Currently the underlying structure in most of the combinators is `StreamD`. In this case the default of all the combinators should be `StreamD` and combinators with `StreamK` as the underlying structure should be suffixed with `K`.

So, `serial` and `serialK`. The best practice for appending a lot of streams is using `serialK`.

This problem is only for `Serial` streams as the underlying structure can be both, `StreamD` and `StreamK`.

`IsStream` is basically polymorphic `Serial`.

adithyaov:
> That is the basic issue, breaking the Nested module in different modules is a peripheral issue.

Makes sense I guess.
I think we might eventually find a good solution to this.
There was a plan of distinguishing combinators of `Serial` & `Async` explicitly.
Thinking about that will probably lead us to a good solution of distinction.

harendra-kumar:
Another try. Divided combinators based on whether they are to be used with `concatMapWith` or `concatPairsWith`. Please see the haddock for details.

PR #983, opened 2021-03-16
Split IsStream.Nesting into Expand and Reduce

harendra-kumar:
Sorts out the mess of combinators in the Nested module.

Resurrecting #982 which was closed due to deletion of fold-docs branch.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
@@ -1,0 +1,810 @@
+-- |
+-- Module      : Streamly.Internal.Data.Stream.IsStream.Expand
+-- Copyright   : (c) 2017 Composewell Technologies
+-- License     : BSD-3-Clause
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+-- Expand a stream by combining two or more streams or by combining streams
+-- with unfolds.
+
+module Streamly.Internal.Data.Stream.IsStream.Expand
+    (
+    -- * Binary Combinators (Linear)
+    -- | Functions ending in the shape:
+    --
+    -- @t m a -> t m a -> t m a@.
+    --
+    -- The functions in this section have a linear or flat n-ary combining
+    -- characterstics. It means that when combined @n@ times (e.g. @a `serial`
+    -- b `serial` c ...@) the resulting expression will have an @O(n)@
+    -- complexity (instead O(n^2) for pair wise combinators described in the
+    -- next section. These functions can be used efficiently with
+    -- 'concatMapWith' et. al.  combinators that combine streams in a linear
+    -- fashion (contrast with 'concatPairsWith' which combines streams as a
+    -- binary tree).
+
+      serial
+    , ahead
+    , async
+    , wAsync
+    , parallel
+    , Par.parallelFst
+    , Par.parallelMin
+
+    -- * Binary Combinators (Pair Wise)
+    -- | Like the functions in the section above these functions also combine
+    -- two streams into a single stream but when used @n@ times linearly they
+    -- exhibit O(n^2) complexity. They are best combined in a binary tree
+    -- fashion using 'concatPairsWith' giving a @n * log n@ complexity.  Avoid
+    -- using these with 'concatMapWith' when combining a large or infinite
+    -- number of streams.
+
+    -- ** Append
+    , append
+
+    -- ** wSerial
+    -- | 'wSerial' is a CPS based stream interleaving functions. It can be
+    -- used with 'concatMapWith' as well, however, the interleaving behavior of
+    -- @n@ streams would be asymmetric giving exponentially more weightage to
+    -- streams that come earlier in the composition.
+    --
+    , wSerial
+    , Serial.wSerialFst
+    , Serial.wSerialMin
+
+    -- ** Interleave
+    -- | 'interleave' is like 'wSerial'  but using a direct style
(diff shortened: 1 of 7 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -0,0 +1,816 @@
...
+    , ahead
+    , async
+    , wAsync
+    , parallel
+    , Par.parallelFst
+    , Par.parallelMin
+
+    -- * Binary Combinators (Pair Wise)
+    -- | Like the functions in the section above these functions also combine
+    -- two streams in a single stream but when used @n@ times linearly they
```
adithyaov, line 38:
```suggestion
    -- two streams into a single stream but when used @n@ times linearly they
```

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -0,0 +1,816 @@
...
+    -- exhibit O(n^2) complexity. They are best combined in a binary tree
+    -- fashion using 'concatPairsWith' giving a @n * log n@ complexity.  Avoid
+    -- using these with 'concatMapWith' when combining a large or infinite
+    -- number of streams.
+
+    -- ** Append
+    , append
+
+    -- ** wSerial
+    -- | 'wSerial' is a CPS based stream interleaving functions. It can also be
```
adithyaov, line 48:
```suggestion
    -- | 'wSerial' is a CPS based stream interleaving functions. It can be
```

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -0,0 +1,816 @@
...
+    -- ** Append Many (Unfold)
+    -- | Unfold and flatten streams.
+    , concatUnfold
+    , concatUnfoldInterleave
+    , concatUnfoldRoundrobin
+
+    -- ** Interpose
+    -- | Insert effects between streams. Like concatUnfold but intersperses an
+    -- effect between the streams. A
+    -- special case of gintercalate.
```
adithyaov, line 99: You can put these in the same line.

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -0,0 +1,816 @@
...
+
+import Streamly.Internal.Data.Stream.Ahead (ahead)
+import Streamly.Internal.Data.Stream.Async (async, wAsync)
+import Streamly.Internal.Data.Stream.IsStream.Common
+    ( concatM
+    , concatMapM
+    , concatMap
+    , smapM
+    , yield
+    , yieldM)
```
adithyaov, line 167: Probably can put it in 1-2 lines.

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -0,0 +1,816 @@
...
+-- /Pre-release/
+{-# INLINE append #-}
+append ::(IsStream t, Monad m) => t m b -> t m b -> t m b
+append m1 m2 = fromStreamD $ D.append (toStreamD m1) (toStreamD m2)
+
+-------------------------------------------------------------------------------
+-- Interleaving
+-------------------------------------------------------------------------------
+
+-- XXX Same as 'wSerial'. We should perhaps rename wSerial to interleave.
```
adithyaov, line 227: I would appreciate the other way around.
`append` ~> `serialD`
`interleave` ~> `wSerialD`
or something similar.

src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
```diff
@@ -235,10 +106,40 @@ module Streamly.Internal.Data.Stream.IsStream.Nesting
     -- , slidingSessionBuffer
     -}
 
-    -- * Transform (Nested Containers)
-    -- Nested splitting
+    -- ** Parsing
```
adithyaov, line 109: You can add a heading similar to the previous two
```
-- * Reduce By Parsers
```

adithyaov (commented):
Looks fine. 
Do we want to differentiate between `K-ish` and `D-ish` combinators?
It probably depends on how in many cases this is required.
If it just for `cons-ish` or `append-ish` combinators then we really don't need to differentiate it via names (no suffix `D` & `K`).

src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
harendra-kumar: Ok for the first one. For moving to the end, the following functions `wordsBy` etc would be implemented as parsers in future.

src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
```diff
@@ -235,10 +106,40 @@ module Streamly.Internal.Data.Stream.IsStream.Nesting
     -- , slidingSessionBuffer
     -}
 
-    -- * Transform (Nested Containers)
-    -- Nested splitting
+    -- ** Reduce By Parsers
```
adithyaov, line 109: Reduce by fold had a level 1 heading `-- * Reduce by folds`

adithyaov (approved):

src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
harendra-kumar: ok, will change it.

PR #989, opened 2021-03-20
Rename concatUnfold to unfoldMany 

### LINKED ISSUE #986: Rename concatUnfold to unfoldMany
This is the unfold counterpart of `foldMany`. This will deprecate the old function.

We can also rename `concatUnfoldInterleave` to `unfoldManyInterleave` and `concatUnfoldRoundrobin` to `unfoldManyRoundRobin`.

rnjtranjan:
Fix #986

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -61,6 +61,7 @@
     * `foldWith` to `concatFoldableWith`
     * `foldMapWith` to `concatMapFoldableWith`
     * `forEachWith` to `concatForFoldableWith`
+* In `Streamly.Prelude`, `concatUnfold` is renamed to `unfoldMany`
 * The following encoding/decoding routines have been renamed:
     * `encodeUtf8Lax` to `encodeUtf8`
     * `encodeLatin1Lax` to `encodeLatin1`
--- a/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
@@ -89,12 +89,13 @@ module Streamly.Internal.Data.Stream.IsStream.Expand
 
     -- ** Append Many (Unfold)
     -- | Unfold and flatten streams.
+    , unfoldMany
+    , unfoldManyInterleave
+    , unfoldManyRoundRobin
     , concatUnfold
-    , concatUnfoldInterleave
-    , concatUnfoldRoundrobin
 
     -- ** Interpose
-    -- | Insert effects between streams. Like concatUnfold but intersperses an
+    -- | Insert effects between streams. Like unfoldMany but intersperses an
     -- effect between the streams. A special case of gintercalate.
     , interpose
     , interposeSuffix
@@ -102,7 +103,7 @@ module Streamly.Internal.Data.Stream.IsStream.Expand
 
     -- ** Intercalate
     -- | Insert Streams between Streams.
-    -- Like concatUnfold but intersperses streams from another source between
+    -- Like unfoldMany but intersperses streams from another source between
     -- the streams from the first source.
     , intercalate
     , intercalateSuffix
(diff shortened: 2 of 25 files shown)


adithyaov (changes requested):
I think you've missed renaming `concatUnfold` to `unfoldMany`. Can you also do that here?

Changelog.md
```diff
@@ -171,7 +171,7 @@
       `Streamly.FileSystem.Handle.read`, `Streamly.Data.Unicode.Stream.decodeUtf8` and
       `splitOnSuffix` with `Streamly.Data.Fold.toList` to split the
        stream to a stream of `String` separated by a newline.
-    * `toHandle` has been deprecated. Please use `intersperse` and `concatUnfold` to
+    * `toHandle` has been deprecated. Please use `intersperse` and `unfoldMany` to
```
adithyaov, line 174: Don't think we should modify the history of the changelog.

adithyaov (changes requested):
`concatUnfold` to `unfoldMany` is a breaking change.
We should deprecate `concatUnfold` and update the changelog accordingly.
I don't think we should modify the history in the changelog.

Changelog.md
```diff
@@ -247,7 +247,7 @@ concurrently using this.
 
 * Add the following to `Streamly.Prelude`:
     * `unfold`, `fold`, `scan` and `postscan`
-    * `concatUnfold` to concat a stream after unfolding each element
+    * `unfoldMany` to concat a stream after unfolding each element
```
harendra-kumar, line 250: Revert this change. We need to add a deprecation note in the unreleased section at the top.

src/Streamly/Internal/Data/Stream/StreamK/Type.hs
```diff
@@ -1181,8 +1181,8 @@ instance Monad m => Monad (Stream m) where
 
 {-
 -- Like concatMap but generates stream using an unfold function. Similar to
--- concatUnfold but for StreamK.
-concatUnfoldr :: IsStream t
+-- unfoldMany but for StreamK.
+unfoldManyr :: IsStream t
     => (b -> t m (Maybe (a, b))) -> t m b -> t m a
-concatUnfoldr = undefined
+unfoldManyr = undefined
```
harendra-kumar, line 1187: We can revert this renaming, this is more of a concatUnfoldr.

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -460,29 +460,29 @@ mergeAsyncByM f m1 m2 =
 -- therefore provide many times better performance.
 --
 -- @since 0.7.0
-{-# INLINE concatUnfold #-}
-concatUnfold ::(IsStream t, Monad m) => Unfold m a b -> t m a -> t m b
-concatUnfold u m = fromStreamD $ D.concatUnfold u (toStreamD m)
+{-# INLINE unfoldMany #-}
+unfoldMany ::(IsStream t, Monad m) => Unfold m a b -> t m a -> t m b
+unfoldMany u m = fromStreamD $ D.unfoldMany u (toStreamD m)
```
harendra-kumar, line 465: We need to retain the old function as it is, just add a `DEPRECATED` pragma. Introduce a new `unfoldMany` function and move the doc from `concatUnfold` to the new function, and change the since notation to 0.8.0. In `concatUnfold` just say "Same as unfoldMany".

src/Streamly/Prelude.hs
```diff
@@ -595,7 +595,7 @@ module Streamly.Prelude
     , concatMapWith
     , concatMap
     , concatMapM
-    , concatUnfold
+    , unfoldMany
```
harendra-kumar, line 598: We need to keep concatUnfold as well, it is only deprecated.

harendra-kumar (changes requested):

PR #990, opened 2021-03-22
Put all memory restrictions under a flag and enable it in the CI

adithyaov:


benchmark/streamly-benchmarks.cabal
```diff
@@ -144,6 +149,7 @@ library
 
 common bench-options
   import: compile-options, optimization-options, bench-depends
+  -- XXX The use of -retsopts is for runtime?
```
pranaysashank, line 152: yes?

benchmark/streamly-benchmarks.cabal
```diff
@@ -19,6 +19,11 @@ flag fusion-plugin
   manual: True
   default: False
 
+flag limit-memory
```
pranaysashank, line 22: huh one more flag, but its in benchmarks so its fine maybe?

pranaysashank (approved):

benchmark/streamly-benchmarks.cabal
adithyaov: Cool

adithyaov: Just to be clear about the question,
```
ghc-options: +RTS ... -RTS
```
Limits the memory for ghc itself while compilation.
Whereas,
```
ghc-options: -rtsopts +RTS ... -RTS
```
is for limiting the memory when the executable is running.
@pranaysashank Is my understanding correct?

.circleci/config.yml
```diff
@@ -185,7 +186,8 @@ jobs:
            environment:
              BUILD: "cabal-v2"
              GHCVER: "8.4.4"
-             CABAL_BUILD_OPTIONS: "--flag debug"
+             # XXX This hopefully overrides the <<*env ?
```
adithyaov, line 189: @pranaysashank Does this override?

benchmark/streamly-benchmarks.cabal
harendra-kumar, line 22: How about naming it `limit-build-mem` to convey that it is build time?

PR #992, opened 2021-03-23
Refactor Unfold module, export additional APIs

harendra-kumar:


adithyaov (approved):
I wonder if we can add some clean smart constuctors.

src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -655,9 +680,37 @@ dropWhileM f (Unfold step inject) = Unfold step' inject'
...
+--
+-- Note 'enumerateFromStepIntegral' is faster for integrals.
+--
+-- /Pre-release/
+--
+{-# INLINE enumerateFromStepNum #-}
+enumerateFromStepNum :: (Monad m, Num a) => a -> Unfold m a a
+enumerateFromStepNum stride = Unfold step return
+    where
+    -- XXX This is numerically unstable.
```
adithyaov, line 699: Numerically stable for floating-point values?

src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -862,6 +901,7 @@ data OuterProductState s1 s2 sy x y =
     OuterProductOuter s1 y | OuterProductInner s1 sy s2 x
 
 -- XXX this can be written in terms of "cross".
+-- XXX Remove this in favor of cross?
```
adithyaov, line 904: We can benchmark and remove it accordingly.

PR #993, opened 2021-03-23
Make enumerateFromStepNum numerically stable

harendra-kumar:


test/Streamly/Test/Data/Unfold.hs
```diff
@@ -203,15 +203,14 @@ enumerateFromStepNum =
                   lst = Prelude.take 10 $ List.unfoldr (\x -> Just (x, x + s)) f
                in testUnfoldD unf f lst
 
+#if MIN_VERSION_base(4,12,0)
```
adithyaov, line 206: Why this pragma?

adithyaov (approved):

test/Streamly/Test/Data/Unfold.hs
harendra-kumar: Fractional enumeration before this version is broken in base as well, it is not numerically stable.

PR #995, opened 2021-03-24
Change signatures of zipWith/zipWithM

harendra-kumar:


src/Streamly/Internal/Data/Unfold/Type.hs
```diff
@@ -125,7 +125,7 @@ yieldM m = Unfold step inject
 
 -- | Discards the unfold input and always returns the argument of 'yield'.
 yield :: Applicative m => b -> Unfold m a b
-yield = yieldM . pure
+yield = yieldM Prelude.. pure
```
adithyaov, line 128: `Prelude..` looks pretty weird. `.` is a very important operator and can be imported normally (unqualified).

src/Streamly/Data/Unfold.hs
```diff
@@ -94,7 +94,6 @@ module Streamly.Data.Unfold
     , dropWhileM
 
     -- ** Zipping
-    , zipWithM
```
adithyaov, line 97: Any reason for not exporting this?

adithyaov (approved):

src/Streamly/Internal/Data/Unfold/Type.hs
harendra-kumar: Its because when we define the Category instance it becomes ambiguous.

src/Streamly/Data/Unfold.hs
harendra-kumar: Just exporting a minimal essential ones in this release, we can export others later. Anyway they are available if someone needs them.

PR #996, opened 2021-03-24
Update the documentation in Data.Fold

adithyaov:
Probably needs to be improved more. The reasons for the changes are mentioned as review-comments.

### DIFF
--- a/src/Streamly/Data/Fold.hs
+++ b/src/Streamly/Data/Fold.hs
@@ -6,16 +6,15 @@
 -- Stability   : experimental
 -- Portability : GHC
 --
--- A 'Fold' is a sink or a consumer of a stream of values.  The 'Fold' type
--- consists of an accumulator and an effectful action that absorbs a value into
--- the accumulator.
+-- A 'Fold' is an abstraction that can act as a consumer for a stream of values.
+--
+-- A fold driver e.g. 'Streamly.Prelude.fold' pushes values from a stream to the
+-- 'Fold' one at a time, reducing the stream to a single value.
 --
 -- >>> import qualified Streamly.Data.Fold as Fold
 -- >>> import qualified Streamly.Prelude as Stream
 --
--- For example, a 'sum' Fold represents adding the input to the accumulated
--- sum.  A fold driver e.g. 'Streamly.Prelude.fold' pushes values from a stream
--- to the 'Fold' one at a time, reducing the stream to a single value.
+-- For example, a 'sum' Fold represents adding the input to the accumulated sum.
 --
 -- >>> Stream.fold Fold.sum $ Stream.fromList [1..100]
 -- 5050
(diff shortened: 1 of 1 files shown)


src/Streamly/Data/Fold.hs
```diff
@@ -6,16 +6,15 @@
 -- Stability   : experimental
 -- Portability : GHC
 --
--- A 'Fold' is a sink or a consumer of a stream of values.  The 'Fold' type
--- consists of an accumulator and an effectful action that absorbs a value into
```
adithyaov, line 10: We should not leak internal details. As the fold type keeps evolving the internal structure will change. It will be rather hard to describe the internal structure in simple terms.

src/Streamly/Data/Fold.hs
```diff
@@ -6,16 +6,15 @@
...
--- the accumulator.
+-- A 'Fold' is an abstraction that can act as a consumer for a stream of values.
+--
+-- A fold driver e.g. 'Streamly.Prelude.fold' pushes values from a stream to the
```
adithyaov, line 11: `driver` is an important part of the design. We also have `fold_` now. We should describe this as soon as possible in the doc.

src/Streamly/Data/Fold.hs
```diff
@@ -28,15 +27,14 @@
 -- >>> Data.List.foldl' (+) 0 [1..100]
 -- 5050
 --
--- 'Fold's have an early termination capability e.g. the 'head' fold would
--- terminate on an infinite stream:
---
--- >>> Stream.fold Fold.head $ Stream.fromList [1..]
--- Just 1
+-- 'Fold's are non-greedy consumers and have the have the capability to
```
adithyaov, line 30: I want to somehow portray the idea of consumption. Consumption of input.

src/Streamly/Data/Fold.hs
```diff
@@ -28,15 +27,14 @@
...
+-- terminate early. Being non-greedy, they only consume the part of the input
+-- stream that is required.
 --
--- The above example is similar to the following right fold:
```
adithyaov, line 37: It is not. It is similar to a terminating `foldl`. With respect to the behavior it is similar to `foldr` I guess but I don't know if it's the correct analogy to use here.

src/Streamly/Data/Fold.hs
```diff
@@ -65,14 +66,17 @@
 -- >>> Stream.fold f stream
 -- ("Even 50","Odd 50")
 --
--- Terminating folds can be combined to parse the stream serially such that the
```
adithyaov, line 68: We should probably introduce the term `Terminating folds` before use. In this case, we can just use `Folds` I guess.

src/Streamly/Data/Fold.hs
```diff
@@ -65,14 +66,17 @@
...
--- consumes the rest of the input until it terminates:
+-- Folds can be combined to parse the stream serially such that the first fold
+-- consumes the input until it terminates and the second fold consumes the rest
+-- of the input until it terminates:
 --
 -- >>> f = Fold.serialWith (,) (Fold.take 8 Fold.toList) (Fold.takeEndBy (== '\n') Fold.toList)
 -- >>> Stream.fold f $ Stream.fromList "header: hello\n"
 -- ("header: ","hello\n")
 --
+-- Please note that if the first fold is a non-terminating fold like 'sum', it
```
adithyaov, line 77: A cautionary statement. Can be removed I guess.

harendra-kumar:
This is obsolete now. Closing it.

PR #997, opened 2021-03-24
Refactor/rename/add basic unfold constructors

harendra-kumar:


src/Streamly/Data/Unfold.hs
```diff
@@ -63,9 +63,13 @@ module Streamly.Data.Unfold
     -- One to one correspondence with
     -- "Streamly.Internal.Data.Stream.IsStream.Generate"
 
+    -- ** Basic Constructors
+    , unfoldrM
+    , function
+    , functionM
```
adithyaov, line 69: `singleton` also made sense. It also indicates the number of elements in the stream. I would prefer `singleton` over `function` but this is also fine I guess. Up to you.

adithyaov (approved):

src/Streamly/Data/Unfold.hs
adithyaov: Ok, scratch that. `singleton` can be confused with `yield`. `function` sounds good.

PR #1003, opened 2021-03-30
Add Since: 0.8.0 annotations to new functions.

pranaysashank:


src/Streamly/Internal/Data/Stream/Async.hs
```diff
@@ -525,7 +525,7 @@ mkAsyncD m = D.Stream step Nothing
 -- kicked off again to evaluate the remaining stream when there is space in the
 -- buffer.  The consumer consumes the stream lazily from the buffer.
 --
--- /Pre-release/
+-- /Since: 0.8.0/
```
harendra-kumar, line 528: Remove the since annotation at line 518 above.

harendra-kumar (approved):

PR #1005, opened 2021-03-30
Refactoring StreamK benchmarks

adithyaov:


benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -122,12 +205,14 @@ init s = do
 tail :: (Monad m, S.IsStream t) => t m a -> m ()
 tail s = S.tail s >>= P.mapM_ tail
 
+-- XXX Should we ass a separate benchmark for null?
```
harendra-kumar, line 208: benchmark for `null` would be useless as it takes too little time.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -122,12 +205,14 @@ init s = do
...
 {-# INLINE nullTail #-}
 nullTail :: Monad m => Stream m Int -> m ()
 nullTail s = do
     r <- S.null s
     when (not r) $ S.tail s >>= P.mapM_ nullTail
 
+-- XXX Should we ass a separate benchmark for head?
```
harendra-kumar, line 215: benchmark for `head` would be useless as it takes too little time.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -166,10 +253,12 @@ composeN n f =
         4 -> transform . f . f . f . f
         _ -> undefined
 
+-- XXX We should rename this to scanl'
 {-# INLINE scan #-}
 scan :: Monad m => Int -> Stream m Int -> m ()
 scan n = composeN n $ S.scanl' (+) 0
 
+-- XXX We should use S.map here instead
```
harendra-kumar, line 261: Then use a separate benchmark for `fmap`?

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -242,6 +331,9 @@ intersperse streamLen n = composeN n $ S.intersperse streamLen
 -- Iteration
 -------------------------------------------------------------------------------
 
+-- XXX iterate is a reserved term
+-- We should probably put iterate at the end as a suffix
+-- Or we can probably call this sourceXN or something.
```
harendra-kumar, line 336: I think its fine as it is, does not make much difference changing it.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -394,18 +488,21 @@ concatStreamsWith outer inner n =
 -- Nested Composition
 -------------------------------------------------------------------------------
 
+-- XXX This can be renamed to applicativeInstance
```
harendra-kumar, line 491: Could be `drainApplicative`.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -394,18 +488,21 @@ concatStreamsWith outer inner n =
...
 {-# INLINE toNullApNested #-}
 toNullApNested :: Monad m => Stream m Int -> m ()
 toNullApNested s = runStream $ do
     (+) <$> s <*> s
 
+-- XXX This can be renamed to monadInstance
```
harendra-kumar, line 497: could be `drainMonad` or `drainBind`.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -414,6 +511,9 @@ toNullNested3 s = runStream $ do
     z <- s
     return $ x + y + z
 
+-- XXX We should indicate monadInstance also in the name.
+-- We can skip the use of filter in the name. Can be misleading.
+-- Maybe monadInstanceConditionOut
```
harendra-kumar, line 516: could be `filterAllOutMonad`.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -442,6 +543,9 @@ filterAllInNested str = runStream $ do
 -- Nested Composition Pure lists
 -------------------------------------------------------------------------------
 
+-- XXX We should remove this? Or move this elsewhere. We have streaming
+-- benchmarks for comparision I guess.
```
harendra-kumar, line 547: There are several list benchmarks here for comparison with lists. It is easy and convenient to see the comparisons when they are here, otherwise you will have to add a separate module for list benchmarks with the same names and then add a comparison in bench.sh.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -40,12 +40,87 @@ import Streamly.Benchmark.Common
...
+-- zipWithM
+-- mergeBy
+-- mergeByM
+
+-- toStreamK (Probably can be skipped)
+-- hoist
+
+-- scanlx'
+
+-}
```
harendra-kumar, line 115: Probably do not add all of them as it may take a lot of time, select a few important ones and add those. Especially those which are actually being used in Streamly.Prelude etc. A lot of these are not actually used.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
adithyaov: There's already a separate benchmark for `fmap`

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
adithyaov: Guess we can keep this for now.

adithyaov: I'm keeping the XXX comment though.

PR #1008, opened 2021-04-01
Implement StreamD version of concatPairWith

rnjtranjan:
Fix # 1007

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
@@ -1038,11 +1038,16 @@ concatPairsWith par f m =
 
 {-# INLINE concatPairsWithD #-}
 concatPairsWithD :: (IsStream t , Monad m) =>
-       (S.Stream m b -> S.Stream m b -> S.Stream m b)
-    -> (a -> S.Stream m b)
+       (t m b -> t m b -> t m b)
+    -> (a -> t m b)
     -> t m a
     -> t m b
-concatPairsWithD c f z = D.fromStreamD $ S.concatPairsWith c f (D.toStreamD z)
+concatPairsWithD par f m =
+    fromStreamD
+        $ D.concatPairsWith
+            (\s1 s2 -> toStreamD $ fromStreamD s1 `par` fromStreamD s2)
+            (toStreamD . f)
+            (toStreamD m)
 
 ------------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -704,6 +705,14 @@ concatPairsWith :: IsStream t =>
     -> t m b
 concatPairsWith = K.concatPairsWith
 
+{-# INLINE concatPairsWithD #-}
+concatPairsWithD :: (IsStream t , Monad m) =>
+       (S.Stream m b -> S.Stream m b -> S.Stream m b)
+    -> (a -> S.Stream m b)
+    -> t m a
+    -> t m b
```
adithyaov, line 713: I have a few problems with this.

Any kind of combination function is more efficient with K-ish
combinators.

Users using IsStream should not be able to know what `StreamD`
is. I believe the high level picture is to abstract it out. Or
even if it isn't I don't see where this can be used.

We can maybe have a `concatPairsWith` in `StreamD` itself and use
`fromStreamD` to lift it here.

Then the signature would be similar to `concatPairsWith`. But
then again, I don't see how and where that would be needed.

rnjtranjan:
Benchmark comparison:

Prelude.Serial(cpuTime)
Benchmark                                default(ms)
---------------------------------------- -----------
Prelude.Serial/o-n-heap/buffered/sortByD     1656.54
Prelude.Serial/o-n-heap/buffered/sortBy       137.57

Prelude.Serial(allocated)
Benchmark                                default(MiB)
---------------------------------------- ------------
Prelude.Serial/o-n-heap/buffered/sortByD      8247.77
Prelude.Serial/o-n-heap/buffered/sortBy        284.02

Prelude.Serial(bytesCopied)
Benchmark                                default(MiB)
---------------------------------------- ------------
Prelude.Serial/o-n-heap/buffered/sortByD        84.97
Prelude.Serial/o-n-heap/buffered/sortBy         60.62

adithyaov:
Looks like `sortByD` is over 10 times slower.

harendra-kumar:
Its probably because of some silly implementation issue, causing too many allocations.

harendra-kumar:
I rebased it, and improved the implementation by using lists but it is still around 7x slower.

harendra-kumar:
This is not going anywhere, closing it.

PR #1009, opened 2021-04-01
Update prelude docs (folds)

harendra-kumar:
* Take into account terminating fold changes.
* Convert examples to doctests
* Use full module names


src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -566,13 +566,31 @@ trace_ eff = Serial.mapM (\x -> eff >> return x)
 
 -- | Scan a stream using the given monadic fold.
 --
+-- Non-terminating folds can be terminated by a scan.  For example,
```
adithyaov, line 569: I don't understand this statement. I can't see how the example below justifies this statement.

adithyaov (approved):

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
harendra-kumar: I will remove it.

Issue #921, opened 2021-02-08
Add intelligent slicing combinators to `Array-ish` modules


adithyaov:
```
unsafeOffset :: Int -> Array.Array a -> Array.Array a
```
and
```
unsafeCut :: Int -> Array.Array a -> Array.Array a
```
or,
```
unsafeSlice :: Int -> Int -> Array.Array a -> Array.Array a
```

@harendra-kumar Thoughts?

adithyaov:
Also see #661

harendra-kumar:
We have slicing operations now. Closing this.

PR #1014, opened 2021-04-02
Add a basic array slicing combinator

adithyaov:
https://github.com/composewell/streamly/issues/921


adithyaov:
We can probably have safe wrappers over `unsafeSlice`

harendra-kumar (approved):

PR #1028, opened 2021-04-06
Rename fold constructors in Data.Fold 

### LINKED ISSUE #1016: Rename fold constructors in Data.Fold to make them consistent with Streamly.Prelude
```
mkFoldl => foldl'
mkFoldlM => foldlM'
mkFoldr => foldr
```

However, `fold` in `Streamly.Prelude` would still remain inconsistent because it is used to run a fold rather than creating a fold, for creation we would still have `mkFold` for the general fold construction case.

rnjtranjan:
Fix #1016

### DIFF
--- a/src/Streamly/Data/Fold.hs
+++ b/src/Streamly/Data/Fold.hs
@@ -119,9 +119,9 @@ module Streamly.Data.Fold
       Fold -- (..)
 
     -- * Constructors
-    , mkFoldl
-    , mkFoldlM
-    , mkFoldr
+    , foldl'
+    , foldlM'
+    , foldr
 
     -- * Folds
     -- ** Accumulators
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -21,7 +21,7 @@ module Streamly.Internal.Data.Fold
     -- * Constructors
     , foldl'
     , foldlM'
-    , foldl1
+    , foldl1'
     , foldr
     , foldrM
     , mkFold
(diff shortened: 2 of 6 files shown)


src/Streamly/Data/Fold.hs
```diff
@@ -261,3 +261,15 @@ import Prelude
                span, splitAt, break, mapM)
 
 import Streamly.Internal.Data.Fold
+
+{-# INLINE foldl' #-}
+foldl' :: Monad m => (b -> a -> b) -> b -> Fold m a b
+foldl' = mkFoldl
```
harendra-kumar, line 267: We should rename `mkFoldl` itself rather than defining a new function.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -19,10 +19,10 @@ module Streamly.Internal.Data.Fold
...
 
     -- * Constructors
-    , mkFoldl
-    , mkFoldlM
+    , foldl'
+    , foldlM'
     , mkFoldl1
-    , mkFoldr
+    , foldr
     , mkFoldrM
```
harendra-kumar, line 26: We can change `mkFoldl1` to `foldl1` and `mkFoldrM` to `foldrM` at the same time. Only `mkFold` and `mkFold_` should remain unchanged.

harendra-kumar (changes requested):

adithyaov (approved):

adithyaov:
@rnjtranjan Can you fix up the initial commits?

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -19,11 +19,11 @@ module Streamly.Internal.Data.Fold
...
 
     -- * Constructors
-    , mkFoldl
-    , mkFoldlM
-    , mkFoldl1
-    , mkFoldr
-    , mkFoldrM
+    , foldl'
+    , foldlM'
+    , foldl1
```
harendra-kumar, line 24: I am sorry, this had to be `foldl1'` instead. The same as we have in "Streamly.Prelude".

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #1036, opened 2021-04-14
Parser fixes

harendra-kumar:


src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -206,20 +217,13 @@ parse_ (PRD.Parser pstep initial extract) (Stream step state) = do
                 let (src0, buf1) = splitAt n (x:getList buf)
                     src  = Prelude.reverse src0 ++ xs
                 gobuf SPEC s (List buf1) (List src) pst1
-            PR.Done _ b -> return (b, Stream step_rest (Left (s, buf)))
+            PR.Done n b -> do
```
adithyaov, line 220: Have a `Done 0` case here as well?

src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -424,7 +424,7 @@ noErrorUnsafeSplitWith func (Parser stepL initialL extractL)
             -- Assume that the first parser can never fail, therefore we do not
             -- need to keep the input for backtracking.
             Partial n s -> return $ Partial n (SeqParseL s)
-            Continue n s -> return $ Partial n (SeqParseL s)
+            Continue n s -> return $ Continue n (SeqParseL s)
```
adithyaov, line 427: This should be `Partial`. `noErrorUnsafeSplitWith` isf or parsers that don't fail.

src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -438,7 +438,7 @@ noErrorUnsafeSplitWith func (Parser stepL initialL extractL)
         r <- stepR st a
         return $ case r of
             Partial n s -> Partial n (SeqParseR f s)
-            Continue n s -> Partial n (SeqParseR f s)
+            Continue n s -> Continue n (SeqParseR f s)
```
adithyaov, line 441: This should be `Partial`. `noErrorUnsafeSplitWith` isf or parsers that don't fail.

src/Streamly/Internal/Data/Parser/ParserD/Type.hs
harendra-kumar: Continue does not mean failure, its only backtracking. The inner parser wants to backtrack but you are dropping the backtrack buffer and continuing from the current position that would make the behavior of the parser incorrect.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
harendra-kumar: Yes, that would avoid an unnecessary `append` which is detrimental to perf.

adithyaov (commented):
You can merge it after adding a `Done 0` case.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
harendra-kumar: the `Done 0` case won't be very useful here as the append would still be needed. And the rest of the code like `take` would anyway be simplified. So I am keeping it as is.

PR #1038, opened 2021-04-17
Implement undefined functions in IsStream.Top module

rnjtranjan:


src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -325,9 +343,9 @@ mergeInnerJoin = undefined
 --
 -- Time: O(m x n)
 --
--- /Unimplemented/
+-- /Pre-release/
```
harendra-kumar, line 346: If I remember correctly, this function had some issue because of which it was marked as unimplemented. Did you review and test it? Does it always produce correct output? @adithyaov do you remember the issue since you reviewed it?

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -325,9 +343,9 @@ mergeInnerJoin = undefined
...
 {-# INLINE leftJoin #-}
-leftJoin :: Monad m =>
+leftJoin :: ( MonadIO m) =>
```
harendra-kumar, line 348: Why is this changed?

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -356,23 +374,44 @@ leftJoin eq s1 s2 = Stream.evalStateT (return False) $ do
...
+-- /Pre-release/
 {-# INLINE hashLeftJoin #-}
-hashLeftJoin :: -- Hashable b =>
-    (a -> b -> Bool) -> t m a -> t m b -> t m (a, Maybe b)
-hashLeftJoin = undefined
+hashLeftJoin ::  (IsStream t, Eq a,  Hashable a, Monad m, Monad (t m)) =>
+    t m a -> t m a -> t m (a, Maybe a)
+hashLeftJoin  s1 s2 =
+    Stream.concatM $ do
+        l2 <- StreamK.toList $ fmap (, ()) s2
```
harendra-kumar, line 383: Same as discussed in hashInnerJoin.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -356,23 +374,44 @@ leftJoin eq s1 s2 = Stream.evalStateT (return False) $ do
...
 {-# INLINE mergeLeftJoin #-}
-mergeLeftJoin :: -- Monad m =>
+mergeLeftJoin :: forall (t :: (Type -> Type) -> Type -> Type) m a b.
+    (IsStream t, Monad m, Monad (t m), MonadTrans t) =>
     (a -> b -> Ordering) -> t m a -> t m b -> t m (a, Maybe b)
-mergeLeftJoin _eq _s1 _s2 = undefined
+mergeLeftJoin cmp s1 s2 = do
+    a <- s1
+    let ss =  Stream.filter (\x -> a `cmp` x == EQ )
+            $ Stream.takeWhile (\x -> a `cmp` x == GT || a `cmp` x == EQ) s2
```
harendra-kumar, line 405: Same as discussed in mergeInnerJoin.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -505,12 +580,22 @@ intersectBy eq s1 s2 =
...
-
+mergeIntersectBy :: (IsStream t, Monad m, Monad (t m), Ord a) =>
+    (a -> a -> Bool) -> t m a -> t m a -> t m a
+mergeIntersectBy eq s1 s2 = concatM $ do
+    -- This may work well when s2 is small
+    let sq = Stream.uniqBy eq s2
+    let k1 = do
+            a <- s1
+            let ss =  Stream.filter (eq a)
+                    $ Stream.takeWhile (a >=) sq
```
harendra-kumar, line 593: Same discussion as in mergeJoins applies here as well. The complexity of the algo should be O(m+n). All merge style operations should have linear complexity as we do not need to go through the streams twice. You can take two sorted lists and see how the algo works, you just need to look at the current elements of the two streams to determine what to do.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -865,6 +866,11 @@ nubBy :: -- (IsStream t, Monad m) =>
     (a -> a -> Bool) -> t m a -> t m a
 nubBy = undefined -- fromStreamD . D.nubBy . toStreamD
 
+{-# INLINE nubByMerge #-}
+nubByMerge :: (IsStream t, Monad m) =>
+    (a -> a -> Bool) -> t m a -> t m a
+nubByMerge eq = fromStreamD . D.nubByMerge eq . toStreamD
```
harendra-kumar, line 872: Isn't this same as `uniqBy` above?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -285,22 +287,38 @@ innerJoin eq s1 s2 = do
...
+    t m a -> t m a -> t m a
+hashInnerJoin s1 s2 =
+    Stream.concatM $ do
+        l2 <- StreamK.toList $ fmap (, ()) s2
+        let res = do
+                e <- s1
+                if e `HM.member` HM.fromList l2
+                then  return e
+                else  StreamK.nil
+        return res
```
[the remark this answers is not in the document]
rnjtranjan: Not clear, may need short discussion.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
harendra-kumar: Let's keep it unimplemented unless we have a correct implementation.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -356,23 +370,34 @@ leftJoin eq s1 s2 = Stream.evalStateT (return False) $ do
...
 --
--- /Unimplemented/
+-- /Pre-release/
 {-# INLINE mergeLeftJoin #-}
-mergeLeftJoin :: -- Monad m =>
+mergeLeftJoin :: forall (t :: (Type -> Type) -> Type -> Type) m a b.
+    (IsStream t, MonadIO m, Eq a, Eq b) =>
     (a -> b -> Ordering) -> t m a -> t m b -> t m (a, Maybe b)
-mergeLeftJoin _eq _s1 _s2 = undefined
+mergeLeftJoin  = Stream.leftJoinByMerge
```
harendra-kumar, line 400: Let's keep the name of the function same in this module as well as in the `Stream` module.

harendra-kumar: This applies to all such names in this module.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -202,6 +203,13 @@ module Streamly.Internal.Data.Stream.IsStream.Transform
 
     -- * Deprecated
     , scanx
+    , removeDupsRight
+    , differenceByMerge
+    , innerJoinByMerge
+    , leftJoinByMerge
+    , rightJoinByMerge
+    , outerJoinByMerge
+    , intersectByMerge
```
harendra-kumar, line 212: These should not be added in the `Deprecated` section.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
harendra-kumar: Won't both of them be equivalent when the stream is sorted? Looks like, only the implementation is different, the algo is the same and the functionality is exactly the same.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -356,23 +370,34 @@ leftJoin eq s1 s2 = Stream.evalStateT (return False) $ do
...
 -- Time: O(m + n)
 --
--- /Unimplemented/
+-- /Pre-release/
 {-# INLINE hashLeftJoin #-}
-hashLeftJoin :: -- Hashable b =>
-    (a -> b -> Bool) -> t m a -> t m b -> t m (a, Maybe b)
-hashLeftJoin = undefined
+hashLeftJoin ::  (IsStream t, Ord a, Monad m, Monad (t m)) =>
+    t m a -> t m a -> t m (a, Maybe a)
```
harendra-kumar, line 376: The signature should be similar to hashInnerJoin above i.e. `t m (k, a) -> t m (k, b) -> t m (k, a, Maybe b)`.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -387,7 +412,7 @@ mergeLeftJoin _eq _s1 _s2 = undefined
 --
 -- Time: O(m x n)
 --
--- /Unimplemented/
+-- /Pre-release/
```
harendra-kumar, line 415: This function has a XXX comment below ("Need to use mutable array"), it does not work correctly, so let's mark it unimplemented unless we fix that.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -443,23 +468,43 @@ outerJoin eq s1 s =
...
 -- Time: O(m + n)
 --
--- /Unimplemented/
+-- /Pre-release/
 {-# INLINE hashOuterJoin #-}
-hashOuterJoin :: -- (Monad m, Hashable b) =>
-    (a -> b -> Ordering) -> t m a -> t m b -> t m (Maybe a, Maybe b)
-hashOuterJoin _eq _s1 _s2 = undefined
+hashOuterJoin :: (IsStream t, Ord a, Monad (t m), MonadAsync m , Semigroup (t m (Maybe a, Maybe a))) =>
+    t m a -> t m a -> t m (Maybe a, Maybe a)
```
harendra-kumar, line 474: See the signature comment in `hashLeftJoin`, same applies here.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -865,6 +873,47 @@ nubBy :: -- (IsStream t, Monad m) =>
...
+rightJoinByMerge eq s1 = fromStreamD . D.mergeByRightInner eq (toStreamD s1) . toStreamD
+
+outerJoinByMerge :: (IsStream t, MonadIO m,  Eq a, Eq b)  =>
+    (a -> b -> Ordering) -> t m a -> t m b -> t m (Maybe a, Maybe b)
+outerJoinByMerge eq s1 = fromStreamD . D.mergeOuterJoin eq (toStreamD s1) . toStreamD    
+
+--mergeByIntersect
+intersectByMerge :: (IsStream t, MonadIO m, Eq a) =>
+    (a -> a -> Ordering) -> t m a -> t m a -> t m a
+intersectByMerge eq s1 = fromStreamD . D.mergeByIntersect eq (toStreamD s1) . toStreamD
```
harendra-kumar, line 913: Why add all these definitions in Transform module and then use them in the Top module? Why not directly define in the Top module? This seems like additional boilerplate code without any use.

src/Streamly/Internal/Data/Stream/StreamD/Generate.hs
```diff
@@ -403,3 +404,24 @@ fromListM = Stream step
     {-# INLINE_LATE step #-}
     step _ (m:ms) = m >>= \x -> return $ Yield x ms
     step _ []     = return Stop
+
+{-# INLINE_NORMAL nubByMerge #-}
+nubByMerge :: (Monad m) => (a -> a -> Bool) -> Stream m a -> Stream m a
+nubByMerge eq (Stream stepa sa) = Stream step (sa, Nothing)
```
harendra-kumar, line 410: Why is this in Generate module, this is not a stream generator?

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -476,6 +487,827 @@ mergeBy
     => (a -> a -> Ordering) -> Stream m a -> Stream m a -> Stream m a
 mergeBy cmp = mergeByM (\a b -> return $ cmp a b)
 
+{-# INLINE mergeByInner #-}
+mergeByInner
+    :: (MonadIO m,  Eq a, Eq b)
+    => (a -> b -> Ordering) -> Stream m a -> Stream m b -> Stream m (a, b)
+mergeByInner cmp = mergeInnerByM (\a b -> return $ cmp a b)
```
harendra-kumar, line 494: Follow the "define before use" guideline.

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -1054,7 +1886,7 @@ parseMany (PRD.Parser pstep initial extract) (Stream step state) =
                  in return $ Skip $ ParseChunksYield pb next
             PRD.IError err -> throwM $ ParseError err
 
-    -- XXX we just discard any leftover input at the end
+    -- XXX we Just discard any leftover input at the end
```
harendra-kumar, line 1889: Why change this?

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -1170,7 +2002,7 @@ parseIterate func seed (Stream step state) =
                  in return $ Skip $ ConcatParseYield pb next
             PRD.IError err -> throwM $ ParseError err
 
-    -- XXX we just discard any leftover input at the end
+    -- XXX we Just discard any leftover input at the end
```
harendra-kumar, line 2005: Why change this?

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -476,6 +487,827 @@ mergeBy
...
+mergeByInner
+    :: (MonadIO m,  Eq a, Eq b)
+    => (a -> b -> Ordering) -> Stream m a -> Stream m b -> Stream m (a, b)
+mergeByInner cmp = mergeInnerByM (\a b -> return $ cmp a b)
+
+{-# INLINE_NORMAL mergeInnerByM #-}
+mergeInnerByM
+    :: (MonadIO m,  Eq a, Eq b)
+    => (a -> b -> m Ordering) -> Stream m a -> Stream m b -> Stream m (a, b)
+mergeInnerByM cmp (Stream stepa ta) (Stream stepb tb) = do
```
harendra-kumar, line 500: Use the same name as the name of the corresponding join combinator in Top module.

harendra-kumar (changes requested):
Let's name the functions like this:

```
join
joinInner
joinInnerHash
joinInnerMerge
joinLeft
joinLeftHash
joinLeftMerge
joinOuter
joinOuterHash
joinOuterMerge

intersectBySorted
differenceBySorted
unionBySorted
```

I have not reviewed the Nesting module, will review in the next iteration.
Please make the changes and push.

harendra-kumar:
Please fix the failing CIs.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -601,8 +660,8 @@ unionBy eq s1 s2 =
...
 --
--- /Unimplemented/
-{-# INLINE mergeUnionBy #-}
-mergeUnionBy :: -- (IsStream t, Monad m) =>
+-- /Pre-release/
+{-# INLINE unionBySorted #-}
+unionBySorted :: (IsStream t, MonadAsync m, Semigroup (t m a)) =>
     (a -> a -> Ordering) -> t m a -> t m a -> t m a
-mergeUnionBy _eq _s1 _s2 = undefined
+unionBySorted eq s1 s2 = s1 <> Stream.removeDupsRight eq s1 s2
```
harendra-kumar, line 667: 1. We should not use the stream `s1` twice, that's a functional behavior issue.
2. This is inefficient because we go through the stream s1 twice, performance issue.

Instead we can go through both the streams in tandem only once and union them.

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -1,26 +1,26 @@
...
+--     forall a b. S.toStreamK (S.unfoldr a b) = K.unfoldr a b #-}```-- transformations are like the "Eliminate" module but they transform a stream
+---- by eliminating parts of the stream instead of eliminating the whole stream.
+---- Then, fromStreamK/toStreamK are inlined in the last phase:
+---- These combinators involve transformation, generation, elimination so can be
+-- {-# INLINE_LATE toStreamK #-}-- classified under any of those.
+---- toStreamK :: Monad m => Stream m a -> K.Stream m a```
+---- Ultimately these operations should be supported by Unfolds, Pipes and Folds,
+-- and this module may become redundant.-- The fallback rules make sure that if we could not fuse the direct style
+-- operations then better use the CPS style operation, because unfused direct
+-- style would have worse performance than the CPS style ops.-- The zipWithM combinator in this module has been adapted from the vector
```
harendra-kumar, line 23: I don't understand all these changes. Is this by accident?

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -183,14 +193,14 @@ data AppendState s1 s2 = AppendFirst s1 | AppendSecond s2
 -- the number of streams being composed increases this may become expensive.
 -- Need to see where the breaking point is between the two.
 --
-{-# INLINE_NORMAL append #-}
+{-# INLINE [1] append #-}
```
harendra-kumar, line 196: why this change?

harendra-kumar (changes requested):
Please address these comments, I will review the Nesting module later.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -277,30 +282,42 @@ innerJoin eq s1 s2 = do
...
-    (a -> b -> Bool) -> t m a -> t m b -> t m (a, b)
-hashInnerJoin = undefined
+-- /Pre-release/
+{-# INLINE joinInnerHash #-}
+joinInnerHash :: (IsStream t, Monad m, Monad (t m), Ord k) =>
+    t m (k, a) -> t m (k, b) -> t m (k, a, b)
+joinInnerHash s1 s2 =
+    Stream.concatM $ do
+        l2 <- Stream.fold (Fold.classify Fold.toList) $ StreamK.adapt s2
+        return $ do
```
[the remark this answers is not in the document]
rnjtranjan: It gives compile error below:
   • Occurs check: cannot construct the infinite type: m ~ t m
      Expected type: t m (Map.Map k [b])
        Actual type: m (Map.Map k [b])

harendra-kumar:
Do you need to bump the upper bound of hashable for ghc 9.0.1?

benchmark/Streamly/Benchmark/Prelude/Serial/Transformation2.hs
```diff
@@ -187,6 +241,17 @@ o_n_heap_buffering value =
...
+        , benchIOSink value "joinInnerHash" joinInnerHash
+        , benchIOSink value "joinInnerMerge" joinInnerMerge
+        , benchIOSink value "joinLeft" joinLeft
+        , benchIOSink value "joinLeftHash" joinLeftHash
+        , benchIOSink value "joinLeftMerge" joinLeftMerge
+        , benchIOSink value "joinOuter" joinOuter
+        , benchIOSink value "joinOuterHash" joinOuterHash
+        , benchIOSink value "joinOuterMerge" joinOuterMerge
+        , benchIOSink value "differenceBySorted" (differenceBySorted 1)
+        , benchIOSink value "unionBySorted" (unionBySorted 1)
```
harendra-kumar, line 254: Can we also add `intersectBySorted` benchmark?

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -350,29 +366,40 @@ leftJoin eq s1 s2 = Stream.evalStateT (return False) $ do
             else StreamK.nil
         Nothing -> return (a, Nothing)
 
--- | Like 'outerJoin' but uses a hashmap for efficiency.
+-- | Like 'joinLeft' but uses a hashmap for efficiency.
```
harendra-kumar, line 369: Add "Unlike 'joinLeft', both streams are evaluated only once".

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -896,6 +900,32 @@ nubBy :: -- (IsStream t, Monad m) =>
     (a -> a -> Bool) -> t m a -> t m a
 nubBy = undefined -- fromStreamD . D.nubBy . toStreamD
 
+{-# INLINE_NORMAL nubByMerge #-}
+nubByMerge :: (Monad m) => (a -> a -> Bool) -> D.Stream m a -> D.Stream m a
```
harendra-kumar, line 904: 1. This is the same as uniqBy if uniqBy is applied on a sorted stream. We do not want to keep two combinators doing the same thing.
2. If you think this implementation is faster, then replace `uniqBy` by this implementation, add a benchmark for uniqBy, and compare the results of the previous implementation with the new one. If this is faster we can keep this otherwise we will keep the old one as it is more idiomatic.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -202,6 +203,8 @@ module Streamly.Internal.Data.Stream.IsStream.Transform
 
     -- * Deprecated
     , scanx
+    , removeDupsRight
```
harendra-kumar, line 206: What is the purpose of this? Where is this used?

src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -476,6 +490,285 @@ mergeBy
...
+
+-------------------------------------------------------------------------------
+-- Remove the matching element from left and duplicates inside right stream
+-- > removeDupsAll compare (fromList [5]) (fromList [2,3,4,4,5])
+-- > [2,3,4]
+
+{-# INLINE_NORMAL removeDupsAll #-}
+removeDupsAll
+    ::(MonadIO m)
+    => (a -> a -> Ordering) -> Stream m a -> Stream m a -> Stream m a
```
harendra-kumar, line 651: What is the use case for this?

harendra-kumar (changes requested):
1. We have mutable arrays now, we can fix joinOuter by using mutable array instead of immtuable.
2. It seems there are many hlint changes in Nesting.hs. Please do not combine hlint or formatting changes with any functionality change PR. They should be done in a separate PR, otherwise it becomes a nightmare to review.

I have not yest reviewed Nesting.hs, will review that next.

harendra-kumar:
Can you also raise an issue about adding tests for these?

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
harendra-kumar: I was talking about `uniqBy`, not `nubBy`.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
```diff
@@ -380,6 +381,156 @@ o_1_space_concat value = sqrtVal `seq`
...
+    then S.drain $ Internal.intersectBySorted compare strm $ mkStreamLen val    
+    else S.drain $ Internal.intersectBySorted compare strm strm   
+
+o_n_heap_buffering :: Int -> [Benchmark]
+o_n_heap_buffering value =
+    [ bgroup "join"
+        [ -- XXX regular join for large no of data hangs for ever (30 min 
+          -- still not done)      
+         -- benchIOSink value "joinInner" (joinInner 0)
+         benchIOSink value "joinInner[Sqrt]" (joinInner sqrtVal)
```
harendra-kumar, line 504: See the concatMap benchmarks. Use `benchIOSrc1`.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
```diff
@@ -380,6 +381,156 @@ o_1_space_concat value = sqrtVal `seq`
...
+toKvMap :: Int -> (Int, Int)
+toKvMap p = (p, p)
+
+mkStreamLen :: (S.IsStream t, S.MonadAsync m) => Int -> t m Int
+mkStreamLen count = sourceUnfoldrM count 0
+
+{-# INLINE joinInner #-}
+joinInner :: (MonadIO m, S.MonadAsync m) => Int -> SerialT m Int -> m ()
+joinInner val strm =
+    if val > 0
```
harendra-kumar, line 396: Do not use a if condition here. Just generate a stream of the `val` size`. See `concatMap` benchmarks which do the same.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
```diff
@@ -380,6 +381,156 @@ o_1_space_concat value = sqrtVal `seq`
...
+intersectBySorted val strm = 
+    if val > 0
+    then S.drain $ Internal.intersectBySorted compare strm $ mkStreamLen val    
+    else S.drain $ Internal.intersectBySorted compare strm strm   
+
+o_n_heap_buffering :: Int -> [Benchmark]
+o_n_heap_buffering value =
+    [ bgroup "join"
+        [ -- XXX regular join for large no of data hangs for ever (30 min 
+          -- still not done)
```
harendra-kumar, line 502: We are using sqrt value, does it still hang? can we remove this comment.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
```diff
@@ -380,6 +381,156 @@ o_1_space_concat value = sqrtVal `seq`
...
+    if val > 0
+    then S.drain $ Internal.intersectBySorted compare strm $ mkStreamLen val    
+    else S.drain $ Internal.intersectBySorted compare strm strm   
+
+o_n_heap_buffering :: Int -> [Benchmark]
+o_n_heap_buffering value =
+    [ bgroup "join"
+        [ -- XXX regular join for large no of data hangs for ever (30 min 
+          -- still not done)      
+         -- benchIOSink value "joinInner" (joinInner 0)
```
harendra-kumar, line 503: Remove this comment and the other such comments below.

rnjtranjan:
re-implemented in PRs : 1415 to 1419

PR #1041, opened 2021-04-19
Document CI requirements

harendra-kumar:


MAINTAINING.md
```diff
@@ -33,6 +36,14 @@
     * Documents in the `docs` directory are consistent with new changes
     * All combinators have time and space complexity annotations
     * All combinators have `since` annotation
+    * Unreleased combinators should be marked with `pre-release` or
+      `internal` annotations
```
adithyaov, line 40:
```suggestion
    * Unreleased combinators should be marked with `Pre-release` or
      `Internal` annotations
```

adithyaov: All small also works.

MAINTAINING.md
```diff
@@ -33,6 +36,14 @@
...
+
+* _Build and Test_:
+
+    * All CIs are green
+    * Manually build and test for all flags (esp. `dev` flag) not covered by CIs
```
adithyaov, line 45: We should probably list the flags here (or somewhere else) for convinence.

adithyaov (approved):

MAINTAINING.md
harendra-kumar: They are all listed on the hackage index page and in the cabal file.

PR #1042, opened 2021-04-19
Enable CI for GHC 9.0.1

rnjtranjan:


benchmark/streamly-benchmarks.cabal
```diff
@@ -139,6 +139,7 @@ common bench-depends
 library
     import: lib-options, bench-depends
     hs-source-dirs: lib
+    mixins: tasty-bench (Test.Tasty.Bench as Criterion, Test.Tasty.Bench as Criterion.Main, Test.Tasty.Bench as Gauge, Test.Tasty.Bench as Gauge.Main)
```
harendra-kumar, line 142: You do not need to "as Criterion.*". We do not use those modules in the code.

harendra-kumar:
Some things to do:

* [ ] Fix compilation - replace parseCLIOpts with parseEnvOpts everywhere
* [ ] test and fix `bin/bench.sh --long`
* [ ] `bin/bench.sh`: for tasty-bench case use `--stdev Infinity` in `--quick` mode to finish the benchmarks faster
* [ ] `bin/bench.sh`: for gauge, fix this todo:   XXX We may have to use "sort | awk" to keep only benchmark names with
    shortest prefix e.g. "a/b/c" and "a/b", we should only keep "a/b"
    otherwise benchmarks will run multiple times.
* [ ] run `shellcheck` on the shell-scripts in `bin` and fix possible issues

rnjtranjan:
Currently, it doesn't support multi-threading. Here is the code snippet from tasty-bench which throws an error for Prelude.Rate/ benchmarks which runs asynchly/aheadly.


instance IsTest Benchmarkable where
  testOptions = pure
    [ Option (Proxy :: Proxy RelStDev)
    -- FailIfSlower and FailIfFaster must be options of a test provider rather
    -- than options of an ingredient to allow setting them on per-test level.
    , Option (Proxy :: Proxy FailIfSlower)
    , Option (Proxy :: Proxy FailIfFaster)
    ]
  run opts b = const $ case getNumThreads (lookupOption opts) of
    1 -> do
      est <- measureUntil True (lookupOption opts) (lookupOption opts) b
      pure $ testPassed $ show (Response est (lookupOption opts) (lookupOption opts))
    **_ -> pure $ testFailed "Benchmarks must not be run concurrently. Please pass --jobs 1 and/or avoid +RTS -N."**

rnjtranjan:
-l option doesn't produce benchmarks with a common prefix.
Here is the list produced for Tasty and Guage as well:

All.Prelude.Rate/o-1-space.asyncly.avgRate.20M
All.Prelude.Rate/o-1-space.asyncly.minRate.1M
All.Prelude.Rate/o-1-space.asyncly.minRate.10M
All.Prelude.Rate/o-1-space.asyncly.minRate.20M
All.Prelude.Rate/o-1-space.asyncly.maxRate.10M
All.Prelude.Rate/o-1-space.asyncly.constRate.1M
All.Prelude.Rate/o-1-space.asyncly.constRate.10M
All.Prelude.Rate/o-1-space.aheadly.avgRate/1M
All.Prelude.Rate/o-1-space.aheadly.minRate/1M
All.Prelude.Rate/o-1-space.aheadly.maxRate/1M
All.Prelude.Rate/o-1-space.aheadly.constRate/1M


Prelude.Serial/o-1-space/generation/IsList.fromList
Prelude.Serial/o-1-space/generation/IsString.fromString
Prelude.Serial/o-1-space/generation/absTimes
Prelude.Serial/o-1-space/generation/enumerate
Prelude.Serial/o-1-space/generation/enumerateFrom
Prelude.Serial/o-1-space/generation/enumerateFromThen
Prelude.Serial/o-1-space/generation/enumerateFromThenTo
Prelude.Serial/o-1-space/generation/enumerateFromTo
Prelude.Serial/o-1-space/generation/enumerateTo
Prelude.Serial/o-1-space/generation/fracFromThenTo
Prelude.Serial/o-1-space/generation/fracFromTo
Prelude.Serial/o-1-space/generation/fromFoldable
Prelude.Serial/o-1-space/generation/fromFoldableM
Prelude.Serial/o-1-space/generation/fromIndices
Prelude.Serial/o-1-space/generation/fromIndicesM
Prelude.Serial/o-1-space/generation/fromList
Prelude.Serial/o-1-space/generation/fromListM
Prelude.Serial/o-1-space/generation/intFromThenTo
Prelude.Serial/o-1-space/generation/intFromTo
Prelude.Serial/o-1-space/generation/integerFromStep
Prelude.Serial/o-1-space/generation/iterate
Prelude.Serial/o-1-space/generation/iterateM
Prelude.Serial/o-1-space/generation/repeat
Prelude.Serial/o-1-space/generation/repeatM
Prelude.Serial/o-1-space/generation/replicate
Prelude.Serial/o-1-space/generation/replicateM
Prelude.Serial/o-1-space/generation/unfoldr
Prelude.Serial/o-1-space/generation/unfoldrM

harendra-kumar:
> Prelude.Serial/o-1-space/generation/fromFoldable
> Prelude.Serial/o-1-space/generation/fromFoldableM

First one is a prefix of the second one.

harendra-kumar:
>  _ -> pure $ testFailed "Benchmarks must not be run concurrently. Please pass --jobs 1 and/or avoid +RTS -N."

We want tasty-bench to run the benchmarks one at a time. The above statement "Benchmarks must not be run concurrently" also makes sense. However, our each benchmark itself is multithreaded, so we need to the `-threaded +RTS -N` options to the executable. 

As the above error is suggesting, can we pass `--jobs 1` to restrict tasty-bench to one benchmark at a time but at the same time pass the GHC threading options for the single benchmark itself to have concurrency?

rnjtranjan:
> > _ -> pure $ testFailed "Benchmarks must not be run concurrently. Please pass --jobs 1 and/or avoid +RTS -N."
> 
> We want tasty-bench to run the benchmarks one at a time. The above statement "Benchmarks must not be run concurrently" also makes sense. However, our each benchmark itself is multithreaded, so we need to pass `-threaded +RTS -N` options to the executable.
> 
> As the above error is suggesting, can we pass `--jobs 1` to restrict tasty-bench to one benchmark at a time but at the same time pass the GHC threading options for the single benchmark itself to have concurrency?

It is working for '-j 1' (--jobs 1 is misleading)

default.nix
```diff
@@ -89,6 +93,7 @@ let haskellPackages =
           [ p.streamly
             p.streamly-benchmarks
             p.streamly-tests
+            p.tasty-bench
```
harendra-kumar, line 96: This file should not be committed as this depends on a local directory.

bin/bench-exec-one.sh
```diff
@@ -102,8 +102,8 @@ bench_rts_opts_specific () {
 
 if test "$USE_GAUGE" -eq 0
 then
-  SUPER_QUICK_OPTIONS=""
-  QUICKER_OPTIONS=""
+  SUPER_QUICK_OPTIONS="--stdev 100"
```
harendra-kumar, line 105: Why not use `Infinity` instead of 100?

benchmark/tasty-bench/tasty-bench.cabal
```diff
@@ -0,0 +1,47 @@
...
+    ghc-options:    -fcontext-stack=30
+
+  build-depends:
+    base >= 4.3 && < 5,
+    containers >= 0.4,
+    deepseq >= 1.1,
+    tasty >= 0.11.3
+  if impl(ghc < 7.8)
+    build-depends:
+      tagged >= 0.2
```
harendra-kumar, line 47: We cannot commit this file.

benchmark/Streamly/Benchmark/Data/Array.hs
```diff
@@ -184,8 +184,8 @@ defStreamSize = defaultStreamSize
 
 main :: IO ()
 main = do
-    (value, cfg, benches) <- parseCLIOpts defStreamSize
-    value `seq` runMode (mode cfg) cfg benches (allBenchmarks value)
+    value <- parseEnvOpts defaultStreamSize
```
harendra-kumar, line 187: Use `defStreamSize` instead as in the old code.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
```diff
@@ -160,12 +160,12 @@ inspect $ 'mergeByM `hasNoType` ''D.Step
...
-        , benchIOSrc1 "serial (2,2,x/4)" (serial4 (value `div` 4))
-        , benchIOSrc1 "append (2,2,x/4)" (append4 (value `div` 4))
-        , benchIOSrc1 "mergeBy (2,x/2)" (mergeBy (value `div` 2))
-        , benchIOSrc1 "mergeByM (2,x/2)" (mergeByM (value `div` 2))
+        [ benchIOSrc1 "serial (2 - x/2)" (serial2 (value `div` 2))
+        , benchIOSrc1 "append (2 - x/2)" (append2 (value `div` 2))
+        , benchIOSrc1 "serial (2 - 2 - x/4)" (serial4 (value `div` 4))
+        , benchIOSrc1 "append (2 - 2 - x/4)" (append4 (value `div` 4))
+        , benchIOSrc1 "mergeBy (2 - x/2)" (mergeBy (value `div` 2))
+        , benchIOSrc1 "mergeByM (2 - x/2)" (mergeByM (value `div` 2))
```
harendra-kumar, line 168: Why was this changed?

benchmark/lib/Streamly/Benchmark/Common.hs
```diff
@@ -134,76 +134,20 @@ defaultStreamSize = 100000
...
+parseEnvOpts defStreamSize = do
+    streamSize <- do
+        r <- lookupEnv "STREAM_SIZE"
+        case r of
+            Just x -> evaluate $ getStreamSize x
+            Nothing -> do
+                setEnv "STREAM_SIZE" (show defStreamSize)
+                return defStreamSize
+
+    streamSize `seq` return streamSize
```
harendra-kumar, line 153: Since tasty-bench supports custom CLI args let's revert this change and keep the original `parseCLIOpts` function for gauge, and implement another version of it for tasty-bench. We basically need to pass the stream size via the `--stream-size` CLI arg.

We will also have to change `bench.sh` so that we pass the stream size via CLI instead of passing it via env var.

benchmark/Streamly/Benchmark/Prelude/Serial/Nested.hs
rnjtranjan: comma "," is causing parsing issue tasty-bench.

harendra-kumar: We need to fix the script, if possible, rather than changing the benchmark names.

harendra-kumar: Benchmark names with "," are printed in double quotes e.g.

```
"All.Data.Fold/o-1-space.composition.tee (all, any)",75139802,5497826
```

bin/bench-exec-one.sh
rnjtranjan: for Infinity  It hangs forever. As per docs it is "Target relative standard deviation of measurements in percents" so max pct is 100.

harendra-kumar: Deviation can be more than 100%. Not sure why it hangs with `Infinity`. Did you try a very large value e.g. 10000? Did you use the latest tasty-bench from github? 

@bodigrim is this tested with `Infinity`?

Bodigrim: `tasty-bench` HEAD supports `--stdev Infinity`, but earlier (released) version does not handle it correctly and hang indeed. Sorry for that.

@harendra-kumar I can make a new release from `tasty-bench` HEAD, if desirable.

benchmark/lib/Streamly/Benchmark/Common.hs
rnjtranjan: Below code snippet in Readme for Custom command-line options is giving compilation error:
**customOpts & benchIngredients** in the below code snippet are not the same Types.

> main :: IO ()
> main = do
>   let customOpts  = [Option (Proxy :: Proxy RandomSeed)]
>       ingredients = includingOptions **customOpts : benchIngredients**
>   opts <- parseOptions ingredients benchmarks

Bodigrim: Snippet works for me fine; `customOpts` and `benchIngredients` are different types, but `includingOptions customOpts` and `benchIngredients` are the same.

You can check a complete example at https://github.com/haskell/vector/blob/a735abb688c870cf724bd620a0952e307252c18f/vector/benchmarks/Main.hs#L49

benchmark/Streamly/Benchmark/Data/Array.hs
```diff
@@ -184,8 +184,13 @@ defStreamSize = defaultStreamSize
 
 main :: IO ()
 main = do
+#if TASTY
+    (value, ingredients) <- parseCLIOpts defStreamSize $ bgroup "All" (allBenchmarks 0)
```
harendra-kumar, line 188: Damn! I did not realize this `All` prefix is being added by us, I assumed tasty adds it automatically. I had to do a lot of special handling in bench.sh because of this. We should not add this, why?

benchmark/Streamly/Benchmark/Data/Array/Stream/Foreign.hs
```diff
@@ -192,10 +192,14 @@ moduleName = "Data.Array.Stream.Foreign"
...
     env <- mkHandleBenchEnv
-    value `seq` runMode (mode cfg) cfg benches
-        (allBenchmarks env)
+#if TASTY
+    (value, ingredients) <- parseCLIOpts defaultStreamSize $ bgroup "All" (allBenchmarks env)    
+    value `seq` defaultMainWithIngredients ingredients $ bgroup "All" (allBenchmarks env)
+#else
+    (value, cfg, benches) <- parseCLIOpts defaultStreamSize
+    value `seq` runMode (mode cfg) cfg benches (allBenchmarks env)
+#endif
```
harendra-kumar, line 202: Define allBenchmarks as `allBenchmarks env _ =` and call `runWithCLIOpts (return . allBenchmarks env)`.

benchmark/Streamly/Benchmark/Data/Parser.hs
```diff
@@ -397,10 +397,16 @@ o_n_heap_serial value =
...
     env <- mkHandleBenchEnv
+#if TASTY
+    (value, ingredients) <- parseCLIOpts defaultStreamSize $ bgroup "All" (allBenchmarks env 0 [])
+    arrays <- IP.toList $ IP.arraysOf 100 $ sourceUnfoldrM value 0  
+    value `seq` defaultMainWithIngredients ingredients $ bgroup "All" (allBenchmarks env value arrays)
+#else
+    (value, cfg, benches) <- parseCLIOpts defaultStreamSize
     arrays <- IP.toList $ IP.arraysOf 100 $ sourceUnfoldrM value 0
     value `seq` runMode (mode cfg) cfg benches (allBenchmarks env value arrays)
+#endif
```
harendra-kumar, line 409: Define allBenchmarks as an IO action `allBenchmarks env value =` and call `runWithCLIOpts (allBenchmarks env)`. `arrays` can then be allocated inside allBenchmarks.

benchmark/Streamly/Benchmark/Prelude/Rate.hs
```diff
@@ -51,48 +52,48 @@ constRate value rate_ = S.constRate rate_ . sourceUnfoldrM value
...
+            [ -- benchIOSrc fromAsync "10K" $ maxRate value 10000
+            benchIOSrc fromAsync "10M" $ maxRate value 10000000
+            ]
+        , bgroup
+            "constRate"
+            [ -- benchIOSrc fromAsync "10K" $ constRate value 10000
+            benchIOSrc fromAsync "1M" $ constRate value 1000000
+            , benchIOSrc fromAsync "10M" $ constRate value 10000000
+            ]
+        ]
```
harendra-kumar, line 85: These are unrelated formatting changes, can be reverted or made in a separate commit.

benchmark/streamly-benchmarks.cabal
```diff
@@ -113,13 +118,24 @@ common bench-depends
...
     , random              >= 1.0   && < 2.0
-    , gauge               >= 0.2.4 && < 0.3
     , transformers        >= 0.4   && < 0.6
     , containers          >= 0.5   && < 0.7
     , typed-process     >= 0.2.3 && < 0.3
     , directory         >= 1.2.2 && < 1.4
     , ghc-prim          >= 0.4   && < 0.8
 
+  if flag(use-gauge)
+    cpp-options:    -DGUAGE
```
harendra-kumar, line 128: use `-DUSE_GAUGE` same as the cabal flag.

benchmark/streamly-benchmarks.cabal
```diff
@@ -113,13 +118,24 @@ common bench-depends
...
+    build-depends:  gauge >= 0.2.4 && < 0.3
+  else
+    cpp-options:    -DTASTY
```
harendra-kumar, line 131: We do not need `-DTASTY`, we can check for `#if !defined(USE_GAUGE)` or use the else case of `#ifdef USE_GAUGE`.

harendra-kumar (changes requested):

benchmark/streamly-benchmarks.cabal
Bodigrim: Actually it's enough to check that `MIN_VERSION_gauge` is defined instead of defining `GAUGE` manually.

benchmark/Streamly/Benchmark/Data/Array.hs
Bodigrim: One must group all benchmarks together, but the name is arbitrary, even an empty string will do.

rnjtranjan: "All" prefix is hard coded in **defaultMain**  method:
defaultMain :: [Benchmark] -> IO ()
defaultMain = Test.Tasty.defaultMainWithIngredients benchIngredients . testGroup "All"

benchmark/streamly-benchmarks.cabal
harendra-kumar: Yeah, that should do.

benchmark/Streamly/Benchmark/Data/Array.hs
harendra-kumar: We can use `defaultMainWithIngredients` instead of `defaultMain` and use the prefix as "" (empty string)?

rnjtranjan: Do we need to remove "All" prefix and fix bench.sh?

harendra-kumar: If we can do that it will be great.

benchmark/Streamly/Benchmark/Data/Array/Stream/Foreign.hs
```diff
@@ -188,18 +188,16 @@ o_1_space_copy_toChunks_group_ungroup env =
 -------------------------------------------------------------------------------
 
 moduleName :: String
-moduleName = "Data.Array.Stream.Foreign"
+moduleName = "Data.Array.Stream.Foreign"
```
harendra-kumar, line 191: Remove trailing space. You can setup your editor to automatically remove trailing spaces on save.

benchmark/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -6,6 +6,7 @@
 -- Maintainer  : streamly@composewell.com
 
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 9: Remove.

benchmark/Streamly/Benchmark/Prelude/Adaptive.hs
```diff
@@ -5,6 +5,8 @@
 -- License     : BSD3
 -- Maintainer  : streamly@composewell.com
 
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 8: Remove.

benchmark/Streamly/Benchmark/Prelude/Ahead.hs
```diff
@@ -5,6 +5,8 @@
 -- License     : BSD3
 -- Maintainer  : streamly@composewell.com
 
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 8: Remove.

benchmark/Streamly/Benchmark/Prelude/Async.hs
```diff
@@ -5,6 +5,8 @@
 -- License     : BSD3
 -- Maintainer  : streamly@composewell.com
 
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 8: Remove.

benchmark/Streamly/Benchmark/Prelude/Concurrent.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 2: Remove.

benchmark/Streamly/Benchmark/Prelude/Parallel.hs
```diff
@@ -6,6 +6,7 @@
 -- Maintainer  : streamly@composewell.com
 
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE CPP #-}
```
harendra-kumar, line 9: Remove.

benchmark/lib/Streamly/Benchmark/Common.hs
```diff
@@ -3,6 +3,7 @@
 {-# OPTIONS_GHC -Wno-orphans #-}
 #endif
 
+
```
harendra-kumar, line 6: remove blank line.

benchmark/lib/Streamly/Benchmark/Common.hs
```diff
@@ -134,18 +153,43 @@ defaultStreamSize = 100000
...
-newtype BenchOpts = StreamSize Int deriving Show
-
 getStreamSize :: String -> Int
 getStreamSize size =
     case (readMaybe size :: Maybe Int) of
         Just x -> x
         Nothing -> error "Stream size must be numeric"
+{--
+parseEnvOpts :: Int -> IO Int
+parseEnvOpts defStreamSize = do
```
harendra-kumar, line 163: We can remove this.

harendra-kumar (changes requested):
There are some more simplifications to be done, I will do those on top once you are done with these.

harendra-kumar:
@rnjtranjan I have made some changes to simplify the implementation and make bench.sh more friendly with lesser quirks. Almost all of the changes from your last commit are addressed in my local workspace while I was reviewing it. So I will just force push.

harendra-kumar:
The prefix match stuff for gauge does not seem to be working.

harendra-kumar:
Fixed it.

harendra-kumar:
I compared benchmarks using gauge before and after this change to see if anything got impacted. 

In case of gauge, splitOn benchmarks - where we needed `perRunEnv` but changed to closing/opening file handles along with the benchmarked action - do not show much change in cpuTime and allocations, but invariably show a significant increase in `bytesCopied` (20-100%). Though it does not look like a big deal, it would be nice to have a purer picture of the stats.

One particular Data.List benchmark showed changes that are somewhat unusual. 

bytesCopied

```
Prelude.Serial/o-n-heap/buffered/showPrec Haskell lists         5760.00                 +166885.00%

```

Allocations:

```
Prelude.Serial/o-n-heap/buffered/showPrec Haskell lists          28142232.00                     +28.96%

```

cpuTime

```
Prelude.Serial/o-n-heap/buffered/showPrec Haskell lists          9652.99                    +118.67%

```

harendra-kumar:
## gauge vs tasty-bench

Comparing the gauge results (minimal iteration tests) with tasty-bench results (--stdev 100000) on the same commit.

## Allocations reporting issue

In many cases where `gauge` shows 0 allocations, tasty-bench shows non-zero allocations. For example, see the `Data.Fold` benchmarks. In most of these benchmarks we know that allocations should be zero. So gauge is probably more accurate, `tasty-bench` may be measuring some extraneous stuff outside the actual benchmark? Does it need to force a GC before measurements or some other such thing to make the results more accurate?

@Bodigrim in case you want to investigate. The benchmarks can be run using the following commands (use ghc 8.8.x so that bench-show can be built):

```
$ cabal run bench:Data.Fold
$ cabal run bench:Data.Fold --flag use-gauge
```

To use the exact script with bench-show (may take long time building for the first time) reporting, but it uses the exact options that I used and shows results in comparable format:

```
$ bin/bench.sh --quick --sort-by-name --targets Data.Fold # for tasty-bench
$ bin/bench.sh --quick --sort-by-name --targets Data.Fold --use-gauge # for gauge
```

I thought one reason could be that tasty-bench is using many more iterations than gauge so its accumulating some garbage while gauge is not.  We are using a large value for stdev to reduce but not sure how much it reduces. But this does not seem to be the as gauge reports same results (zero allocations) even when many samples are taken and many iterations are run (using the --slow in bench.sh uses the default settings of gauge):

```
$ bin/bench.sh --slow --use-gauge --targets Data.Fold

Data.Fold(allocated)
Benchmark                                                          default(Bytes)
------------------------------------------------------------------ --------------
Data.Fold/o-n-space/sequence_/100                                     50413952.00
Data.Fold/o-1-space/composition/demuxWith [sum, length]               15151696.00
Data.Fold/o-1-space/composition/demuxDefaultWith [sum, length] sum    15151691.64
Data.Fold/o-1-space/composition/classifyWith sum                      14346701.60
Data.Fold/o-n-heap/elimination/toStream                                8704113.33
Data.Fold/o-n-heap/elimination/toList                                  6341550.29
Data.Fold/o-n-heap/elimination/toStreamRev                             6341538.29
Data.Fold/o-n-heap/elimination/toListRev                               3945781.33
Data.Fold/o-1-space/transformation/rsequence                                 0.00
Data.Fold/o-1-space/transformation/rmapM                                     0.00
Data.Fold/o-1-space/transformation/pipe-mapM                                 0.00
Data.Fold/o-1-space/transformation/mapMaybe                                  0.00
Data.Fold/o-1-space/transformation/map                                       0.00
Data.Fold/o-1-space/elimination/variance                                     0.00
Data.Fold/o-1-space/elimination/takeEndBy_                                   0.00
Data.Fold/o-1-space/elimination/take                                         0.00
Data.Fold/o-1-space/elimination/sum (foldMap)                                0.00
Data.Fold/o-1-space/elimination/sum                                          0.00
Data.Fold/o-1-space/elimination/stdDev                                       0.00
Data.Fold/o-1-space/elimination/sink                                         0.00
Data.Fold/o-1-space/elimination/product                                      0.00
Data.Fold/o-1-space/elimination/or                                           0.00
Data.Fold/o-1-space/elimination/null                                         0.00
Data.Fold/o-1-space/elimination/notElem                                      0.00
Data.Fold/o-1-space/elimination/minimumBy                                    0.00
Data.Fold/o-1-space/elimination/minimum                                      0.00
Data.Fold/o-1-space/elimination/mean                                         0.00
Data.Fold/o-1-space/elimination/mconcat                                      0.00
Data.Fold/o-1-space/elimination/maximumBy                                    0.00
Data.Fold/o-1-space/elimination/maximum                                      0.00
Data.Fold/o-1-space/elimination/lookup                                       0.00
Data.Fold/o-1-space/elimination/length                                       0.00
Data.Fold/o-1-space/elimination/last                                         0.00
Data.Fold/o-1-space/elimination/index                                        0.00
Data.Fold/o-1-space/elimination/head                                         0.00
Data.Fold/o-1-space/elimination/foldMapM                                     0.00
Data.Fold/o-1-space/elimination/foldMap                                      0.00
Data.Fold/o-1-space/elimination/findIndex                                    0.00
Data.Fold/o-1-space/elimination/find                                         0.00
Data.Fold/o-1-space/elimination/elemIndex                                    0.00
Data.Fold/o-1-space/elimination/elem                                         0.00
Data.Fold/o-1-space/elimination/drainN                                       0.00
Data.Fold/o-1-space/elimination/drainBy                                      0.00
Data.Fold/o-1-space/elimination/drain                                        0.00
Data.Fold/o-1-space/elimination/any                                          0.00
Data.Fold/o-1-space/elimination/and                                          0.00
Data.Fold/o-1-space/elimination/all                                          0.00
Data.Fold/o-1-space/composition/unzip (sum, length)                          0.00
Data.Fold/o-1-space/composition/tee (sum, length)                            0.00
Data.Fold/o-1-space/composition/tee (all, any)                               0.00
Data.Fold/o-1-space/composition/serialWith (all, any)                        0.00
Data.Fold/o-1-space/composition/partition (sum, length)                      0.00
Data.Fold/o-1-space/composition/many drain (take 1)                          0.00
Data.Fold/o-1-space/composition/distribute [sum, length]                     0.00
```

In array show benchmarks e.g. `Data.Array/o-1-space/generation/show` and the same for other array types, the results for tasty-bench are different from that of gauge. In fact, I had to increase the GHC heap limit from 16MB to 32 MB to make it succeed.

Another case of wildly different allocations is in `Data.Array.Prim/o-1-space/generation/fromList . intFromTo `, gauge shows 0 whereas tasty-bench shows allocations in MB.

## Quick Benchmarking

Even though we are using --stdev 1000000, tasty-bench still seems to be slower. In some other benchmark suites it seems even more slow.

```
$ time bin/bench.sh --use-gauge --quick --targets Prelude.Serial --raw
real    0m24.195s
user    0m19.636s
sys     0m4.622s

$ time bin/bench.sh --quick --targets Prelude.Serial --raw
real    0m42.999s
user    0m37.597s
sys     0m7.741s
```

## Reports

For data on all benchmarks see these reports:

[charts.gauge.txt](https://github.com/composewell/streamly/files/6609742/charts.gauge.txt)
[charts.tasty-bench.txt](https://github.com/composewell/streamly/files/6609750/charts.tasty-bench.txt)

Bodigrim:
> In many cases where `gauge` shows 0 allocations, `tasty-bench` shows non-zero allocations. For example, see the `Data.Fold` benchmarks. 

I see that for many `Data.Fold` benchmarks `tasty-bench` reports 50-60 B allocated. Is it what you are refering to? I believe this is due to measuring `randomRIO (1,1)` in 
https://github.com/composewell/streamly/blob/56e1807742553c2d3a48c3c5135c8a6bd2238100/benchmark/Streamly/Benchmark/Data/Fold.hs#L52
If I replace it with equivalent `f $ source value 1`, spurious allocations disappear. What's the rationale behind `randomRIO (1,1)` instead of `pure 1`?

Bodigrim:
> Does it need to force a GC before measurements or some other such thing to make the results more accurate?

I don't think so, `tasty-bench` [runs](https://github.com/Bodigrim/tasty-bench/blob/e63b5bf1beed1cba3320e6f1698f12972d696dd3/Test/Tasty/Bench.hs#L836) [`performGC`](https://hackage.haskell.org/package/base-4.14.0.0/docs/System-Mem.html#v:performGC) before each measurement. 

> (use ghc 8.8.x so that `bench-show` can be built)

How to build it on macOS? It complains that "<command line>: can't load framework: Cocoa (not found)".

> Even though we are using `--stdev 1000000`, `tasty-bench` still seems to be slower. 

This is fixed in `tasty-bench` HEAD with `--stdev Infinity`, which is 3x faster than `--stdev 1000000`.

> Another case of wildly different allocations is in `Data.Array.Prim/o-1-space/generation/fromList . intFromTo`, `gauge` shows 0 whereas `tasty-bench` shows allocations in MB.

Interesting. I am afraid, I do not know much about `streamly` to investigate it. But basically I doubt that these allocations are fake. It's more likely that something did not inline?.. While retaining API compatibility, `tasty-bench` internals are implemented differently.

harendra-kumar:
> How to build it on macOS? It complains that ": can't load framework: Cocoa (not found)".

You need to have Xcode installed. See https://streamly.composewell.com/streamly-0.8.0/Compiling.html#mac-osx . Let me know if you are still not able to compile or any additional steps that you took to compile so that we can update the guide.

harendra-kumar:
> I see that for many Data.Fold benchmarks tasty-bench reports 50-60 B allocated. Is it what you are refering to? I believe this is due to measuring randomRIO (1,1) in

In my runs there were allocations in KBs, see the attached files above. Need to recheck if that is accurate or there is some issue in processing the data. Yes, `randomRIO` is to stop memoization, it predates the `nfAppIO` API. I also tried to replace it with `nfAppIO` at some point but there were issues (I do not remember now) so I let it be the way it is.

harendra-kumar:
@Bodigrim It seems when I use `--csv` option it shows more allocations and bytesCopied:

```
$ cabal run bench:Data.Fold -- --stdev 1000000 +RTS -T -K36K -M36M -RTS -p /elimination.product/ --csv=fold.csv
Up to date
All
  Data.Fold/o-1-space
    elimination
      product: OK
         70 μs ± 112 μs,  13 KB allocated, 5.9 KB copied

All 1 tests passed (0.01s)
$ cabal run bench:Data.Fold -- --stdev 1000000 +RTS -T -K36K -M36M -RTS -p /elimination.product/
Up to date
All
  Data.Fold/o-1-space
    elimination
      product: OK
         58 μs ±  88 μs,   0 B  allocated,   0 B  copied

All 1 tests passed (0.01s)
```

Bodigrim:
> Yes, `randomRIO` is to stop memoization

I'm not sure it achieves this goal then. `randomRIO` is agressively inlined and has a special case for `high == low`. I expect that it is equivalent to `pure 1`, and GHC happily pushes it through `>==`.

> It seems when I use `--csv` option it shows more allocations and bytesCopied:

Ah, interesting, thanks. I'll try to pin point a leak (unfortunately, it's all `String`s).

harendra-kumar:
> This is fixed in tasty-bench HEAD with --stdev Infinity, which is 3x faster than --stdev 1000000.
Great!

harendra-kumar:
> I'm not sure it achieves this goal then. randomRIO is agressively inlined and has a special case for high == low. I expect that it is equivalent to pure 1, and GHC happily pushes it through >==.

You may be right. Earlier I used a range, but later I changed it to a constant assuming randomRIO is not special casing it. I did not check the core. But as long as our benchmark results are reasonable its fine (I observed that when there is such an issue the timings are so low that we get to know), that means we may not even need it. We can check removing it and comparing.

harendra-kumar:
Results seem reasonable. Some issues will be taken care of in the next release of `tasty-bench`. There is a `--csv` issue which I am sure @Bodigrim will take care of. So I am going to merge this now.

PR #1048, opened 2021-04-23
Expose the Fold.Tee module for general use

harendra-kumar:


src/Streamly/Data/Fold/Tee.hs
```diff
@@ -0,0 +1,40 @@
...
+-- Copyright   : (c) 2021 Composewell Technologies
+-- License     : BSD-3-Clause
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+-- The 'Tee' type is a newtype wrapper over the 'Streamly.Data.Fold.Fold' type providing
+-- distributive 'Applicative', 'Semigroup', 'Monoid', 'Num', 'Floating' and
+-- 'Fractional' instances. The input received by the composed 'Tee' is
+-- replicated and distributed to both the constituent Tee's.
```
adithyaov, line 12:
```suggestion
-- replicated and distributed to both the constituent Tees.
```

Or `Tee-s`?

adithyaov (approved):
We also update the changelog for new features right?

src/Streamly/Data/Fold/Tee.hs
harendra-kumar: Will change.

harendra-kumar:
> We also update the changelog for new features right?

Added to changelog.

PR #1049, opened 2021-04-23
New exports

harendra-kumar:
* Exports some Internal APIs.
* Updates some docs
* Change signature of intercalate

Can review commit by commit.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
@@ -548,27 +548,22 @@ gintercalate unf1 str1 unf2 str2 =
         unf1 (D.toStreamD str1)
         unf2 (D.toStreamD str2)
 
--- XXX The order of arguments in "intercalate" is consistent with the list
--- intercalate but inconsistent with gintercalate and other stream interleaving
--- combinators. We can change the order of the arguments in other combinators
--- but then 'interleave' combinator may become a bit unintuitive because we
--- will be starting with the second stream.
-
--- > intercalate seed unf str = gintercalate unf str unf (repeatM seed)
--- > intercalate a unf str = unfoldMany unf $ intersperse a str
+-- > intercalate unf seed str = gintercalate unf str unf (repeatM seed)
 --
 -- | 'intersperse' followed by unfold and concat.
 --
--- > unwords = intercalate " " Unfold.fromList
+-- > intercalate unf a str = unfoldMany unf $ intersperse a str
+-- > intersperse = intercalate (Unfold.function id)
+-- > unwords = intercalate Unfold.fromList " "
 --
--- >>> Stream.toList $ Stream.intercalate " " Unfold.fromList $ Stream.fromList ["abc", "def", "ghi"]
+-- >>> Stream.toList $ Stream.intercalate Unfold.fromList " " $ Stream.fromList ["abc", "def", "ghi"]
 -- "abc def ghi"
 --
--- /Pre-release/
+-- @since 0.8.0
 {-# INLINE intercalate #-}
 intercalate :: (IsStream t, Monad m)
-    => b -> Unfold m b c -> t m b -> t m c
-intercalate seed unf str = D.fromStreamD $
+    => Unfold m b c -> b -> t m b -> t m c
+intercalate unf seed str = D.fromStreamD $
     D.unfoldMany unf $ D.intersperse seed (toStreamD str)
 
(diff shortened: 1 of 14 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -548,27 +548,22 @@ gintercalate unf1 str1 unf2 str2 =
...
 
--- XXX The order of arguments in "intercalate" is consistent with the list
--- intercalate but inconsistent with gintercalate and other stream interleaving
--- combinators. We can change the order of the arguments in other combinators
--- but then 'interleave' combinator may become a bit unintuitive because we
--- will be starting with the second stream.
-
--- > intercalate seed unf str = gintercalate unf str unf (repeatM seed)
--- > intercalate a unf str = unfoldMany unf $ intersperse a str
+-- > intercalate unf seed str = gintercalate unf str unf (repeatM seed)
```
adithyaov, line 551: Have this in the haddock block?

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -583,21 +578,22 @@ gintercalateSuffix unf1 str1 unf2 str2 =
         unf1 (D.toStreamD str1)
         unf2 (D.toStreamD str2)
 
--- > intercalateSuffix seed unf str = gintercalateSuffix unf str unf (repeatM seed)
--- > intercalateSuffix a unf str = unfoldMany unf $ intersperseSuffix a str
+-- > intercalateSuffix unf seed str = gintercalateSuffix unf str unf (repeatM seed)
```
adithyaov, line 581: In haddock block?

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
harendra-kumar: `gintercalate` is not exposed.

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
harendra-kumar: `gintercalateSuffix` is not exposed.

src/Streamly/Internal/Network/Socket.hs
```diff
@@ -106,25 +105,27 @@ import qualified Streamly.Internal.Data.Array.Foreign.Type as A
...
+-- by 'forSocketM' rather than any exception raised by 'action'.
 --
--- @since 0.7.0
-{-# INLINE handleWithM #-}
-handleWithM :: (MonadMask m, MonadIO m) => (Socket -> m ()) -> Socket -> m ()
-handleWithM f sk = finally (f sk) (liftIO (Net.close sk))
+-- @since 0.8.0
+{-# INLINE forSocketM #-}
+forSocketM :: (MonadMask m, MonadIO m) => (Socket -> m ()) -> Socket -> m ()
+forSocketM f sk = finally (f sk) (liftIO (Net.close sk))
```
adithyaov, line 117: `forSocketM` sounds like a loop.

src/Streamly/Internal/Network/Socket.hs
```diff
@@ -106,25 +105,27 @@ import qualified Streamly.Internal.Data.Array.Foreign.Type as A
...
 -- computation.
 --
--- @since 0.7.0
-{-# INLINE handleWith #-}
-handleWith :: (IsStream t, MonadAsync m, MonadCatch m)
+-- /Inhibits stream fusion/
+--
+-- /Internal/
+{-# INLINE withSocket #-}
+withSocket :: (IsStream t, MonadAsync m, MonadCatch m)
```
adithyaov, line 126: This is still internal so it does not matter but it would be nice if the signature structure is similar to `forSocketM`.

adithyaov (approved):

src/Streamly/Internal/Network/Socket.hs
harendra-kumar: The argument order is according to the names. `with` is like `withFile` and `for` is opposite of that. The signature of forSocketM is to avoid using `flip` for common use cases. See streamly-examples for a use case. If `withSocket` is also used like that then we can have a `forSocket` with changed argument order.

src/Streamly/Internal/Network/Socket.hs
harendra-kumar: `for` is the opposite of `with`, the confusion may be because we are used to `for` as a loop. Other alternative is `useSocketM`, any other better/short alternatives?

PR #1050, opened 2021-04-24
Console.Stdio module for standard input/output ops

harendra-kumar:


design/container-api.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Key Access (Maps)
+
+* getKey
+* findKey (test existence)
+* createKey (insert new key for mutable maps)
+* putKey (insert or update for mutable maps)
+* updateKey (update existing or fail)
+* deleteKey  (delete existing or fail)
+* destroyKey (delete existing or not)
```
rnjtranjan, line 54: we can use "tryDeleteKey" instead of destroyKey.

rnjtranjan (approved):

design/container-api.md
harendra-kumar: destroy is for unconditional delete. Anyway we can work out finer details when we actually use these.

Issue #659, opened 2020-08-23
Factor out common code in encodeUtf8 and encodeUtf8Lax


harendra-kumar:
Both of these routines do the same thing, the only difference is how they handle invalid characters in the stream. I propose factoring out the encoding without any invalid char handling in an API called `encodeUtf8Unsafe` and use the following routines on the input stream before using `encodeUtf8Unsafe`:

* `replaceInvalid` to replace invalid chars in the input stream
* `dropInvalid` to drop invalid chars
* `errorOnInvalid` to error out on invalid chars

PR #1052, opened 2021-04-24
Refactor encodeUtf8 and encodeUtf8'

rnjtranjan:
Fix for #659


harendra-kumar (approved):
Looks good, though I have a change implementing encoding using unfolds which will supersede this. But we can still commit it in case we need to go back to old implementation for some reason.

harendra-kumar (changes requested):
encodeUtf8D_ should also be implemented using encodeUtf8DGeneric .

harendra-kumar (approved):

harendra-kumar:
I am going to squash and merge.

PR #1055, opened 2021-04-26
Use unfold to implement utf8 encode

harendra-kumar:
This makes it more modular and the unfold is reusable. The benchmarks do not show significant degradation. The encode/decode utf8 benchmark is down by 3% and encode/decodeArray down by 10%.

```
Unicode.Stream(cpuTime)
Benchmark                                                                                        default(0)(ms) default(1) - default(0)(%)
------------------------------------------------------------------------------------------------ -------------- --------------------------
Unicode.Stream/o-1-space/decode-encode/toChunks/encodeUtf8' . decodeUtf8Arrays (1/10)                    124.69                      +9.26
Unicode.Stream/o-1-space/decode-encode/encodeUtf8 . decodeUtf8 (1/10)                                    136.58                      +2.49

Unicode.Stream(allocated)
Benchmark                                                                                        default(0)(MiB) default(1) - default(0)(%)
------------------------------------------------------------------------------------------------ --------------- --------------------------
Unicode.Stream/o-1-space/decode-encode/encodeUtf8 . decodeUtf8 (1/10)                                      28.33                       0.00
Unicode.Stream/o-1-space/decode-encode/toChunks/encodeUtf8' . decodeUtf8Arrays (1/10)                      26.31                       0.00

Unicode.Stream(bytesCopied)
Benchmark                                                                                        default(0)(KiB) default(1) - default(0)(%)
------------------------------------------------------------------------------------------------ --------------- --------------------------
Unicode.Stream/o-1-space/decode-encode/toChunks/encodeUtf8' . decodeUtf8Arrays (1/10)                       6.02                     +18.70
Unicode.Stream/o-1-space/decode-encode/encodeUtf8 . decodeUtf8 (1/10)                                       8.38                       0.00
```

### DIFF
--- a/src/Streamly/Internal/Unicode/Stream.hs
+++ b/src/Streamly/Internal/Unicode/Stream.hs
@@ -38,6 +38,7 @@ module Streamly.Internal.Unicode.Stream
     , encodeLatin1_
 
     -- ** UTF-8 Encoding
+    , readCharUtf8'
     , encodeUtf8
     , encodeUtf8'
     , encodeUtf8_
@@ -96,7 +97,7 @@ import Streamly.Internal.Data.Stream.Serial (SerialT)
 import Streamly.Internal.Data.Stream.StreamD (Stream(..), Step (..))
 import Streamly.Internal.Data.SVar (adaptState)
 import Streamly.Internal.Data.Tuple.Strict (Tuple'(..))
-import Streamly.Internal.Data.Unfold (Unfold)
+import Streamly.Internal.Data.Unfold.Type (Unfold(..))
 import Streamly.Internal.System.IO (unsafeInlineIO)
 
 import qualified Streamly.Internal.Data.Unfold as Unfold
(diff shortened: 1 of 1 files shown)


rnjtranjan (approved):

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -772,79 +774,36 @@ ord4 c = assert (n >= 0x10000)  (WCons x1 (WCons x2 (WCons x3 (WCons x4 WNil))))
...
-                                        if isSurrogate c
-                                        then replaceInvalid s
-                                        else Skip (EncodeState s (ord3 c))
-
-                            | otherwise -> Skip (EncodeState s (ord4 c))
-                Skip s -> Skip (EncodeState s WNil)
-                Stop -> Stop
-    step' _ (EncodeState s (WCons x xs)) = return $ Yield x (EncodeState s xs)
+{-# INLINE_NORMAL readCharUtf8' #-}
+readCharUtf8' :: Monad m => Unfold m Char Word8
```
harendra-kumar, line 795: This will be moved to the `Char` module. We can perhaps rename this to just `readUtf8'` or just `read'`.  Need to think about consistency between folds/parsers and unfold naming. The opposite operation to parse a Char from Word8 stream would be `write` or `writeUtf8` or `parse`?

Or would `unfold`, `fold`, `parse` be better names than `read`, `write` etc. That would require a lot of changes to the existing modules though, but we can do it if its worth it and more intuitive. Also, in that case we may have to change the `unfold`, `fold` in stream to something like `runUnfold`, `runFold`.

PR #912, opened 2021-02-06
Add parseArray for parsing stream of arrays

adithyaov:
- [x] ~Add a `MonadThrow` constraint~
- [x] Add tests for `parseArray`
- [x] Add benchmarks for `parseArray`

We need to compare `parseArray` with `parselMx'`

PR #1056, opened 2021-04-26
Array stream fold and parse routines

harendra-kumar:


src/Streamly/Internal/Data/Array/Stream/Foreign.hs
```diff
@@ -457,18 +525,23 @@ _toArraysOf n = FL.chunksOf n (A.writeNF n) FL.toStream
...
+-- This can be generalized to any type provided it can be unfolded to a stream
+-- and it can be combined using a semigroup operation.
+--
+-- XXX This should be written using CPS (as parseK) if we want it to scale wrt
+-- to the number of times it can be called on the same stream.
 {-# INLINE_NORMAL parseD #-}
 parseD ::
        forall m a b. (MonadIO m, MonadThrow m, Storable a)
-    => PRD.Parser m (Array a) b
+    => PRD.Parser m a b
```
adithyaov, line 544: Can you compare the performance of this with `parseArray` in https://github.com/composewell/streamly/pull/912/

adithyaov: I'm guessing the one in #912 would be more efficient if there is a lot of backtracking.

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
```diff
@@ -480,6 +553,105 @@ parseD (PRD.Parser pstep initial extract) stream@(D.Stream step state) = do
...
+        case pRes of
+            PR.Partial 0 pst1 ->
+                 gobuf SPEC s (List []) fp next pst1
+            PR.Partial n pst1 -> do
+                assert (n <= Prelude.length (x:getList backBuf)) (return ())
+                let src0 = Prelude.take n (x:getList backBuf)
+                    arr0 = A.fromListN n (Prelude.reverse src0)
+                    !(Ptr curAddr) = next
+                    arr1 = Array (ForeignPtr curAddr contents) (Ptr end)
+                    src = arr0 <> arr1
```
adithyaov, line 592: You could probably save the arrays in the `backBuf` instead.

adithyaov: Never mind, this looks good. Need to think about extensive backtracking where you need to append arrays multiple times triggering a lot of `realloc`

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
```diff
@@ -457,18 +525,23 @@ _toArraysOf n = FL.chunksOf n (A.writeNF n) FL.toStream
...
 -- GHC parser does not accept {-# ANN type [] NoSpecConstr #-}, so we need
 -- to make a newtype.
 {-# ANN type List NoSpecConstr #-}
 newtype List a = List {getList :: [a]}
 
+-- This can be generalized to any type provided it can be unfolded to a stream
+-- and it can be combined using a semigroup operation.
+--
+-- XXX This should be written using CPS (as parseK) if we want it to scale wrt
+-- to the number of times it can be called on the same stream.
```
adithyaov, line 540: Can you open an issue and assign it to me?

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
```diff
@@ -480,6 +553,105 @@ parseD (PRD.Parser pstep initial extract) stream@(D.Stream step state) = do
...
+-------------------------------------------------------------------------------
+-- Elimination - Running Array Folds and parsers
+-------------------------------------------------------------------------------
+
+{-# INLINE_NORMAL parseArrD #-}
+parseArrD ::
+       forall m a b. (MonadIO m, MonadThrow m, Storable a)
+    => PRD.Parser m (Array a) b
+    -> D.Stream m (Array.Array a)
+    -> m (b, D.Stream m (Array.Array a))
```
adithyaov, line 645: Please remind me why we have a special parser for parsing arrays?
Shouldn't we backtrack arrays instead of backtracking elements? Did this have any use-case?

adithyaov: Looks like we have similar folds as well.

adithyaov: It's pre-existing code so guess there were some practical applications.

adithyaov: Ok, nvm.

adithyaov (approved):

adithyaov:
Looks good. Few ignorable comments.

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
harendra-kumar: > Can you compare the performance of this with `parseArray` in #912

I will let you do that. If you can create a benchmark that proves that the DLL implementation is significantly better we will replace it with that.

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
harendra-kumar: We did this whole thing for streamly-lz4 and tar. We are only using foldArr_ in streamly-lz4. I don't remember if tar required parseArr as well.

src/Streamly/Internal/Data/Array/Stream/Foreign.hs
harendra-kumar: ok.

harendra-kumar:
I am going to merge this and raise issues for any future work.

PR #1060, opened 2021-04-27
Update README for 0.8.0

harendra-kumar:
* Updated the README to have mostly example based illustration.
* Moved the unused parts of the old README to docs/Overview.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,31 @@ Let's go through some practical examples to see it working. You
 can find the working code of these examples in the [streamly-examples
 repository](https://github.com/composewell/streamly-examples).
 
+## Installing and using
+
+Please see [INSTALL.md](./INSTALL.md) for instructions on how to use streamly
+with your Haskell build tool or package manager.
+
+## Documentation
+
+Streamly comes with batteries included, see [the streamly documentation
+portal](https://streamly.composewell.com) for available modules. Modules
+are divided in two categories:
+
+* Released Modules: these are modules that have a stable API, any API changes
+  conform to a versioning policy.
+* Pre-release Modules: These modules are not yet released due to some planned
+  changes in near future, they will be released soon.
+
+Pre-release APIs:  Some of the APIs that are recently introduced
+and require some soak time for stability are kept in the
+internal modules corresponding to the released module (e.g.
+Streamly.Internal.Data.Fold).  We usually try to change even the
+unstable APIs in a major release version.
+
 ## Types Overview
+
+Quick overview of basic types for this tutorial:
 
 * `SerialT IO a` is a serial stream of values of type `a` in IO Monad.
 * `AsyncT IO a` is a concurrent (async) stream of values of type `a` in IO
@@ -134,7 +158,7 @@ excellent modularity.  Experienced Haskellers would notice that we have
 not used bytestrings, we simply use a stream of `Word8`, simplifying the
 program.
 
-## Performance
+## Word Counting Performance
 
 We compare two equivalent implementations, one using Haskell Streamly and the
 other using C. The
@@ -504,26 +528,6 @@ via a [compiler plugin](https://github.com/composewell/fusion-plugin).
 We hope to bring these optimizations to GHC in future but until
 then we recommend that you use the plugin for performance sensitive
 applications.
-
-## Installing and using
-
-Please see [INSTALL.md](./INSTALL.md) for instructions on how to use streamly
-with your Haskell build tool or package manager.
-
-Streamly comes with batteries included, see [the
-documentation](https://streamly.composewell.com) for available modules. Modules
-are divided in two categories:
-
-* Released Modules: these are modules that have a stable API, any API changes
-  conform to a versioning policy.
-* Pre-release APIs:  Some of the APIs that are recently introduced and
-  require some soak time for stability are kept in the
-  internal modules corresponding to the released module (e.g.
-  Streamly.Internal.Data.Fold).
-* Pre-release Modules: These modules are not yet released due to some planned
-  changes in near future, they will be released soon.
-
-We usually try to change even the unstable APIs in a major release version.
 
 ## Support
 
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@
 ## Idiomatic Haskell at the speed of C
 
 Streamly is a Haskell library/framework providing basic building blocks
-or combinators to perform fundamental as well as advanced programming
-tasks with ease. The key features it provides are:
+or combinators to build safe, scalable, modular and high performance
+software systems.  The key features it provides are:
 
 * Speed of C
 * Safety of Haskell
@@ -43,10 +43,10 @@ repository](https://github.com/composewell/streamly-examples).
 
 ## Modular Word Counting
 
-The `Fold` data type in streamly represents a consumer of stream. In
-this example, we will use individual folds to count bytes, words and
-lines in a file. We will see how these individual folds can be composed
-together to do all the three at once with the same performance.
+In this example, we will use folds to count bytes, words and lines in
+a file.  A `Fold` is a composable stream consumer.  We will see how
+individual folds can be composed together to perform all the three
+counts at once with the same performance.
 
 See [WordCountModular.hs](https://github.com/composewell/streamly-examples/blob/master/examples/WordCountModular.hs)
 for full working code including imports that we may have omitted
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -11,572 +11,528 @@
...
+* `Fold IO a b` is a representation of a function that converts a stream of
+  type `a` to a final accumulator of type `b` in IO Monad.
 
-## Streaming Pipelines
+## Modular Word Counting
 
-The following snippet provides a simple stream composition example that reads
-numbers from stdin, prints the squares of even numbers and exits if an even
-number more than 9 is entered.
+The `Fold` data type in streamly represents a consumer of stream. In
```
adithyaov, line 46: Is this the correct place to put info about the `Fold` data type?

README.md
```diff
@@ -11,572 +11,528 @@
...
+countAll = Tee.toFold $ (,,) <$> Tee Fold.length <*> Tee nlines <*> Tee nwords
+
+wc :: String -> IO (Int, Int, Int)
+wc file =
+    File.toBytes file    -- SerialT IO Word8
+  & Stream.fold countAll -- IO (Int, Int, Int)
 ```
 
-## Concurrency Auto Scaling
+### Observations
```
adithyaov, line 132: I don't know how I feel about observations. Is there a better way to paraphrase this at the end?

README.md
```diff
@@ -11,572 +11,528 @@
...
-Semigroup and Monoid instances can be used to fold streams serially or
-concurrently. In the following example we compose ten actions in the
-stream, each with a delay of 1 to 10 seconds, respectively. Since all the
-actions are concurrent we see one output printed every second:
+## Concurrent Word Counting
 
-``` haskell
-import qualified Streamly.Prelude as S
-import Control.Concurrent (threadDelay)
+To do word counting in parallel we divide the stream in chunks (arrays),
```
adithyaov, line 168:
```suggestion
To count words in parallel we divide the stream into chunks (arrays),
```

README.md
```diff
@@ -11,572 +11,528 @@
...
+countArray arr =
+      Stream.unfold Array.read arr            -- SerialT IO Word8
+    & Stream.decodeLatin1                     -- SerialT IO Char
+    & Stream.foldl' count (Counts 0 0 0 True) -- IO Counts
 ```
-### Serial
+When combining the counts in two contiguous chunks, we would also need to
+know whether the first element of the next chunk was a space char or
+non-space to know whether the same word is continuing to the next chunk or
+if it is a new word. So add that too, giving (firstCharWasSpace, Counts).
```
adithyaov, line 191:
```suggestion
if it is a new word. Hence the return type should be of the form `(Bool, Counts)`. The first element of the tuple is `True` if the first character in the chunk is space.
```

README.md
```diff
@@ -11,572 +11,528 @@
...
+```
 
-loops = do
-    x <- S.fromFoldable [1,2]
-    y <- S.fromFoldable [3,4]
-    S.yieldM $ putStrLn $ show (x, y)
+Note that `Stream.aheadly` is the only difference in a concurrent and
+non-concurrent program. If we remove that we still have a perfectly valid,
+well performing serial program. Notice, how succinctly and idiomatically
+we expressed the concurrent word counting problem.
```
adithyaov, line 230: This is a better way to state observations than to have a separate heading for it.

README.md
```diff
@@ -11,572 +11,528 @@
...
 ```
 
-`AsyncT` is a stream monad transformer. If you are familiar with a list
-transformer, it is nothing but `ListT` with concurrency semantics. For example,
-the semigroup operation `<>` is concurrent. This makes `foldMap` concurrent
-too. You can replace `AsyncT` with `SerialT` and the above code will become
-serial, exactly equivalent to a `ListT`.
+Notice, how succinctly and idiomatically we expressed the concurrent
+directory tree traversal problem. Let us know if you can do it
+significantly better with any other language or framework.
```
adithyaov, line 388: This can be told commonly for all examples in the end. We can go with the assumption that folk generally don't read the text surrounding the example.

README.md
```diff
@@ -11,572 +11,528 @@
...
 
-### wc -l
+## Concurrency
 
-``` haskell
-wcl = S.length . S.splitOn (== 10) FL.drain . S.unfold FH.read
-main = withArg wcl >>= print
-```
+Streamly uses lock-free synchronization for concurrent operations. It employs
+auto-scaling of the degree of concurrency based on demand. For CPU bound tasks
```
adithyaov, line 442: >  It employs auto-scaling of the degree of concurrency based on demand.

Can this be rephrased, It's not easy to understand this.

README.md
```diff
@@ -11,572 +11,528 @@
...
-      S.fold avg
-    . S.splitOn (== 10) FL.length
-    . S.unfold FH.read
+As you have seen above in the word count example, streamly enables
+highly modular abstractions with the best possible performance (close to
+an equivalent C program).
 
-    where avg      = (/) <$> toDouble FL.sum <*> toDouble FL.length
-          toDouble = fmap (fromIntegral :: Int -> Double)
+For example, none of the existing Haskell libraries can work with byte
```
adithyaov, line 455: Although true, don't know how people will take it.

README.md
```diff
@@ -11,572 +11,528 @@
...
+by employing efficient abstractions like `Unfold`s and terminating
+`Fold`s. Byte level stream operations make programming simpler because
+you do not have to deal with chunking and re-combining.
 
-main = withArg avgll >>= print
-```
+If you can write a program significantly faster in some other way or
+with some other language, please let us know and we will improve. We
+still have some areas to improve but we will surely get there for most
+cases.
```
adithyaov, line 465: > We still have some areas to improve but we will surely get there for most cases.

We can probably omit this.

README.md
```diff
@@ -11,572 +11,528 @@
...
 
-### Line Length Histogram
+## Benchmarks
 
-`classify` is not released yet, and is available in
-`Streamly.Internal.Data.Fold`
+When it comes to stream operations, Haskell lists is the fastest
+existing implementation even though it supports only pure (not monadic)
+streams. Other streaming libraries are many times slower compared to
+lists, so lists is our gold standard for performance comparison.
```
adithyaov, line 472: > Other streaming libraries are many times slower compared to lists

Don't know how people take direct comparative statements.

README.md
```diff
@@ -11,572 +11,528 @@
...
-```
+This chart show only those operations where streamly and list
+performance differs by more than 10%. For all other operations the
+performance of both is more or less comparable.
+
+See [streaming
+benchmarks](https://github.com/composewell/streaming-benchmarks) for
+detailed comparison with other Haskell streaming libraries.
+
+## Performance and Modularity
```
adithyaov, line 489: We already have a `Performance` heading. Should this be named to something else. The content sounds more like `PS`-ish, notes-ish.

adithyaov (commented):
Looks good for the most part. Few minor comments. Example-based docs are straightforward and easier to read.

README.md
harendra-kumar: We are talking about composing folds in this para, so a brief intro of fold is needed.

README.md
harendra-kumar: We can remove it or just add a line in the previous section itself.

README.md
harendra-kumar: ok.

README.md
harendra-kumar: ok.

README.md
harendra-kumar: ok.

README.md
harendra-kumar: will do.

README.md
harendra-kumar: This is a key differentiation.

README.md
harendra-kumar: ok.

README.md
```diff
@@ -11,572 +11,528 @@
...
+    & Stream.maxThreads numCapabilities  -- AheadT IO (Bool, Counts)
+    & Stream.fromAhead                   -- SerialT IO (Bool, Counts)
+    & Stream.foldl' addCounts (False, Counts 0 0 0 True) -- IO (Bool, Counts)
+```
 
-loops = do
-    x <- S.fromFoldable [1,2]
-    y <- S.fromFoldable [3,4]
-    S.yieldM $ putStrLn $ show (x, y)
+Note that `Stream.aheadly` is the only difference in a concurrent and
```
rnjtranjan, line 227: `Stream.aheadly`  is not obvious about above "Stream.fromAhead"

README.md
harendra-kumar: That's a known and acknowledged fact, also proven by streaming-benchmarks below. We can still remove or rephrase it if it does not sound ok.

rnjtranjan (approved):

README.md
harendra-kumar: This remained after renaming. should be fixed.

adithyaov (approved):

PR #1065, opened 2021-05-23
streamly.cabal: Update description and tested-with

harendra-kumar:


streamly.cabal
```diff
@@ -45,8 +37,8 @@ license-file:        LICENSE
 tested-with:         GHC==8.0.2
                    , GHC==8.4.4
                    , GHC==8.6.5
-                   , GHC==8.8.1
-                   , GHC==8.10.1
+                   , GHC==8.8.4
+                   , GHC==8.10.4
```
adithyaov, line 41: The CIs run on more versions like `8.10.2`, `8.8.3` etc. Were they purposefully omitted?

adithyaov (approved):

streamly.cabal
harendra-kumar: Listing all minor versions would be very noisy, just latest should be enough.

PR #1077, opened 2021-05-31
Fix Pre-Release to 0.8.0

rnjtranjan:


src/Streamly.hs
```diff
@@ -18,16 +18,16 @@
 -- @
 -- > import "Streamly"
 -- > import "Streamly.Prelude" (cons, consM)
--- > import qualified "Streamly.Prelude" as S
+-- > import qualified "Streamly.Prelude" as Stream
```
harendra-kumar, line 21: The "Streamly" module is deprecated We can revert this file.

src/Streamly/Internal/Console/Stdio.hs
```diff
@@ -69,7 +69,7 @@ read = Unfold.lmap (\() -> stdin) Handle.read
 -- > getBytes = Handle.toBytes stdin
 -- > getBytes = Stream.unfold Stdio.read ()
 --
--- /Pre-release/
+-- @since 0.8.0
```
harendra-kumar, line 72: Only those APIs that are exposed to users through Streamly.Console.Stdio should be marked as 0.8.0, others should remain `Pre-release`.

src/Streamly/Memory/Array.hs
```diff
@@ -30,7 +30,7 @@
 --
 -- This module is designed to be imported qualified:
 --
--- > import qualified Streamly.Array as A
+-- > import qualified Streamly.Memory.Array as Array
```
harendra-kumar, line 33: This is a deprecated module. We do not need to change it.

harendra-kumar (changes requested):
Need to revert all those 0.8.0 annotations that are not exposed via released modules.

harendra-kumar (changes requested):
There are some more APIs in other modules that we need to mark 0.8.0. Go to the streamly website/reference docs, browse the released modules one by one and search for `pre-release` in the browser.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -547,7 +547,7 @@ read = Producer.simplify producer
 
 -- | Unfold an array into a stream in reverse order.
 --
--- /Pre-release/
+-- @since 0.8.0
```
harendra-kumar, line 550: This is not exposed anywhere, so we should keep it `pre-release`.

harendra-kumar (changes requested):

harendra-kumar (approved):

harendra-kumar:
Please squash the commits.

PR #1082, opened 2021-06-04
Fix Tutorial doctests

rnjtranjan:


src/Streamly/Tutorial.hs
```diff
@@ -554,40 +581,50 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
...
--- @
--- ThreadId 36: Delay 3
--- ThreadId 36: Delay 2
--- ThreadId 36: Delay 1
+-- >>> :{
+--   delay n = Stream.fromEffect $ do
+--      threadDelay (n * 1000000)
+--      tid <- myThreadId
+--      putStrLn (show tid ++ ": Delay " ++ show n)
+-- :}
```
harendra-kumar, line 602: Why are we defining `delay` again and again? Its already been defined before.

src/Streamly/Tutorial.hs
```diff
@@ -1165,18 +1244,13 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
...
--- (1,3)
--- (2,3)
--- (1,4)
--- (2,4)
+-- Stream.drain . Stream.fromWAsync $ do
+--     x <- Stream.fromFoldable [1,2]
+--     y <- Stream.fromFoldable [3,4]
+--     Stream.fromEffect $ putStrLn $ show (x, y)
+--
+-- print (1,3) (2,3) (1,4) (2,4) in any order
```
harendra-kumar, line 1253: We can apply 1 sec delay using`Stream.delay` on each stream and then the output should be predictable.

harendra-kumar (changes requested):

src/Streamly/Tutorial.hs
rnjtranjan: Tried with delay still output is not predictable:
 
Stream.drain . Stream.fromWAsync $ do
     x <- Stream.delay 1 $ Stream.fromFoldable [1,2]
     y <- Stream.delay 1 $ Stream.fromFoldable [3,4]
     Stream.fromEffect $ putStrLn $ show (x, y)
:}'
expected: (1,3) 
          (2,3) 
          (1,4) 
          (2,4) 
 but got: (1,3)
          (1,4)
           ^
          (2,3)
          (2,4)

harendra-kumar (approved):
I am going to make a few small changes and merge it.

PR #1084, opened 2021-06-06
Test CI for ghc-head

rnjtranjan:


adithyaov:
Looks like `ghcup` cannot find `head`.
Does `ghcup install ghc head` work?
Can you check the available versions by using `ghcup list`

adithyaov:
Both the following fail:
```
sudo -- sh -c apt-get -y install ghc-head
```
```
ghcup install ghc head
```
One of them should succeed.

adithyaov:
` ghc-head` is available in HVR PPA

harendra-kumar:
> ` ghc-head` is available in HVR PPA

I used to run ghc-head CIs earlier. In fact we have an example in packcheck/travis config:

```
  #- env: BUILD=cabal-v2 GHCVER=head
  #  addons: {apt: {packages: [cabal-install-head,ghc-head], sources: [hvr-ghc]}}
```

rnjtranjan:
ghcup list does not list "ghc head".

.github/workflows/haskell.yml
```diff
@@ -41,63 +41,17 @@ jobs:
...
-          - name: 8.10.4+stack+lts-17.8
-            build: stack
-            resolver: lts-17.8
-            sdist_options: "--ignore-check"
-            stack_build_options: "--flag streamly-benchmarks:-opt"
-            runner: ubuntu-latest
-          - name: 8.10.2+macOS
-            ghc_version: 8.10.2
+          - name: ghc_HEAD
+            ghc_version: head
```
harendra-kumar, line 48: Can you try "base-4.16.0.0" instead of "head"? You will have to remove the GHCVER env var line in the `env` section above, other wise packcheck will fail because the ghc version is not actually base-4.16.0.0.

harendra-kumar (changes requested):

.github/workflows/haskell.yml
```diff
@@ -61,6 +62,12 @@ jobs:
           - 8.4.4+debug
         cabal_version: ["3.4"]
         include:
+          - name: ghc_HEAD
+            ghc_version: 9.2.0.20210422 
+            build: cabal-v2
+            cabal_build_options: "--allow-newer=hsc2hs"
```
harendra-kumar, line 68: Use `cabal_build_options: "--allow-newer"` to allow newer versions of everything.

.github/workflows/haskell.yml
```diff
@@ -61,6 +62,12 @@ jobs:
           - 8.4.4+debug
         cabal_version: ["3.4"]
         include:
+          - name: ghc_HEAD
+            ghc_version: 9.2.0.20210422
```
harendra-kumar, line 66: Use `ghc_version: 9.2.0` so that we do not have to keep changing the version.

harendra-kumar (changes requested):
Change the ghc version to `9.2` instead of `9.2.0`. ghcup does not install if specified as 9.2.0.

harendra-kumar:
Let us install ghc via packcheck itself - https://github.com/composewell/packcheck/issues/67 . It should be easy.

harendra-kumar:
For now we can skip using `haskell/actions/setup@v1` in the ghc head test and install it manually (using `ghcup install ghc`) in the script section before calling packcheck.

harendra-kumar:
Added #1429 . Closing this.

PR #1096, opened 2021-06-10
Move the tutorial module to streamly-docs package

harendra-kumar:


cabal.project
```diff
@@ -1,3 +1,4 @@
 packages:   streamly.cabal
           , test/streamly-tests.cabal
           , benchmark/streamly-benchmarks.cabal
+          , docs/streamly-docs.cabal
```
adithyaov, line 4: `test` vs `benchmark` vs `docs`

singular vs plural

adithyaov: `documentation` vs `documents (docs)`

streamly.cabal
```diff
@@ -102,11 +102,11 @@ extra-source-files:
...
-    docs/Overview.md
-    docs/Build.md
-    docs/streamly-vs-async.md
-    docs/streamly-vs-lists.md
-    docs/transformers.md
+    docs/*.md
+    docs/Setup.hs
+    docs/Streamly/Doc/Tutorial.hs
+    docs/doctests.hs
+    docs/streamly-docs.cabal
```
adithyaov, line 109: This is going to be very tiresome. You'll have 100 documents in the near future in some hirerchy.
`docs/Streamly/Doc/Tutorial/Part1.md`

Also, The module structure generates a hierarchy in the documentation. Do you want `Streamly/Doc/` in that hierarchy? 
Don't think in terms of modules.
- Drop the `Streamly.Doc` prefix.
- Use `Tutorial.hs` (or `Tutorial/Server.hs`) for the tutorial.
Mix in `md` (`rst`) and add them in `extra-doc-files` in `streamly-doc.cabal`

adithyaov (commented):
Fre comments.

If this is just the initial commit to move stuff and you're going to work on the rest later, consider this approved.

cabal.project
harendra-kumar: This came to my mind as well. I thought we can change the test/benchmark directory names to plural later to make everything consistent.

streamly.cabal
harendra-kumar: The prefix was only to be consistent with tests and benchmarks, but we do not have to be. It can be dropped. We are not going to import these modules or expose them to users so the module names do not matter.

adithyaov (approved):

PR #1098, opened 2021-06-11
Separate chart build from benchmarks

rnjtranjan:


benchmark/bench-show/bench-chart.cabal
```diff
@@ -0,0 +1,20 @@
+cabal-version:      2.2
+name:               bench-chart
+version:            0.0.0
+synopsis:           Chart for Benchmarks
```
harendra-kumar, line 4: "Benchmark report generation"

benchmark/bench-show/bench-chart.cabal
```diff
@@ -0,0 +1,20 @@
...
+description: Chart is separated from the benchmarks.
```
harendra-kumar, line 5: We can say something like - "Benchmark reporting application is not included in the overall cabal project so that the dependencies of both can remain independent. Benchmark reporting has a lot of dependencies that usually lag behind when new GHC releases arrive."

benchmark/bench-show/bench-chart.cabal
```diff
@@ -0,0 +1,20 @@
...
+
+-------------------------------------------------------------------------------
+-- benchmark comparison and presentation
+-------------------------------------------------------------------------------
+
+executable chart
```
harendra-kumar, line 11: Let's call it `bench-report`. Also name this directory as `bench-report`.

benchmark/bench-show/bench-chart.cabal
```diff
@@ -0,0 +1,20 @@
...
+  default-language: Haskell2010
+  ghc-options: -Wall
+  hs-source-dirs: .
+  main-is: Chart.hs
```
harendra-kumar, line 15: BenchReport.hs

harendra-kumar (changes requested):

harendra-kumar (changes requested):
`no-charts` flag can now be removed. Also, squash the commits. Meanwhile I will try testing it.

rnjtranjan (commented):
looks good.

PR #1100, opened 2021-06-12
Upgrade nix, use a local .cabal dir

harendra-kumar:


adithyaov:
Can you check why the CIs are failing?

default.nix
```diff
@@ -123,6 +123,7 @@ let haskellPackages =
             else [];
         # Use a better prompt
         shellHook = ''
+          export CABAL_DIR="$(pwd)/.cabal.nix"
```
adithyaov, line 126: > The CABAL_DIR might be dropped in the future, when cabal-install starts to use XDG Directory specification.

https://cabal.readthedocs.io/en/3.4/installing-packages.html

harendra-kumar:
> Can you check why the CIs are failing?

appveyor is failing because the same known issue. doctests failure seems to be due to some script or infra issue. coverage test is know to fail often because of timeout.

default.nix
harendra-kumar: That would be even better then we will have the XDG environment variables to use see https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html .

.github/workflows/haskell.yml
```diff
@@ -50,52 +52,56 @@ jobs:
...
             runner: ubuntu-latest
-          - name: 8.8.4+stack+lts-16.31
-            build: stack
-            resolver: lts-16.31
-            sdist_options: "--ignore-check"
-            stack_build_options: "--flag streamly-benchmarks:-opt"
+          - name: 8.8.4
+            ghc_version: 8.8.4
+            build: cabal
+            cabal_project: cabal.project
```
adithyaov, line 93: Don't you need to disable sdist build if you want to include `cabal.project`?

default.nix
adithyaov: What is `.cabal.nix`? How does `CABAL_DIR` understand that?

harendra-kumar: It is just a location for cabal to store its config/store, non-existing dir to begin with, when you invoke cabal it will create its stuff there. CABAL_DIR is a way to tell that location to cabal.

.github/workflows/haskell.yml
harendra-kumar: CABAL_PROJECT if specified is used irrespective of sdist build or not.

PR #1104, opened 2021-06-14
Add "released" and "pre-release" to the "Stability" key in modules

adithyaov:
Remove all `Stability: experimental`?
```
find src -name "*.hs" -exec sed -i -r -e '/^-- Stability.*: experimental$/d' {} \;
```


harendra-kumar (approved):

harendra-kumar:
> Remove all `Stability: experimental`?

We can, but it is fine otherwise too. We can consider anything other than "release" or "pre-release" as internal/dev.

PR #1105, opened 2021-06-15
Add coverage build to github CI.

pranaysashank:


harendra-kumar:
We are not using coverage for a long time now.

PR #1106, opened 2021-06-15
Bump Win32 upper bound to work with GHC 9.0

georgefst:
Follows #933, which only works for other OSes.


harendra-kumar:
Thanks @georgefst ! Perhaps we need a 9.0.1 CI for macos/windows.

Issue #1087, opened 2021-06-08
Fix the performance of Foldable instance of streams for GHC-9


harendra-kumar:
Until a GHC fix becomes available we can use the workaround suggested in this issue: https://gitlab.haskell.org/ghc/ghc/-/issues/19895 .

harendra-kumar:
We need to fix maximum and minimum methods of Foldable. These are the benchmark comparison results ghc-8 vs ghc-9:

```
Benchmark                                                                                                     default(0)(Bytes) default(1) - default(0)(%)
------------------------------------------------------------------------------------------------------------- ----------------- --------------------------
All.Prelude.Serial/o-1-space.foldable.maximum                                                                          39427.00                  +31702.47
All.Prelude.Serial/o-1-space.foldable.minimum                                                                          39435.00                  +27284.34
```

Compare the two again after fixing.

PR #1111, opened 2021-06-18
Fix performance of Foldable instance of streams for GHC-9

rnjtranjan:


rnjtranjan:
Benchmark comparison for Prelude.Serial/o-1-space.foldable
**Without fix in GHC 9.0.1:**
minimum:                                                           OK
2.0 ms ± 2.8 ms, 9.9 MB allocated,  68 KB copied

maximum:                                                           OK (0.01s)
3.5 ms ± 2.9 ms,  13 MB allocated,  84 KB copied

**With oneShot fix in GHC 9.0.1::**
minimum:                                                           OK (0.01s)
 79 μs ± 206 μs,   0 B  allocated,   0 B  copied

maximum:                                                           OK (0.01s)
 99 μs ± 179 μs,   0 B  allocated,   0 B  copied

harendra-kumar (approved):
Please squash the commits.

harendra-kumar:
@rnjtranjan need to add a CPP conditional for the 8.0.2 CI failure.

rnjtranjan:
It's giving parse error like 
_src/Streamly/Internal/Data/Stream/Serial.hs:247:868: error:
    parse error on input ‘where’
    |
247 | FOLDABLE_INSTANCE(SerialT)_

when I put :

#if MIN_VERSION_base(4,15,0)                                                  \    
        where {f' x k = oneShot $ \z -> k $! f z x};                          \
#else                                                                                          \ 
        where {f' x k z = k $! f z x};                                                \
#endif                                                                                       \

I tried many diff combination with no success.

harendra-kumar:
You can define `oneShot` in `Streamly.Internal.BaseCompat`. For older versions you can define it as `id`.

rnjtranjan:
Fix #1087

src/Streamly/Internal/BaseCompat.hs
```diff
@@ -55,3 +57,10 @@ unsafeWithForeignPtr = GHCForeignPtr.unsafeWithForeignPtr
 #else
 unsafeWithForeignPtr = withForeignPtr
 #endif
+
+oneShot :: (a -> b) -> a -> b
+#if MIN_VERSION_base(4,15,0)
```
harendra-kumar, line 62: Our build was failing only for 8.0, which means in other version onseShot is present. So we should use min version of base that was shipped with 8.2.

PR #1129, opened 2021-06-24
toStream/toStreamRev should be able to fold to any monad

### LINKED ISSUE #930: toStream/toStreamRev should be able to fold to any monad
For example it could be:

```
 toStream :: (Monad m, Monad n) => Fold m a (SerialT n a) 
 toStream :: (Monad m, Monad n) => SerialT m a -> m (SerialT n a) 
```

This will allow us to fold it to either Identity or IO as needed.

rnjtranjan:
Fix #930

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
@@ -151,7 +151,6 @@ where
 
 import Control.Monad.Catch (MonadThrow)
 import Control.Monad.IO.Class (MonadIO(..))
-import Data.Functor.Identity (Identity (..))
 import Foreign.Storable (Storable)
 import Streamly.Internal.Data.Parser (Parser (..))
 import Streamly.Internal.Data.SVar (MonadAsync, defState)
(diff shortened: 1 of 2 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
```diff
@@ -800,8 +799,11 @@ toHandle h = go
 -- /Pre-release/
 --
 {-# INLINE toStream #-}
-toStream :: Monad m => SerialT m a -> m (SerialT Identity a)
-toStream = foldr K.cons K.nil
+toStream :: (Monad m, Monad n)
+    => (forall x. m x -> n x) -> SerialT m a -> m (SerialT n a)
+toStream  f strm = do
```
adithyaov, line 804:
```suggestion
toStream f strm = do
```

src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
```diff
@@ -800,8 +799,11 @@ toHandle h = go
 -- /Pre-release/
 --
 {-# INLINE toStream #-}
-toStream :: Monad m => SerialT m a -> m (SerialT Identity a)
-toStream = foldr K.cons K.nil
+toStream :: (Monad m, Monad n)
+    => (forall x. m x -> n x) -> SerialT m a -> m (SerialT n a)
```
harendra-kumar, line 803: This is not what I had in mind. It was supposed to be like this:

```
toStream :: Monad m => SerialT m a -> m (SerialT n a)
toStream = foldr K.cons K.nil
```

PR #1133, opened 2021-06-30
Add a document about streaming paradigms

harendra-kumar:


adithyaov (approved):
Since we are listing out the available libraries it would be nice to also show where steamly fits into this.

adithyaov:
Add this to the `extra-doc-files`.

harendra-kumar:
> Since we are listing out the available libraries it would be nice to also show where steamly fits into this.

Last entry in the first table is streamly.

PR #1135, opened 2021-07-04
Add diff-style and diff-cutoff to bench-report

harendra-kumar:
So that we can use it for streaming-benchmarks use cases as well.

### DIFF
--- a/benchmark/bench-report/bin/bench-runner.sh
+++ b/benchmark/bench-report/bin/bench-runner.sh
@@ -180,7 +180,9 @@ bench_exec_one() {
   local BENCH_NAME
   # XXX this is a hack to make the "/" separated names used in the functions
   # determining options based on benchmark name. For tasty-bench the benchmark
-  # names are separated by "." instead of "/".
+  # names are separated by "." instead of "/" and are prefixed by "All".
+  #
+  # For example, All.Prelude.Serial/x.y.z => Prelude.Serial/x/y/z
   if test "$USE_GAUGE" -eq 0
   then
     # Remove the prefix "All."
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -177,8 +177,6 @@ run_target () {
 
   echo "Running executable $target_name ..."
 
-  # Needed by bench-exec-one.sh
-  export BENCH_EXEC_PATH=$target_prog
   mkdir -p $(dirname $(get_tix_file $target_name))
   export HPCTIXFILE=$(get_tix_file $target_name)
 


benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -176,15 +178,28 @@ bench_exec_one() {
   local BENCH_NAME1
   local BENCH_NAME2
   local BENCH_NAME
+  # XXX this is a hack to make the "/" separated names used in the functions
+  # determining options based on benchmark name. For tasty-bench the benchmark
+  # names are separated by "." instead of "/".
```
adithyaov, line 183: Can you give a small example of what this hack script intends to do?

benchmark/bench-report/bin/build-lib.sh
```diff
@@ -204,47 +204,3 @@ run_build () {
...
-# $1: package name
-# $2: component
-# $3: targets
-# $4: args generator func
-run_targets() {
-    for i in $3
-    do
-      run_target $1 $2 $i $4
-    done
-}
```
adithyaov, line 250: These were not used in benchmarking? They are used in `test.sh` now. Why were they added to benchmark lib in the first place?

bin/test.sh
```diff
@@ -156,9 +156,53 @@ fi
...
+  local target_prog
+  target_prog=$(cabal_target_prog $package_name $component $target_name) || \
+    die "Cannot find executable for target $target_name"
+
+  echo "Running executable $target_name ..."
+
+  # Needed by bench-exec-one.sh
+  export BENCH_EXEC_PATH=$target_prog
+  mkdir -p $(dirname $(get_tix_file $target_name))
+  export HPCTIXFILE=$(get_tix_file $target_name)
```
adithyaov, line 183: bench stuff in `test.sh`?

adithyaov: Are these required in bench lib?

adithyaov (commented):
I don't see any obvious problems other than the code refactor in the scripts.

benchmark/bench-report/bin/build-lib.sh
harendra-kumar: Originally they were common but then became different for bench and test.

bin/test.sh
harendra-kumar: These are remains of earlier code, can be removed.

benchmark/bench-report/bin/bench-runner.sh
harendra-kumar: I will try adding an example.

PR #1140, opened 2021-07-12
Add INLINE pragmas on lefts and rights

adithyaov:


harendra-kumar (approved):
we can add it on catMaybes too.

PR #1145, opened 2021-07-18
Comment out writeArray and remove unsafeIndex in Array.Foreign(.Mut)

adithyaov:


src/Streamly/Internal/Data/Array/Foreign.hs
```diff
@@ -104,7 +104,7 @@ module Streamly.Internal.Data.Array.Foreign
     -- , writeChunksOfFrom
     -- , ...
 
-    , writeIndex
+    --, writeIndex
```
harendra-kumar, line 107: Instead of commenting it out please move to Data.Array.Foreign.Mut module. Even the commented out writeSlice etc should be moved there.

harendra-kumar (changes requested):
unsafeIndexIO can be renamed to unsafeIndex and we can make it work for any MonadIO.

src/Streamly/Internal/Data/Array/Foreign.hs
adithyaov: Yes, there needs to be a lot of learning. This is only a minor change for some correctness.

adithyaov:
> unsafeIndexIO can be renamed to unsafeIndex and we can make it work for any MonadIO.

Noted.

harendra-kumar (approved):

PR #1154, opened 2021-07-26
Use property test haddock markup

### LINKED ISSUE #1136: Use property test haddock markup
Use the `prop>` haddock markup in docs wherever we have equivalence properties specified. Also, use `cabal-docspec` to verify those properties.

rnjtranjan:
Fix #1136


rnjtranjan:
Another PR https://github.com/composewell/streamly/pull/1175

PR #1155, opened 2021-07-26
Mutable array and array cleanup in general

adithyaov:


src/Streamly/Internal/System/IO.hs
```diff
@@ -0,0 +1,52 @@
+{-# LANGUAGE UnboxedTuples #-}
+
+-- |
+-- Module      : Streamly.Internal.System.IO
+-- Copyright   : (c) 2020 Composewell Technologies
+-- License     : BSD3-3-Clause
```
harendra-kumar, line 6: BSD-3-Clause

harendra-kumar (approved):
Squash the last two commits.

PR #1162, opened 2021-07-28
Continuation of #1155

adithyaov:


src/Streamly/Internal/Data/Array/Foreign/Mut.hs
```diff
@@ -44,14 +49,46 @@ module Streamly.Internal.Data.Array.Foreign.Mut
...
+    -- * Composable Folds
+    , writeN
+    , write
+
+    -- * Unfolds
+    , read
+    , readRev
+    , producer
+    , flattenArrays
+    , flattenArraysRev
```
harendra-kumar, line 66: We can avoid exporting producer, flatten*, we can export on need basis.

src/Streamly/Internal/Data/Array/Foreign/Mut.hs
```diff
@@ -44,14 +49,46 @@ module Streamly.Internal.Data.Array.Foreign.Mut
...
+
+    -- * To containers
+    , toStreamD
+    , toStreamDRev
+    , toStreamK
+    , toStreamKRev
```
harendra-kumar, line 72: Avoid exporting these as they are specific to stream types.

harendra-kumar (approved):

PR #1163, opened 2021-07-28
Some changes for streamly process

harendra-kumar:
Need to check any perf impact due to `bracket` changes.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Exception.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Exception.hs
@@ -13,6 +13,7 @@ module Streamly.Internal.Data.Stream.IsStream.Exception
     , after
     , bracket_
     , bracket
+    , bracket'
     , onException
     , finally_
     , finally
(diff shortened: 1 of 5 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
```diff
@@ -494,6 +495,18 @@ unfoldManyRoundRobin ::(IsStream t, Monad m)
...
+--
+-- | Like 'unfoldMany' but unfolds @Left a@ into @Left b@ and @Right a@ into
+-- @Right b@.
+--
+-- /Pre-release/
+{-# INLINE unfoldManyEither #-}
+unfoldManyEither ::(IsStream t, Monad m) =>
+    Unfold m a b -> t m (Either a a) -> t m (Either b b)
+unfoldManyEither u m = fromStreamD $ D.unfoldManyEither u (toStreamD m)
+
```
adithyaov, line 509: Wouldn't `unfoldMany` on an unfold of type `Unfold (Either a a) (Either b b)` work?

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -262,11 +264,16 @@ bracket_ bef aft bet =
...
-    gbracket bef (inline MC.try) aft
-        (\a (e :: SomeException) _ -> aft a >> return (nilM (MC.throwM e))) bet
+bracket :: (MonadAsync m, MonadCatch m) =>
+       m b
+    -> (b -> m c)
+    -> (b -> m d)
+    -> (b -> m e)
+    -> (b -> Stream m a)
+    -> Stream m a
+bracket bef aft exc gc bet =
```
adithyaov, line 274: Can we keep the order of handlers same as in `gbracket`?
`bef exe aft gt`?
Basically keep a consistent order?

src/Streamly/Internal/Data/Stream/IsStream/Exception.hs
```diff
@@ -167,7 +168,24 @@ bracket_ bef aft bet = D.fromStreamD $
 bracket :: (IsStream t, MonadAsync m, MonadCatch m)
     => m b -> (b -> m c) -> (b -> t m a) -> t m a
 bracket bef aft bet = D.fromStreamD $
-    D.bracket bef aft (toStreamD . bet)
+    D.bracket bef aft aft aft (toStreamD . bet)
```
adithyaov, line 171: I don't know if I like the inconsistency between the API between StreamD and IsStream. We could add another combinator and eventually replace bracket with it.

adithyaov (commented):
Can you open an issue to add benchmarks and tests to the new combinators?

src/Streamly/Internal/Data/Stream/IsStream/Expand.hs
harendra-kumar: We need a transformation using `Unfold m a b`. If we can transform `Unfold m a b` to `Unfold m (f a) (f b)` using a `f a -> f b` then it might work.

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
harendra-kumar: gbracket is the one that should be changed, I did not change it to keep the change minimal.

src/Streamly/Internal/Data/Stream/IsStream/Exception.hs
harendra-kumar: We can call it `D.bracket'`, it will be consistent with `bracket'`.

adithyaov (approved):
LGTM.
Can you open an issue to add benchmarks and tests to the new combinators?

harendra-kumar:
No perf difference in exception benchmarks observed.

PR #1164, opened 2021-07-29
Add Internal.Unicode.Utf8

adithyaov:


adithyaov:
Need to add some benchmarks with a little complex pipeline.

src/Streamly/Internal/Unicode/Utf8.hs
```diff
@@ -0,0 +1,66 @@
+-- |
+-- Module      : Streamly.Internal.Unicode.Utf8
+-- Copyright   : (c) 2020 Composewell Technologies
```
harendra-kumar, line 3: 2021

harendra-kumar (commented):
Writing benchmarks right from the beginning may be a good idea.

harendra-kumar (approved):

PR #1170, opened 2021-07-31
Fix a fusion issue failing in bench:Unicode.Stream

adithyaov:


harendra-kumar (approved):
Can you check performance diff before and after your overall changes committed till now?

adithyaov:
> Can you check performance diff before and after your overall changes committed till now?

Regardless of the perf, this PR should be merged. Will get back to you with the array & filesystem benchmark comparison.

PR #1174, opened 2021-08-01
Add Parser.choice

adithyaov:


test/Streamly/Test/Data/Parser.hs
```diff
@@ -372,6 +372,25 @@ takeWhile1 =
...
+                      checkListEqual parsed_list $ take (min i j) ls
+                  Left _ -> property False
+
+    where
+
+    parser i = P.fromFold (FL.take i FL.toList)
+
+
+
+
```
harendra-kumar, line 393: Remove extra blank lines.

benchmark/Streamly/Benchmark/Data/Parser/ParserD.hs
```diff
@@ -64,6 +64,20 @@ benchIOSink value name f =
...
+
+{-# INLINE choice2 #-}
+choice2 :: MonadCatch m => Int -> SerialT m a -> m ()
+choice2 value =
+    IP.parseD
+        (PR.choice
+             [ PR.fromFold (FL.take (value `div` 2) FL.drain)
+             , PR.fromFold (FL.take (value `div` 2) FL.drain)
+             ])
+
```
harendra-kumar, line 80: We already have a benchmark for `choice` which uses `asum`, we can rename that to `choice (asum)`. For this benchmark we should use exactly the same test as for `asum` to make a fair comparison. Also, just one benchmark is enough.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -929,13 +929,11 @@ concatMap f p = K.toParserK $ D.concatMap (K.fromParserK . f) (K.fromParserK p)
...
 --
--- /Unimplemented/
+-- /Internal/
 --
 {-# INLINE choice #-}
-choice ::
-    -- Foldable t =>
-    t (Parser m a b) -> Parser m a b
-choice _ps = undefined
+choice :: (Functor t, Foldable t, MonadCatch m) => t (Parser m a b) -> Parser m a b
```
harendra-kumar, line 935: Going beyond 80 cols.

src/Streamly/Internal/Data/Parser.hs
```diff
@@ -929,13 +929,11 @@ concatMap f p = K.toParserK $ D.concatMap (K.fromParserK . f) (K.fromParserK p)
 --
 -- This is same as 'asum' but more efficient.
 --
--- /Unimplemented/
+-- /Internal/
```
harendra-kumar, line 932: `Pre-release`.

test/Streamly/Test/Data/Parser.hs
```diff
@@ -372,6 +372,25 @@ takeWhile1 =
         where
             predicate = (== 0)
 
+choice :: Property
+choice =
```
harendra-kumar, line 376: Can we test `asum` as well using this same test?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -883,16 +883,14 @@ sequence _f _p = undefined
...
--- /Unimplemented/
+-- /Internal/
 --
 {-# INLINE choice #-}
-choice ::
-    -- Foldable t =>
-    t (Parser m a b) -> Parser m a b
-choice _ps = undefined
+choice :: (MonadCatch m, Foldable t) => t (Parser m a b) -> Parser m a b
+choice = foldl1 shortest
```
harendra-kumar, line 893: `shortest` is marked as `Broken`, so we need to mark this also broken unless we review and fix that.

src/Streamly/Internal/Data/Parser.hs
adithyaov: `Broken`

test/Streamly/Test/Data/Parser.hs
adithyaov: Using the same test failed, so probably not. I have not looked into why it failed.

benchmark/Streamly/Benchmark/Data/Parser/ParserD.hs
```diff
@@ -351,7 +357,8 @@ o_n_space_serial value =
     , benchIOSink value "sequenceA_/100" $ sequenceA_ (value `div` 100)
     , benchIOSink value "sequence/100" $ sequence (value `div` 100)
     , benchIOSink value "sequence_/100" $ sequence_ (value `div` 100)
-    , benchIOSink value "choice/100" $ choice (value `div` 100)
+    , benchIOSink value "choice (asum)/100" $ choiceAsum (value `div` 100)
+    , benchIOSink value "choice" $ choice (value `div` 100)
```
harendra-kumar, line 361: Should name it to `choice/100` to indicate that this benchmark is 100 times shorter than usual.

test/Streamly/Test/Data/Parser.hs
harendra-kumar: That means there is something wrong with this or with asum. We should file an issue if we are not investigating it right now.

harendra-kumar (approved):
There are a few minor comments.

harendra-kumar (approved):

test/Streamly/Test/Data/Parser.hs
adithyaov: See https://github.com/composewell/streamly/issues/1191

PR #1175, opened 2021-08-03
Add Haddock property test

rnjtranjan:


rnjtranjan:
fix #1136

rnjtranjan:
All the prop> tests are failing due to compile issues.

src/Streamly/Internal/Data/Pipe.hs
```diff
@@ -928,16 +928,16 @@ splitSuffixBy' predicate f m =
...
 -- The following law holds when the separator is non-empty and contains none of
 -- the elements present in the input lists:
 --
--- > splitOn . intercalate == id
+-- prop> splitOn . intercalate == id
 --
 -- The following law always holds:
 --
--- > concat . splitOn . intercalate == concat
+-- prop> concat . splitOn . intercalate == concat
```
harendra-kumar, line 940: Let's not fix the `Pipe` module, this is an unfinished module.

src/Streamly/Internal/Data/Stream/IsStream/Common.hs
```diff
@@ -547,12 +547,12 @@ concatM generator = concatMapM (\() -> generator) (fromPure ())
...
 -- 'splitOnSeq' is an inverse of 'intercalate'. The following law always holds:
 --
--- > intercalate . splitOn == id
+-- prop> intercalate . splitOn == id
 --
 -- The following law holds when the separator is non-empty and contains none of
 -- the elements present in the input lists:
 --
--- > splitOn . intercalate == id
+-- prop> splitOn . intercalate == id
```
harendra-kumar, line 555: The description above says this does not always hold, so this cannot be marked with `prop>`.

src/Streamly/Prelude.hs
```diff
@@ -600,8 +600,8 @@ module Streamly.Prelude
     --
     -- The following equations hold for lists:
     --
-    -- > scanl f z xs == map (foldl f z) $ inits xs
-    -- > scanr f z xs == map (foldr f z) $ tails xs
+    -- prop> scanl f z xs == map (foldl f z) $ inits xs
+    -- prop> scanr f z xs == map (foldr f z) $ tails xs
```
harendra-kumar, line 604: These examples are conceptual examples based on Data.List, we cannot doctest these for streams, if at all we want to doctest them then we should use the Data.List versions of these functions.

harendra-kumar:
There is nothing left to do in this one.

PR #1176, opened 2021-08-03
Add retry

adithyaov:


src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -349,3 +352,53 @@ _handle f (Stream step state) = Stream step' (Left state)
...
+    -> Stream m a
+retry emap0 han0 (Stream step0 state0) = Stream step state
+
+    where
+
+    state = WithOldStream emap0 state0
+
+    {-# INLINE_LATE step #-}
+    step gst (WithOldStream emap st) = do
+        -- XXX gtry? Whats the difference?
```
harendra-kumar, line 377: I don't understand this.

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -349,3 +352,53 @@ _handle f (Stream step state) = Stream step' (Left state)
...
+        res <- step1 gst state1
+        return
+            $ case res of
+                  Yield x st1 -> Yield x $ WithNewStream (Stream step1 st1)
+                  Skip st1 -> Skip $ WithNewStream (Stream step1 st1)
+                  Stop -> Stop
+
+    {-# INLINE han #-}
+    han e emap st =
```
harendra-kumar, line 397: I would prefer full name `handler`.

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -349,3 +352,53 @@ _handle f (Stream step state) = Stream step' (Left state)
...
+        case Map.lookup e emap of
+            Just i
+                | i > 0 ->
+                    let emap1 = Map.insert e (i - 1) emap
+                     in return $ Skip $ WithOldStream emap1 st
+                | otherwise -> MC.throwM e
```
harendra-kumar, line 403: We should use the default handler for this as well so that the user has a single way to deal with all of them. Otherwise they will have to catch this separately.

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -349,3 +352,53 @@ _handle f (Stream step state) = Stream step' (Left state)
...
+        eres <- MC.try $ step0 gst st
+        case eres of
+            Left e -> han e emap st
+            Right res ->
+                return
+                    $ case res of
+                          Yield x st1 -> Yield x $ WithOldStream emap st1
+                          Skip st1 -> Skip $ WithOldStream emap st1
+                          Stop -> Stop
+    step gst (WithNewStream (Stream step1 state1)) = do
```
harendra-kumar, line 387: Use `UnStream` here, we have seen bad performance if `Stream` is used. See `concatMap` for StreamD.

harendra-kumar (changes requested):
Would be nice if we have a simple test to show that it works?

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
adithyaov: That's what I've written initially, but I don't remember why I changed to this.

src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
adithyaov: Me neither. I think it was essentially about `ExceptionMonad vs MonadCatch`

test/Streamly/Test/Prelude/Serial.hs
```diff
@@ -674,3 +710,5 @@ main = hspec
 
     describe "Nesting" $ do
         prop "foldIterateM" foldIterateM
+
+    describe "Tests for retry" retry
```
harendra-kumar, line 714: Shouldn't  this go under `"Tests for exceptions"` above? Also the code block should be moved next to `exceptionOps`.

harendra-kumar (approved):
Looks good. The test code may be organized better.

PR #1178, opened 2021-08-04
Add an array parser with DLL

adithyaov:


adithyaov:
Closing this for now. We can use it if we think this will be better.

Pros of this over existing:
- It would be significantly faster if there is considerable backtracking. (Need to prove it with a benchmark)

Cons of this over existing:
- Relatively complex code.
- The existing had similarities with `parse`.

PR #1189, opened 2021-08-07
Update documentation and expose a few APIs in FileSystem.Handle

harendra-kumar:
These changes are mostly motivated by the FileSystem.Handle changes, so in the same PR:

* update documentation in some Stream functions
* Add some unimplemented function in Data.Unfold
* Update some notes in array stream module

### DIFF
--- a/src/Streamly/FileSystem/Handle.hs
+++ b/src/Streamly/FileSystem/Handle.hs
@@ -9,49 +9,59 @@
 -- Stability   : released
 -- Portability : GHC
 --
--- Read and write streams and arrays to and from file handles. File handle IO
--- APIs are quite similar to "Streamly.Data.Array.Foreign" read write APIs. In that
--- regard, arrays can be considered as in-memory files or files can be
--- considered as on-disk arrays.
+-- >>> import qualified Streamly.FileSystem.Handle as Handle
 --
--- Control over file reading and writing behavior in terms of buffering,
--- encoding, decoding is in the hands of the programmer, the 'TextEncoding',
--- 'NewLineMode', and 'Buffering' options of the underlying handle provided by
--- GHC are not needed and ignored.
+-- Read and write byte streams and array streams to and from file handles
+-- ('Handle').
 --
--- = Programmer Notes
+-- The 'TextEncoding', 'NewLineMode', and 'Buffering' options of the underlying
+-- GHC 'Handle' are ignored by these APIs. Please use @Streamly.Unicode.*@
+-- modules for encoding and decoding a byte stream, use stream splitting
+-- operations in "Streamly.Prelude" to create a stream of lines or to split the
+-- input stream on any other type of boundaries.
 --
--- > import qualified Streamly.FileSystem.Handle as Handle
+-- To set the read or write start position use 'hSeek' on the 'Handle', the
+-- 'Streamly.Prelude.before' combinator may be used to do that on a streaming
+-- combinator.  To restrict the length of read or write use the stream trimming
+-- operations like 'Streamly.Prelude.take'.
+--
+-- Note that a 'Handle' is inherently stateful, therefore, we cannot use these
+-- APIs from multiple threads without serialization; reading or writing in one
+-- thread would affect the file position for other threads.
 --
 -- For additional, experimental APIs take a look at
 -- "Streamly.Internal.FileSystem.Handle" module.
---
--- = Performance Notes
---
--- In some cases the stream type based APIs in the
--- "Streamly.Internal.FileSystem.Handle" module may be more efficient compared
--- to the unfold/fold based APIs exposed from this module because of better
--- fusion by GHC. However, with the streamly fusion GHC plugin (upcoming) these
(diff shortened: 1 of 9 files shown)


rnjtranjan (commented):
There are few docspec failures.

harendra-kumar:
Fixed.

rnjtranjan (approved):

src/Streamly/FileSystem/Handle.hs
```diff
@@ -9,68 +9,76 @@
...
--- GHC are not needed and ignored.
+-- Read and write byte streams and array streams to and from file handles
+-- ('Handle').
 --
--- = Programmer Notes
+-- The 'TextEncoding', 'NewLineMode', and 'Buffering' options of the underlying
+-- GHC 'Handle' are ignored by these APIs. Please use @Streamly.Unicode.*@
+-- modules for encoding and decoding a byte stream, use stream splitting
+-- operations in "Streamly.Prelude" to create a stream of lines or to split the
+-- input stream on any other type of boundaries.
```
adithyaov, line 21: We would probably have this in `Streamly.Utf8`

src/Streamly/FileSystem/Handle.hs
```diff
@@ -9,68 +9,76 @@
...
+    -- sequentially.  Read requests to the IO device are performed in chunks
+    -- limited to a maximum size of
+    -- 'Streamly.Internal.System.IO.defaultChunkSize'.  Note that the size of
+    -- the actual chunks in the resulting stream may be less than the
+    -- @defaultChunkSize@ but it can never exceed it.  If the whole stream is
+    -- not consumed, it is possible that we may have read slightly more from
+    -- the IO device than what the consumer needed.  Unless specified otherwise
+    -- in the API, writes are collected into chunks of
+    -- 'Streamly.Internal.System.IO.defaultChunkSize' before they are written
+    -- to the IO device.
```
adithyaov, line 59:
```suggestion
    -- * Streaming IO
    -- | Read or write a stream of data to or from a file or device
    -- sequentially.  
    --
    -- Read requests to the IO device are performed in chunks
    -- limited to a maximum size of
    -- 'Streamly.Internal.System.IO.defaultChunkSize'.  Note that the size of
    -- the actual chunks in the resulting stream may be less than the
    -- @defaultChunkSize@ but it can never exceed it.  If the whole stream is
    -- not consumed, it is possible that we may have read slightly more from
    -- the IO device than what the consumer needed.  
    --
    -- Writes are collected into chunks of
    -- 'Streamly.Internal.System.IO.defaultChunkSize' before they are written
    -- to the IO device unless specified otherwise in the API
```

adithyaov: Basically split them a little. When combined, suddenly you read about `write` API with the whole context of `read` in mind.

src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -356,17 +360,31 @@ discardSecond = lmap fst
 swap :: Unfold m (a, c) b -> Unfold m (c, a) b
 swap = lmap Tuple.swap
 
+-- See StreamD.scanlM' for implementing this.
+--
+-- | Scan the input of an 'Unfold' to change it in a stateful.
```
adithyaov, line 365: I don't understand this.

src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -356,17 +360,31 @@ discardSecond = lmap fst
...
+{-# INLINE_NORMAL lscanlM' #-}
+lscanlM' :: -- Monad m =>
+    (b -> a -> m b) -> m b -> Unfold m b c -> Unfold m a c
+lscanlM' = undefined
+
 -------------------------------------------------------------------------------
 -- Output operations
 -------------------------------------------------------------------------------
 
+-- XXX Do we need this combinator or the stream based idiom is enough?
```
adithyaov, line 377: This is a nice combinator. We can have this for completeness.

src/Streamly/Internal/Data/Unfold.hs
adithyaov: ... in a stateful manner

adithyaov (approved):
Looks good.

Recheck the perf after lifting to a higher monad `MonadIO`

src/Streamly/FileSystem/Handle.hs
harendra-kumar: I knew about it, but then I thought maybe its fine. will change.

src/Streamly/Internal/Data/Unfold.hs
harendra-kumar: We can change the implementation to stream based idiom, if the perf is not affected.

src/Streamly/Internal/Data/Unfold.hs
harendra-kumar: typo, will fix.

harendra-kumar:
Fixed.

PR #1197, opened 2021-08-11
Implement common FS events

rnjtranjan:


rnjtranjan:
Fixes #865

rnjtranjan:
working on Mac XXX comments.

src/Streamly/Internal/FileSystem/Event.hs
```diff
@@ -0,0 +1,236 @@
...
+#elif defined(CABAL_OS_LINUX)
+watch = Event.watchWithFlags
+    [ 0x00000002    -- File was modified.
+    , 0x00000004    -- Metadata changed.
+    , 0x00000100    -- Subfile was created.
+    , 0x00000200    -- Subfile was deleted.
+    , 0x00000400    -- Root was deleted.
+    , 0x00000040    -- File was moved from X.
+    , 0x00000080    -- File was moved to Y.
+    ]
```
harendra-kumar, line 119: This is not right, it makes the defaultConfig of windows module different than what `watch` is using, which is terrible. We should set the flags in the `defaultConfig` of the windows module.

harendra-kumar (changes requested):
I think we should remove watchWithFlags and watchRecursiveWithFlags. Instead the Config can be used to set these flags and `watchWith` will behave according to the config supplied.

test/Streamly/Test/FileSystem/Event.hs
```diff
@@ -181,19 +184,28 @@ driver ::
...
+                  createDirectoryLink fp (fp  ++ "SymLink")
+                  pre (fp  ++ "SymLink")
+                  let eventStream = checker ec (fp  ++ "SymLink") sync events
+                      fsOps = S.fromEffect $ runFSOps (fp  ++ "SymLink") sync
+                  fmap fromJust $ S.head $ eventStream `S.parallelFst` fsOps
+            else do
+                  pre fp
+                  let eventStream = checker ec fp sync events
+                      fsOps = S.fromEffect $ runFSOps fp sync
+                  fmap fromJust $ S.head $ eventStream `S.parallelFst` fsOps
```
harendra-kumar, line 208: The code in the "else" part is common with the "if" branch, you can just pass `fp` or `fp ++ "SymLink".`

harendra-kumar (changes requested):

rnjtranjan:
All event types except below has been tested:
IN_UNMOUNT
IN_Q_OVERFLOW
IN_ONLYDIR
IN_DONT_FOLLOW
IN_MASK_ADD
IN_ONESHOT

harendra-kumar:
merged via #1328

PR #1205, opened 2021-08-18
Refactor SVar module

harendra-kumar:
Review can start with the module structure changes in streamly.cabal. This is a major refactor of the SVar module and some other high level modules associated with concurrency.

* Break it up into smaller modules with clearly separated functionality
* Keep minimal stuff in the lower level module SVar/Type, and push the
  rest in a higher level module to reduce dependencies of lower level
  modules. This will help in separating the lower modules in streamly-core
  package.


harendra-kumar:
I compared the benchmarks before and after the refactor, there is no noticeable change.

src/Streamly/Internal/Data/SVar/Worker.hs
```diff
@@ -0,0 +1,544 @@
...
+    (
+    -- * Adjusting Limits
+      decrementYieldLimit
+    , incrementYieldLimit
+    , decrementBufferLimit
+    , incrementBufferLimit
+    , resetBufferLimit
+
+    -- * Rate Control
+    , Work (..)
```
adithyaov, line 20: Should `WorkInfo` also be in this module?

src/Streamly/Internal/Data/SVar/Worker.hs
```diff
@@ -0,0 +1,544 @@
...
+    tid <- myThreadId
+    void $ sendToProducer sv (ChildStop tid Nothing)
+
+-------------------------------------------------------------------------------
+-- Send exceptions
+-------------------------------------------------------------------------------
+
+{-# NOINLINE handleFoldException #-}
+handleFoldException :: SVar t m a -> SomeException -> IO ()
+handleFoldException sv e = do
```
adithyaov, line 536: The consumer fold?

src/Streamly.hs
```diff
@@ -210,7 +210,8 @@ module Streamly {-# DEPRECATED "Please use \"Streamly.Prelude\" instead." #-}
 where
 
 import Data.Semigroup (Semigroup(..))
-import Streamly.Internal.Data.SVar (MonadAsync, Rate(..))
+import Streamly.Internal.Control.Concurrent (MonadAsync)
+import Streamly.Internal.Data.SVar (Rate(..))
```
adithyaov, line 214: We'll have to maintain deprecated modules? Do we remove them after a few releases after deprecating them?

adithyaov (approved):

adithyaov:
Looks good

src/Streamly.hs
harendra-kumar: We will remove after a while.

src/Streamly/Internal/Data/SVar/Worker.hs
harendra-kumar: Yes.

src/Streamly/Internal/Data/SVar/Worker.hs
harendra-kumar: No it is used in the SVar type definition so has to be in SVar.Type.

PR #1213, opened 2021-08-26
Add a CI that fails on trailing spaces in code

rnjtranjan:


.circleci/config.yml
```diff
@@ -388,7 +388,12 @@ jobs:
               HLINT_OPTIONS: lint --cpp-include=src --cpp-include=test
               HLINT_TARGETS: src test benchmark
             command: |
-              bash -c "$PACKCHECK $BUILD"
+              bash -c "$PACKCHECK $BUILD" || exit 1
+              if [ $? -eq 0 ]
```
harendra-kumar, line 392: This test is redundant as we will already exit if the previous commnds fails.

Issue #1208, opened 2021-08-19
ghcjs build is failing with out-of-memory


harendra-kumar:
The issue seems to be with `Streamly.Internal.Unicode.Stream` module with the latest commit (c9232b0e25d895d268eb9f38a6b1fa0e01b75dc6) in that file. On the previous commit it works.

adithyaov:
Related to https://github.com/composewell/streamly/issues/1193

harendra-kumar:
We need to test at commits between 9a253dffd8b4a320dd200ef8e3fdfcd18483aee7 https://github.com/composewell/streamly/commit/c9232b0e25d895d268eb9f38a6b1fa0e01b75dc6 . We can bisect to find out the real culprit.

harendra-kumar:
Looks like some template haskell linking is going on when it hangs:

```
[113 of 120] Compiling Streamly.Internal.Unicode.Stream ( src/Streamly/Internal/Unicode/Stream.hs, /home/harendra/composewell/streamly1/dist-newstyle/build/x86_64-linux/ghcjs-8.4.0.1/streamly-0.8.0/build/Streamly/Internal/Unicode/Stream.js_o ) [Streamly.Data.Array.Foreign changed]
[...]

Issue #1207, opened 2021-08-19
Implement limit-build-mem build flag for library compilation as well


harendra-kumar:
We have the flag to restrict heap for benchmarks and test compilation, but not for the library. We have an instance where ghcjs failed with out-of-memory. Should implement this flag for library as well and enable for one ghc build and ghcjs build as well.

harendra-kumar:
Fixed by #1222 .

PR #1222, opened 2021-09-05
Fix GHCJS build taking too much memory

harendra-kumar:


harendra-kumar:
Fixes #1207 . Fixes #1208 .

adithyaov:
This is more like bypassing the check, not fixing the problem. We should still keep the issues open and update them accordingly.

adithyaov (approved):

harendra-kumar:
> This is more like bypassing the check, not fixing the problem. We should still keep the issues open and update them accordingly.

I will close them unless you have a plan on how to fix it.

adithyaov:
> I will close them unless you have a plan on how to fix it.

I don't, yet.

Issue #533, opened 2020-05-15
split the streamly package into multiple packages


harendra-kumar:
The streamly package is approaching 200 files and takes significant time to compile with more than 70 source modules, we may want to look at the possibility of splitting the package into smaller packages to make the ecosystem modular. Here is one approach:

1) `streamly`: remains as it is, a batteries included master package, including everything
2) `streamly-core`: Unfold/Stream/Pipe/Fold/Parser/Array/Time modules
3) `streamly-network`: the networking stuff
4) `streamly-filesystem`: the filesystem stuff
5) `streamly-unicode`: the unicode text processing stuff
6) `streamly-json`: the json parsing, serialization stuff
7) ... many other streamly packages like `streamly-zlib`, `streamly-statistics`

Programmers can mix and match the packages they want to include. This can possibly reduce the amount of time take to build, though streamly-core would still remain quite heavy.

We can continue to have all the packages in the same mega repository.

Bodigrim:
Would it be theoretically possible to have a hypothetical `streamly-core` depending on GHC boot libraries only? Or does it fundamentally entail something else?

harendra-kumar:
If you look at the dependencies of streamly, other than boot libs this is what we have:

* atomic-primops (for concurrent streams)
* monad-control (for concurrent streams)
* heaps (for concurrent streams)
* lockfree-queue (for concurrent streams)
* network

So if we just take out the concurrent streams and network stuff in separate packages, which is very much possible, then we will depend only boot libraries.

harendra-kumar:
We can try a feasibility prototype after the 0.8.0 release in next few weeks.

harendra-kumar:
I am working on separating a `streamly-core` package. There is a decision to be made about the safe resource management combinators `bracket` and `finally`. Those are important operations that would be useful in a core package. On the other hand they require a dependency on `monad-control`. Another class of operations that require a dependency on monad-control is time related combinators like `takeInterval` or `intervalsOf`. If we move these operations in another package then we could depend only on boot packages. 

@Bodigrim do you have any suggestions or opinions on this?

Bodigrim:
Take my words with a grain of salt, because I am only a potential user and not well acquainted with the library, but I'd slightly prefer only boot libraries in dependencies of `streamly-core`. The reason is that `monad-control` entails several more transitive dependencies, which are all well-maintained, but nevertheless elongate the critical path.

@hasufell what do you think?

harendra-kumar:
[...]

PR #1223, opened 2021-09-08
Move IsStream type class definition out of the low level StreamK module

harendra-kumar:
## Description

Move `IsStream` definition out of the low level `StreamK` module and push it up in a higher level IsStream/Type module. This is a step towards creating a streamly-core package devoid of any non-boot library dependencies, as mentioned in #533 . Also, this will facilitate moving to monomorphic stream types and modules in future, currently we use polymorphic streams based on IsStream type class.

The IsStream type class contains MonadAsync type in the consM method. This refactor allows lifting MonadAsync, therefore, dependency on monad-control out of the low level modules.

The main changes are in StreamK/Type.hs, StreamK.hs, IsStream/Type.hs, Stream/Serial.hs, the rest of the changes are mostly a fallout of these changes. Need to review these files and a few other files with important changes, the rest of the changes can be glanced over.

## Things to check

* Performance remains the same
* Documentation is not lost
* Exposed APIs do not change - cabal-diff
* Concurrent APIs remain concurrent

## Performance Notes

The `foldrS` benchmark has regressed by around 100 times, but the original perf numbers were good because of
a trivial foldr/build fusion. If we see `foldrSMap` which actually performs a map
using `foldrS` then it has improved by 2x.

```
Benchmark  default(0)(ns) default(1)(ns)
All.Prelude.Serial/o-1-space.mapping.foldrS 37636.90 3002730.00
All.Prelude.Serial/o-1-space.mapping.foldrSMap 6872930.00 2895490.00
```

`iterateState` benchmark now requires a SPECIALIZE pragma to specialize.

Other degradations, but the core looks the same for all these cases, may want to verify once again:

```
ParserD.serialWith 50% degradation, many
Parser serialWith/split_ 50%
Array.Foreign writeN/writeLastN/fromList 50% deg
```

Other than these there are no significant regressions, there are many improvements though.


src/Streamly/Internal/Data/Stream/IsStream/Type.hs
```diff
@@ -0,0 +1,698 @@
...
+-- parameterized by that stream.
+--
+-- XXX Since t is coercible we should be able to coerce k
+-- mkStream k = fromStream $ MkStream $ coerce k
+--
+-- | Build a stream from an 'SVar', a stop continuation, a singleton stream
+-- continuation and a yield continuation.
+{-# INLINE_EARLY mkStream #-}
+mkStream :: IsStream t
+    => (forall r. State K.Stream m a
```
adithyaov, line 260: I remember there being an issue or a comment about fixing `K.Stream` in `State`. No comment. Just stating it here.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
```diff
@@ -228,19 +228,6 @@ parse_ (PRD.Parser pstep initial extract) stream@(Stream step state) = do
 -- Specialized Folds
 ------------------------------------------------------------------------------
 
--- | Run a streaming composition, discard the results.
-{-# INLINE_LATE drain #-}
-drain :: Monad m => Stream m a -> m ()
--- drain = foldrM (\_ xs -> xs) (return ())
-drain (Stream step state) = go SPEC state
```
adithyaov, line 235: Why was this removed/moved?

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
```diff
@@ -471,7 +480,7 @@ sourceConcatMapId val n =
 {-# INLINE concatMapBySerial #-}
 concatMapBySerial :: Int -> Int -> Int -> IO ()
 concatMapBySerial outer inner n =
-    S.drain $ S.concatMapBy S.serial
+    S.drain $ S.concatMapWith S.serial
```
adithyaov, line 483: Are `concatMapWith` and `concatMapBy` same?

adithyaov (approved):
I went through them. Looks good. Did you tally and make sure you've not removed any API by mistake? You're using some tool I believe.

benchmark/Streamly/Benchmark/Data/Stream/StreamK.hs
harendra-kumar: concatMapBy is older name, later renamed to concatMapWith.

src/Streamly/Internal/Data/Stream/StreamD/Eliminate.hs
harendra-kumar: It is required in low level modules like instances of SerialT and other types.

harendra-kumar:
`cabal-diff` is the tool. There is one change to the exposed API, I will fix that. There are several changes to Internal APIs. I am attaching the diff.

[API diff of master with 0.8.0](https://github.com/composewell/streamly/files/7149354/diff.txt)

bin/test.sh
```diff
@@ -59,14 +59,14 @@ do
     # options with arguments
     --targets) shift; TARGETS=$1; shift ;;
     --with-compiler) shift; CABAL_WITH_COMPILER=$1; shift ;;
-    --cabal-build-options) shift; CABAL_BUILD_OPTIONS+=$1; shift ;;
-    --hpc-report-options) shift; HPC_REPORT_OPTIONS=$1; shift ;;
-    --rtsopts) shift; RTS_OPTIONS=$1; shift ;;
+    --cabal-build-options) shift; CABAL_BUILD_OPTIONS+=" $1"; shift ;;
```
adithyaov, line 62: I fixed this in another change. I'll undo that.

harendra-kumar:
I am going to squash and merge it.

PR #1229, opened 2021-09-10
Deriving Semigroup, Monoid, Num via Ap (Tee m a) b.

Icelandjack:


Icelandjack:
Nvm not supported by ghc 8

PR #1231, opened 2021-09-14
Add a CI that fails if line length exceeds 80

### LINKED ISSUE #1218: Add a CI that fails if line length exceeds 79
We can run something like this on all .hs files:

```
awk '{if (length($0) > 79) {print "Line length exceeds 79: " length($0); print $0; exit 1;}}'
```

We can add this in the hlint CI itself. If some files have too many issues to fix we can add it to hlintignore file. Ideally we should fix line lengths in at least those files that are hlint free so that we do not end up adding those files to the hlintignore.

rnjtranjan:
Fix #1218

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -391,8 +391,15 @@ jobs:
               bash -c "$PACKCHECK $BUILD" || exit 1
               echo "Checking trailing spaces..."
               count=$(find . -name "*.hs" -exec grep -H '\ $' {} \; | tee /dev/tty | wc -l)
-              count2=$(awk '{ if (length($0) > 80) {print "Line length exceeds 80 : " length($0); print $0; print  FILENAME;  } }' *.hs | tee /dev/tty | wc -l)
-              exit $((count+count2))
+              if $count > 0
+              then exit 1
+              fi
+              git diff > tmp_len
+              count2=$(awk '{ if (length($0) > 80) {print "Line length exceeds 80 : " length($0); print $0; } }' tmp_len | tee /dev/tty | wc -l)
+              rm tmp_len
+              if $count2 > 0
+              then exit 2
+              fi              
         - *save
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -418,4 +418,4 @@ workflows:
       #- coveralls-ghc-8_10_2:
       #    name: GHC 8.10.2 + inspection + coverage + Werror
       - hlint-src:
-          name: Hlint src + Trailing Spaces
+          name: Hlint src + Trailing Spaces + Column Length
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -394,12 +394,12 @@ jobs:
               if [ $count > 0 ]
               then exit 1
               fi
-              git diff > tmp_len
+              git diff master HEAD "*.hs" | grep '^+' | grep -v '^+++' > tmp_len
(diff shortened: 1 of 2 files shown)


harendra-kumar:
hlint is fixed in my refactoring commit.

.circleci/config.yml
```diff
@@ -390,7 +390,8 @@ jobs:
             command: |
               bash -c "$PACKCHECK $BUILD" || exit 1
               count=$(find . -name "*.hs" -exec grep -H '\ $' {} \; | tee /dev/tty | wc -l)
-              exit $count
+              count2=$(awk '{ if (length($0) > 80) {print "Line length exceeds 80 : " length($0); print $0; print  FILENAME;  } }' *.hs | tee /dev/tty | wc -l)
+              exit $((count+count2))
```
harendra-kumar, line 394: We should use separate failures for trailing spaces and for line length. How are we going to know which one failed?

harendra-kumar (changes requested):

.circleci/config.yml
```diff
@@ -390,7 +390,15 @@ jobs:
             command: |
               bash -c "$PACKCHECK $BUILD" || exit 1
               count=$(find . -name "*.hs" -exec grep -H '\ $' {} \; | tee /dev/tty | wc -l)
-              exit $count
+              if [ $count > 0 ]
+              then exit 1
+              fi
+              git diff > tmp_len
```
harendra-kumar, line 396: This won't work. The output will always be empty as there is nothing changed in the checked out tree. You need to do git diff between the current branch HEAD and the target branch in which the PR is to be merged. The current branch and target branch info should be available in some env vars.

Also you could just directly pipe the output of diff to the command below rather than putting it in a file first.

.circleci/config.yml
```diff
@@ -390,7 +390,15 @@ jobs:
...
+              count2=$(awk '{ if (length($0) > 80) {print "Line length exceeds 80 : " length($0); print $0; } }' tmp_len | tee /dev/tty | wc -l)
```
harendra-kumar, line 397: The git diff output has deleted lines, changed lines and other info. We need to grep only the lines with a leading `+`  (`grep '^+') and filter out the lines with a leading `+++` (grep -v `^+++`) from that.

harendra-kumar (changes requested):
Rebase, to resolve conflicts.

.circleci/config.yml
adithyaov: In github, `${{ github.event.pull_request.head.sha }}` is the pull request base commit.
```
git diff ${{ github.event.pull_request.head.sha }} HEAD 
```
will work for all pull requests as expected.

In circleci, I am unable to immediately find the env var that has the branch or the commit being compared to.
Some interesting env vars are,
- `CIRCLE_BRANCH`: Current branch name
- `CIRCLE_SHA1`: Current checkout latest hash

A trivial way if the merge branch is always origin/master is:
```
git diff origin/master HEAD 
```

You can probably use the github CI.
```
git diff ${{ github.event.pull_request.head.sha }} HEAD
```
should work as expected.

`dit diff` might produce lines that might be over 80 that we have no control of.
These might be the git info messages, separators, etc.
We need to filter only lines that start with a `+`.
We can probably solve the whole task with `sed`.

.circleci/config.yml
adithyaov: Ah, yes!

rnjtranjan:
ner PR opened: https://github.com/composewell/streamly/pull/1557

PR #1233, opened 2021-09-15
Make sequence concurrent

### LINKED ISSUE #1232: Make "sequence" concurrent
The sequence operation docs say it is concurrent but in reality it is not. That's because it is using "S.sequence" instead of using "K.sequence" which would have been concurrent.

```
{-# INLINE sequence #-}
sequence :: (IsStream t, MonadAsync m) => t m (m a) -> t m a
sequence m = fromStreamS $ S.sequence (toStreamS m)
```

Another way to define it concurrently would be to simply use `sequence = mapM id`. `mapM` is already concurrent, so this would be correct definition.

We can write a test case to check this. We already have a test case for `mapM` which can be reused to check for the concurrency of sequence as well.

rnjtranjan:
Fix #1232


test/Streamly/Test/Prelude/Concurrent.hs
```diff
@@ -164,6 +164,10 @@ concurrentOps constr desc eq t = do
         concurrentMapM constr eq $ \n mv stream ->
             t $ S.mapM (mvarSequenceOp mv n) stream
 
+    prop1 (desc <> " sequence") $
+        concurrentMapM constr eq $ \n mv stream ->
+            t $ S.sequence $ S.map (mvarSequenceOp mv n) stream
+
```
harendra-kumar, line 170: Can you commit this before the fix commit and ensure that the test fails?

harendra-kumar (commented):
Check the performance of sequence before and after.

test/Streamly/Test/Prelude/Concurrent.hs
rnjtranjan: Not sure about the comment. Can you elaborate on this?

rnjtranjan: ok got it. need to validate the test should fail without fix.

rnjtranjan: Instead of failing test case hangs forever.

rnjtranjan: After 10 min.
Prelude.Concurrent
  Stream concurrent operations
    default
      aheadly fromFoldableM
        +++ OK, passed 10 tests.
      aheadly unfoldrM
        +++ OK, passed 10 tests.
      aheadly mapM
        +++ OK, passed 10 tests.
*** Exception: thread blocked indefinitely in an MVar operation

harendra-kumar: That's a failure, its fine.

rnjtranjan:
Benchmark comparision:
Before:	 
Prelude.Serial/o-1-space
    mapping
      sequence: OK (0.23s)
        406 μs ±  30 μs,   0 B  allocated,   0 B  copied
		
After:
Prelude.Serial/o-1-space
    mapping
      sequence: OK (0.21s)
        399 μs ±  27 μs,   0 B  allocated,   0 B  copied

PR #1235, opened 2021-09-15
Update doctests of the Generate module

harendra-kumar:


src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
```diff
@@ -529,21 +542,18 @@ iterateMSerial step = fromStreamS . S.iterateM step
...
 -- the first element of the tuple in terms of the second.
 --
--- @
--- import Streamly.Internal.Data.Stream.IsStream as Stream
--- import System.IO.Unsafe (unsafeInterleaveIO)
---
--- main = do
---     Stream.mapM_ print $ Stream.mfix f
+-- >>> import Streamly.Internal.Data.Stream.IsStream as Stream
+-- >>> import System.IO.Unsafe (unsafeInterleaveIO)
```
rnjtranjan, line 546: These import can go to $setup

src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
```diff
@@ -574,16 +582,20 @@ fromFoldable :: (IsStream t, Foldable f) => f a -> t m a
...
--- @
+-- >>> fromFoldableM = Prelude.foldr Stream.consM Stream.nil
 --
 -- Construct a stream from a 'Foldable' containing monadic actions.
 --
--- @
--- drain $ fromSerial $ S.fromFoldableM $ replicateM 10 (threadDelay 1000000 >> print 1)
--- drain $ fromAsync  $ S.fromFoldableM $ replicateM 10 (threadDelay 1000000 >> print 1)
--- @
+-- >>> pr n = threadDelay 1000000 >> print n
```
rnjtranjan, line 589: "pr n = threadDelay 1000000 >> print n" is defined in multiple places can be defined one place.

src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
harendra-kumar: First one can go but the second one is very specific to this code so better placed here. The setup section is usually not visible to users which makes those imports a bit obscure.

src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
harendra-kumar: The problem is that if we define it at one place then the name has to be intuitive to know what it is doing. By redefining readers know immediately from the definition. I think we have defined it as "delay" in other modules but the problem is that we have another "delay" in prelude.

rnjtranjan (approved):

PR #1243, opened 2021-09-26
Add strip API to Array

rnjtranjan:


src/Streamly/Internal/Data/Array.hs
```diff
@@ -239,3 +240,30 @@ streamFold f arr = f (toStream arr)
...
+
+-- | Truncate the array at the beginning and end as long as the predicate
+-- holds true.
+strip :: (a -> Bool) -> Array a -> Array a
+strip p arr =
+    let lastIndex = length arr - 1
+        indexR = getIndexR lastIndex
+        indexL = getIndexL (min 0 lastIndex)
+    in if indexL > indexR
+       then cloneArray arr 0 0
```
harendra-kumar, line 252: Create a `nil` array API for an empty array.

src/Streamly/Internal/Data/Array.hs
```diff
@@ -239,3 +240,30 @@ streamFold f arr = f (toStream arr)
...
+       else cloneArray arr indexL (indexR - indexL + 1)
```
harendra-kumar, line 253: If there is nothing to strip or if the array is empty then we can return the original array itself.

src/Streamly/Internal/Data/Array.hs
```diff
@@ -239,3 +240,30 @@ streamFold f arr = f (toStream arr)
...
 getIndexUnsafe :: Array a -> Int -> a
 getIndexUnsafe = indexArray
+
+-- | Truncate the array at the beginning and end as long as the predicate
+-- holds true.
+strip :: (a -> Bool) -> Array a -> Array a
+strip p arr =
+    let lastIndex = length arr - 1
+        indexR = getIndexR lastIndex
+        indexL = getIndexL (min 0 lastIndex)
```
harendra-kumar, line 250: If indexR is < 0 then we do not need to determine indexL.

src/Streamly/Internal/Data/Array.hs
```diff
@@ -239,3 +240,30 @@ streamFold f arr = f (toStream arr)
...
+            case p (indexArray arr idx) of
+            True ->  getIndexR (idx - 1)
+            False -> idx
+
+    getIndexL idx
+        | idx < 0 = 0
+        | otherwise =
+            case p (indexArray arr idx) of
+            True -> getIndexL (idx + 1)
+            False -> idx
```
harendra-kumar, line 269: Won't indexR and indexL crash as we are not checking the bounds of the array?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array.hs
rnjtranjan: getIndexR idx we are checking the boundry .
 getIndexL idx will not be executed in case of getIndexR returns -1.

harendra-kumar:
We can add the same API to foreign array as well.

rnjtranjan (commented):
looks good.

PR #1244, opened 2021-09-26
implement overlay in Array APIs

rnjtranjan:


src/Streamly/Internal/Data/Array.hs
```diff
@@ -239,3 +241,14 @@ streamFold f arr = f (toStream arr)
 {-# INLINE getIndexUnsafe #-}
 getIndexUnsafe :: Array a -> Int -> a
 getIndexUnsafe = indexArray
+
+-- | Update multiple indices of an array.
+overlay :: (PrimMonad m) => Array a -> SerialT m (Int, a) -> m (Array a)
+overlay arr m = do
```
harendra-kumar, line 247: Instead of `overlay`, we can use the same API for this purpose as we are using in `src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs` i.e. `putIndices`.

Can you implement this on the lines `putIndices` with the same signature?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array.hs
```diff
@@ -70,6 +71,7 @@ import qualified Streamly.Internal.Data.Ring as RB
 import qualified Streamly.Internal.Data.Stream.StreamD as D
 
 import Data.Primitive.Array hiding (fromList, fromListN)
+import Control.Monad.Primitive
```
harendra-kumar, line 74: Import symbols explicitly instead of importing the whole module.

src/Streamly/Internal/Data/Array.hs
```diff
@@ -313,3 +315,24 @@ strip p arr =
...
     getIndexL idx = if p (indexArray arr idx) then getIndexL (idx + 1) else idx
+
+-- | Write an input stream of (index, value) pairs to an array. Throws an
+-- error if any index is out of bounds.
+--
+-- /Pre-release/
+{-# INLINE putIndices #-}
+putIndices :: PrimMonad m
+    => Array a -> Fold m (Int, a) ()
+putIndices arr = FL.mkFoldM step initial extract
```
harendra-kumar, line 326: Can't we use `FL.foldlM'`?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array.hs
```diff
@@ -71,7 +71,7 @@ import qualified Streamly.Internal.Data.Ring as RB
 import qualified Streamly.Internal.Data.Stream.StreamD as D
 
 import Data.Primitive.Array hiding (fromList, fromListN)
-import Control.Monad.Primitive
+import Control.Monad.Primitive (PrimMonad)
```
harendra-kumar, line 74: This import should come after line 57.

harendra-kumar (changes requested):
One small suggestion.
Squash the commits.

PR #1245, opened 2021-09-27
Optimize and use StreamK.mergeBy as default

harendra-kumar:
* Optimize StreamK.mergeBy
* Use StreamK.mergeBy as default
* Add a separate combinator to use StreamD.mergeBy

### DIFF
--- a/src/Streamly/Internal/Data/Stream/StreamK.hs
+++ b/src/Streamly/Internal/Data/Stream/StreamK.hs
@@ -789,30 +789,134 @@ zipWithM f = go
 mergeByM :: Monad m =>
     (a -> a -> m Ordering) -> Stream m a -> Stream m a -> Stream m a
 mergeByM cmp = go
-    where
-    go mx my = mkStream $ \st yld sng stp -> do
-        let mergeWithY a ra =
-                let stop2 = foldStream st yld sng stp mx
-                    single2 b = do
-                        r <- cmp a b
-                        case r of
-                            GT -> yld b (go (a `cons` ra) nil)
-                            _  -> yld a (go ra (b `cons` nil))
-                    yield2 b rb = do
-                        r <- cmp a b
-                        case r of
-                            GT -> yld b (go (a `cons` ra) rb)
-                            _  -> yld a (go ra (b `cons` rb))
-                 in foldStream st yield2 single2 stop2 my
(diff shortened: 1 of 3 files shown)


src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -789,30 +789,134 @@ zipWithM f = go
...
+                    GT -> yld y (goX x mx ry)
+                    _  -> yld x (goY mx y ry)
+         in foldStream st yield single stop my
+
+    goY0 mx y = mkStream $ \st yld sng _ -> do
+        let stop = sng y
+            single x = do
+                r <- cmp x y
+                case r of
+                    GT -> yld y mx
```
rnjtranjan, line 834: It should be GT -> yld y (fromPure x)

adithyaov, line 834:
```suggestion
                    GT -> yld y (fromPure x)
```
Won't it run the the effect twice if `mx` is given?

harendra-kumar: Good catch!

src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -789,30 +789,134 @@ zipWithM f = go
...
+                case cmp x y of
+                    GT -> yld y (goX x mx ry)
+                    _  -> yld x (goY mx y ry)
+         in foldStream st yield single stop my
+
+    goY0 mx y = mkStream $ \st yld sng _ -> do
+        let stop = sng y
+            single x = do
+                case cmp x y of
+                    GT -> yld y mx
```
adithyaov, line 901:
```suggestion
                    GT -> yld y (fromPure x)
```

adithyaov (changes requested):
Small bug needs to be fixed

src/Streamly/Internal/Data/Stream/StreamK.hs
harendra-kumar: This is just a copy of the previous, so any change in that has to be done here as well. I could not find a good way to avoid the copy other than using CPP.

adithyaov (approved):

PR #1246, opened 2021-09-27
Fix/enhance benchmarks for concatPairsWith

harendra-kumar:
* Add a benchmark for mergeByMFused
* Move interleaving benchmarks to WSerial module
* Rename some wserial benchmarks
* Add a Merge module for mergeBy/sorting benchmarks
* Add concatPairsWithZip to ZipSerial module


adithyaov (approved):
Didn't look too deep into it but looks good overall.

PR #1257, opened 2021-10-05
Implementation of groupByRollingEither

rnjtranjan:
Fix for #1256


harendra-kumar:
I will try it out in `sort`.

Issue #1254, opened 2021-09-29
Add a strip API for foreign Array.


rnjtranjan:

adithyaov:
1. What does `strip` do? What would be it's signature? What would you strip for a general array? What is the assumption on the elements?

adithyaov:
https://github.com/composewell/streamly/issues/1215 supersedes this issue.

PR #1260, opened 2021-10-06
Add strip API to  Foreign Array

rnjtranjan:
Fix for #1254

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
+++ b/benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
@@ -61,6 +61,22 @@ o_n_space_serial_marray value array =
         $ MArray.partitionBy (> (value `div` 2))
     ]
 
+o_1_space_serial_marray :: Int -> MArray.Array Int -> [Benchmark]
+o_1_space_serial_marray value array =
+    [ benchIO "strip (< 0)" (const array)
+        $ MArray.strip (< 0)
+    , benchIO "strip (> 0)" (const array)
+        $ MArray.strip (> 0)
+    , benchIO "strip (< value/2)" (const array)
+        $ MArray.strip (< (value `div` 2))
+    , benchIO "strip (> value/2)" (const array)
+        $ MArray.strip (> (value `div` 2))
+    , benchIO "strip (odd)" (const array)
+        $ MArray.strip odd
+    , benchIO "strip (even)" (const array)
+        $ MArray.strip even
+    ]
+
 -------------------------------------------------------------------------------
 -- Driver
 -------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1597,3 +1599,37 @@ cmp arr1 arr2 =
...
+strip eq arr@Array{..} =
+    let p = unsafeForeignPtrToPtr aStart
+        q = aEnd
+        len = length arr
+        st = getStart p len
+        end = do
+            st' <- st
+            if st' == q
+            then return q
+            else getLast $ q `plusPtr` negate (sizeOf (undefined :: a))
```
harendra-kumar, line 1614: Use the macros from ArrayMacros.h to manipulate the pointers.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1597,3 +1599,37 @@ cmp arr1 arr2 =
...
+    in go st end
```
harendra-kumar, line 1615: `go` is not a recursive function. We can inline the definition right here, that would be more readable.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1597,3 +1599,37 @@ cmp arr1 arr2 =
 instance NFData (Array a) where
     {-# INLINE rnf #-}
     rnf Array {} = ()
+
+strip :: forall a m. (Storable a, MonadIO m) =>
+    (a -> Bool) -> Array a -> m (Array a)
+strip eq arr@Array{..} =
+    let p = unsafeForeignPtrToPtr aStart
```
harendra-kumar, line 1606: We will need a `touch` to ensure the pointer is not garbage collected.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1597,3 +1599,37 @@ cmp arr1 arr2 =
...
+        q = aEnd
+        len = length arr
+        st = getStart p len
+        end = do
+            st' <- st
```
harendra-kumar, line 1611: The action `st` is being called at two places, we are effectively searching for start twice, instead we can call it once and use the result.

harendra-kumar (changes requested):
This is based on very old commit, the array type has changed since. Can you rebase and fix?

Can you write a basic test for it?

harendra-kumar:
Can you add a benchmark as well?

rnjtranjan (commented):
Looks good.

PR #1267, opened 2021-10-09
Simplify many/manyPost

harendra-kumar:
Minor simplifications, no functional change.


adithyaov:
Did you check for any performance change?

PR #1279, opened 2021-10-17
Some refactoring changes

harendra-kumar:
See individual commits. Review commit by commit.


harendra-kumar:
There should not be any functionality or perf change in this set of changes.

PR #1305, opened 2021-10-28
Implement scan for folds

rnjtranjan:
Fix for # 1303

### DIFF
--- a/src/Streamly/Internal/Data/Fold.hs
+++ b/src/Streamly/Internal/Data/Fold.hs
@@ -380,15 +380,50 @@ transform (Pipe pstep1 pstep2 pinitial) (Fold fstep finitial fextract) =
 
     extract (Tuple' _ fs) = fextract fs
 
--- See StreamD/Transform.scanOnce for implementing this.
---
 -- | Scan the input of a 'Fold' to change it in a stateful manner using another
 -- 'Fold'.
---
--- /Unimplemented/
+-- XXX if any Fold is done assuming finish the resulting Fold?
+-- /Pre-release/
 {-# INLINE scan #-}
-scan :: Fold m a b -> Fold m b c -> Fold m a c
-scan = undefined
+scan :: Monad m => Fold m a b -> Fold m b c -> Fold m a c
+scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
+    Fold step initial extract
+
+    where
+
+    initial = do
+        r1 <- initialL
+        r2 <- initialR
+        case r1 of
+            Done b1 -> return $ case r2 of
+                Done b2 -> Done b2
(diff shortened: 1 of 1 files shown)


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -384,11 +384,42 @@ transform (Pipe pstep1 pstep2 pinitial) (Fold fstep finitial fextract) =
...
+scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
+    Fold step initial extract
+
+    where
+
+    initial = do
+        r1 <- initialL
+        r2 <- initialR
+        case r1 of
+            Done _ -> case r2 of
```
harendra-kumar, line 400: Why are we discarding the result of first fold? We have to give it to the second fold.

harendra-kumar: In fact the code for `initial` is usually the same as the code for `step` if you replace the "stepL (fromJust s1) x` with `initial`. So you can create a common function and pass the `action` to it. In case of initial it would be `initial` and in case of `step` it would be `stepL (fromJust s1) x` etc.

harendra-kumar (changes requested):
Add benchmark and test too.

rnjtranjan:
Issue #1313  is created for test and benchmark.

src/Streamly/Internal/Data/Fold.hs
```diff
@@ -391,39 +390,31 @@ scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
...
-
-    extract (_, s2, _) = extractR (fromJust s2)
+        r <- initialR
+        case r of
+            Partial sR -> runStep initialL sR
+            Done b -> return $ Done b
+
+    step (sL, sR) x = runStep (stepL sL x) sR
+
+    extract = extractR . snd
```
rnjtranjan, line 417: extract for initial case (for stream scan very first value when both the Folds will be in the initial state).
extract should be something:
extract (Just s1, s2) = do        
        l1 <- extractL s1
        r2 <- stepR (fromJust s2) l1
        case r2 of
            Partial fs2 -> extractR fs2
            Done b2 -> return b2

harendra-kumar: We have already done all this in `initial` when we called `runStep`. See this code in `runStep`:

```
Partial sL -> do
                !b <- extractL sL
                rR <- stepR sR b
                return
                    $ case rR of
                        Partial sR1 -> Partial (sL, sR1)
                        Done bR -> Done bR
```

All we need to do in extract is to extract `sR1`.

rnjtranjan: kool.

Issue #1304, opened 2021-10-28
Implement scan for Unfold


harendra-kumar:
In Streamly.Internal.Data.Unfold we have:

```
-- See StreamD.scanOnce for implementing this.
--
-- | Scan the output of an 'Unfold' to change it in a stateful manner.
--
-- /Unimplemented/
{-# INLINE_NORMAL scan #-}
scan :: Fold m b c -> Unfold m a b -> Unfold m a c
scan = undefined
```

harendra-kumar:
Fixed by #1306

PR #1306, opened 2021-10-28
Implement scan for Unfold #1304

rnjtranjan:


harendra-kumar (changes requested):
Can we add a benchmark and test as well? We should already have benchmark and test for scan, just need to adapt it.

rnjtranjan:
Issue #1312 is created for test and benchmark.

rnjtranjan (commented):
looks good.

PR #1309, opened 2021-10-30
Mutable array module cleanup

harendra-kumar:


harendra-kumar:
Same as PR#1299 which got closed due to branch deletion.

PR #1317, opened 2021-11-06
Fix missing touch

harendra-kumar:
Adding a touchForeignPtr led to significant regressions. Fixing that required a bigger change to use a simpler type in the array instead of using ForeignPtr.

### DIFF
--- a/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
+++ b/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE UnboxedTuples #-}
 -- |
 -- Module      : Streamly.Internal.Data.Array.Foreign.Mut.Type
 -- Copyright   : (c) 2020 Composewell Technologies
@@ -16,6 +17,12 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Type
     -- * Type
     -- $arrayNotes
       Array (..)
+    , ArrayContents
+    , arrayToFptrContents
+    , fptrToArrayContents
+    , unsafeWithArrayContents
+    , nilArrayContents
+    , touch
 
     -- * Constructing and Writing
     -- ** Construction
@@ -196,6 +203,10 @@ where
 
 #include "inline.hs"
 
+#ifdef USE_C_MALLOC
+#define USE_FOREIGN_PTR
+#endif
+
 import Control.Exception (assert)
 import Control.DeepSeq (NFData(..))
 import Control.Monad (when, void)
@@ -207,17 +218,29 @@ import Data.Semigroup (Semigroup(..))
 #endif
 import Data.Word (Word8)
 import Foreign.C.Types (CSize(..), CInt(..))
-import Foreign.ForeignPtr (touchForeignPtr, castForeignPtr)
 import Foreign.ForeignPtr.Unsafe (unsafeForeignPtrToPtr)
+#ifndef USE_FOREIGN_PTR
+import Foreign.Marshal.Alloc (mallocBytes)
+#endif
 import Foreign.Ptr (plusPtr, minusPtr, castPtr, nullPtr)
 import Foreign.Storable (Storable(..))
-import GHC.Base (build)
+import GHC.Base
+    ( build, touch#, IO(..), byteArrayContents#
+    , Int(..), newAlignedPinnedByteArray#
+    )
+#ifndef USE_FOREIGN_PTR
+import GHC.Base (RealWorld, MutableByteArray#)
+#endif
 #if __GLASGOW_HASKELL__ < 802
 #define noinline
 #else
 import GHC.Base (noinline)
 #endif
-import GHC.ForeignPtr (ForeignPtr(..))
+import GHC.Exts (unsafeCoerce#)
+import GHC.ForeignPtr (ForeignPtr(..), ForeignPtrContents(..))
+#ifdef USE_C_MALLOC
+import GHC.ForeignPtr (mallocForeignPtrAlignedBytes)
+#endif
 import GHC.Ptr (Ptr(..))
 
 import Streamly.Internal.BaseCompat
@@ -225,9 +248,9 @@ import Streamly.Internal.Data.Fold.Type (Fold(..))
 import Streamly.Internal.Data.Producer.Type (Producer (..))
 import Streamly.Internal.Data.SVar.Type (adaptState)
 import Streamly.Internal.Data.Unfold.Type (Unfold(..))
-
 import Streamly.Internal.System.IO
     (defaultChunkSize, arrayPayloadSize, unsafeInlineIO)
+import System.IO.Unsafe (unsafePerformIO)
 
 #ifdef DEVBUILD
 import qualified Data.Foldable as F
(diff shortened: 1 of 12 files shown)


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -300,12 +308,39 @@ memcmp p1 p2 len = do
     r <- c_memcmp p1 p2 (fromIntegral len)
     return $ r == 0
 
--- | Same as unsafeWithForeignPtr but lifted to any MonadIO monad.
-{-# INLINE unsafeWithForeignPtrM #-}
-unsafeWithForeignPtrM :: MonadIO m => ForeignPtr a -> (Ptr a -> m b) -> m b
-unsafeWithForeignPtrM fo f = do
-  r <- f (unsafeForeignPtrToPtr fo)
-  liftIO $ touchForeignPtr fo
```
adithyaov, line 300: Why remove this?

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -244,20 +259,13 @@ import qualified Streamly.Internal.Data.Fold.Type as FL
...
 import Prelude hiding
     (length, foldr, read, unlines, splitAt, reverse, truncate)
 
-#if MIN_VERSION_base(4,10,0)
-import Foreign.ForeignPtr (plusForeignPtr)
-#else
-import GHC.Base (Int(..), plusAddr#)
-import GHC.ForeignPtr (ForeignPtr(..))
-plusForeignPtr :: ForeignPtr a -> Int -> ForeignPtr b
-plusForeignPtr (ForeignPtr addr c) (I# d) = ForeignPtr (plusAddr# addr d) c
```
adithyaov, line 250: Is there something incorrect here?

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -300,12 +308,39 @@ memcmp p1 p2 len = do
...
+fptrToArrayContents :: ForeignPtrContents -> ArrayContents
+arrayToFptrContents :: ArrayContents -> ForeignPtrContents
+#ifdef USE_FOREIGN_PTR
+fptrToArrayContents = ArrayContents
+arrayToFptrContents (ArrayContents contents) = contents
+#else
+fptrToArrayContents (PlainPtr mbarr) = ArrayContents mbarr
+fptrToArrayContents _ = error "Unsupported foreign ptr"
+arrayToFptrContents (ArrayContents contents) = PlainPtr contents
+#endif
```
adithyaov, line 335: Can you please elaborate on this?

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -634,6 +699,11 @@ arrayChunkBytes = 1024
 -- Snoc
 -------------------------------------------------------------------------------
 
+{-# INLINE touch #-}
+touch :: ArrayContents -> IO ()
+touch (ArrayContents contents) =
```
adithyaov, line 704: Add this on the top of the file? Define before use?

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -343,8 +378,8 @@ data Array a =
     Storable a =>
 #endif
     Array
-    { arrContents :: {-# UNPACK #-} !ForeignPtrContents -- ^ first address
-    , arrStart :: {-# UNPACK #-} !(Ptr a)
+    { arrContents :: UNPACKIF !ArrayContents
+    , arrStart :: {-# UNPACK #-} !(Ptr a)      -- ^ first address
```
adithyaov: Why this change?

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1155,24 +1177,24 @@ arraysOf n (D.Stream step state) =
...
-            D.Stop -> return $ D.Skip (GroupYield start end bound GroupFinish)
+                    then D.Skip (GroupYield contents start end' bound (GroupStart s))
+                    else D.Skip (GroupBuffer s contents start end' bound)
+            D.Skip s -> return $ D.Skip (GroupBuffer s contents start end bound)
+            D.Stop -> return $ D.Skip (GroupYield contents start end bound GroupFinish)
 
-    step' _ (GroupYield start end bound next) =
-        return $ D.Yield (Array start end bound) next
+    step' _ (GroupYield contents start end bound next) =
+        return $ D.Yield (Array contents start end bound) next
```
adithyaov, line 1271: Don't we ever touch foreign ptr ever here? (Question for the original code)

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1355,7 +1391,7 @@ toListFB c n Array{..} = go (unsafeForeignPtrToPtr aStart)
         -- XXX
         let !x = unsafeInlineIO $ do
                     r <- peek p
-                    touchForeignPtr aStart
+                    touch arrContents
```
adithyaov, line 1452: `touchFPContents` vs `touch`.

src/Streamly/Internal/Data/Array/Foreign/Type.hs
```diff
@@ -223,7 +226,8 @@ fromPtr n ptr = unsafeInlineIO $ do
     fptr <- newForeignPtr_ ptr
     let end = ptr `plusPtr` n
     return $ Array
-        { aStart = fptr
+        { arrStart = ptr
+        , arrContents = undefined
```
adithyaov: What happens when you `touch` `undefined`?
If we use `fromPtr`, will we be able to use the result in other combinators?

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -437,8 +437,7 @@ replacementChar = '\xFFFD'
 decodeUtf8WithD :: Monad m
     => CodingFailureMode -> Stream m Word8 -> Stream m Char
 decodeUtf8WithD cfm (Stream step state) =
-    let A.Array p _ = utf8d
-        !ptr = unsafeForeignPtrToPtr p
+    let A.Array contents ptr _ = utf8d
```
adithyaov: Perform `!ptr = p` like you did above? Is that necessary?

adithyaov (approved):
I did not see any obvious error. LGTM

src/Streamly/Internal/Data/Array/Foreign/Type.hs
```diff
@@ -139,7 +138,11 @@ data Array a =
     Storable a =>
 #endif
     Array
-    { arrContents :: {-# UNPACK #-} !ForeignPtrContents -- first address
+    { arrContents ::
+#ifndef USE_FOREIGN_PTR
```
adithyaov, line 142: Don't you need to initialize the flag accordingly here?

harendra-kumar: Remains undefined by default.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
harendra-kumar: Looks ok to me.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
harendra-kumar: Moved.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
harendra-kumar: Looks like a bug. But may not occur in practice depending on how the array is used. We can try adding a touch and check if it degrades perf.

PR #1326, opened 2021-11-09
Add Streamly.Internal.Data.Array.Mut.Type

adithyaov:


src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,527 @@
...
+-------------------------------------------------------------------------------
+-- Array Data Type
+-------------------------------------------------------------------------------
+
+data Array a =
+    Array
+        { arrContents# :: MutableArray# RealWorld a
+        , arrStart :: {-# UNPACK #-}!Int
+        , arrLen :: {-# UNPACK #-}!Int
+        , arrSpace :: {-# UNPACK #-}!Int
```
harendra-kumar, line 189: can call it arrFree - space might mean used space, total space or free space.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+    arr1 <- newArray i
+    let !newLen@(I# newLen#) = min i (arrLen arr)
+        !(I# arrS#) = arrStart arr
+        !(I# arr1S#) = arrStart arr1
+        arrC# = arrContents# arr
+        arr1C# = arrContents# arr1
+    liftInIO
+        $ \s# ->
+              let s1# = copyMutableArray# arrC# arrS# arr1C# arr1S# newLen# s#
+               in (# s1#, arr1 {arrLen = newLen, arrFree = i} #)
```
harendra-kumar, line 303: Is `arrFree = i` correct? When we are reallocating the free index should remain the same as earlier. `arrFree = i` means nothing is free.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+
+-- | Really really unsafe, appends the element into the first array, may
+-- cause silent data corruption or if you are lucky a segfault if the index
+-- is out of bounds.
+--
+-- /Internal/
+{-# INLINE snocUnsafe #-}
+snocUnsafe :: MonadIO m => Array a -> a -> m (Array a)
+snocUnsafe arr@Array {..} a = do
+    assert (arrLen + 1 <= arrFree) (return ())
```
harendra-kumar, line 335: Assuming `arrFree` indicates the free index, this should be `arrLen <= arrFree` as the `arrLen` index is beyond the last index and it should be free.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+-- in bytes determined by the @sizer oldSize@ function, where @oldSize@ is the
+-- original size of the array.
+--
+-- Note that the returned array may be a mutated version of the original array.
+--
+-- /Pre-release/
+{-# INLINE snocWith #-}
+snocWith :: MonadIO m => (Int -> Int) -> Array a -> a -> m (Array a)
+snocWith sizer arr@(Array {..}) x = do
+    if arrLen + 1 <= arrFree
```
harendra-kumar, line 359: `arrLen <= arrFree`?

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+        $ \s# ->
+              let !(I# i#) = arrStart + n
+               in readArray# arrContents# i# s#
+
+-- | /O(1)/ Lookup the element at the given index. Index starts from 0.
+--
+{-# INLINE getIndex #-}
+getIndex :: MonadIO m => Array a -> Int -> m a
+getIndex arr@(Array {..}) i =
+    if i < arrLen && i > 0
```
harendra-kumar, line 403: Bug: `i >= 0`.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+--
+-- We can try this if the unfold has any performance issues.
+{-# INLINE_NORMAL toStreamD #-}
+toStreamD :: MonadIO m => Array a -> D.Stream m a
+toStreamD arr@(Array {..}) =
+    D.mapM (getIndexUnsafe arr) $ D.enumerateFromToIntegral 0 (arrLen - 1)
+
+{-# INLINE toStreamK #-}
+toStreamK :: MonadIO m => Array a -> K.Stream m a
+toStreamK arr@(Array {..}) = K.mapM (getIndexUnsafe arr) $ K.unfoldr step 0
```
harendra-kumar, line 471: Would it be faster if we K.unfoldrM instead of using mapM on unfoldr? depends on whether mapM can fuse, I guess it doesn't for CPS streams.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+    {-# INLINE_LATE step #-}
+    step Array {..}
+        | assert (arrLen >= 0) (arrLen == 0) = return D.Stop
+    step arr@(Array {..}) = do
+        x <- getIndexUnsafe arr 0
+        return $ D.Yield x $ arr {arrStart = arrStart + 1, arrLen = arrLen - 1}
+
+-- | Unfold an array into a stream.
+--
+-- @since 0.7.0
```
harendra-kumar, line 532: Remove this.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+producer = Producer step return return
+
+    where
+
+    {-# INLINE_LATE step #-}
+    step Array {..}
+        | assert (arrLen >= 0) (arrLen == 0) = return D.Stop
+    step arr@(Array {..}) = do
+        x <- getIndexUnsafe arr 0
+        return $ D.Yield x $ arr {arrStart = arrStart + 1, arrLen = arrLen - 1}
```
harendra-kumar, line 528: May be better to use a simpler custom data type for the state in this loop. Would be easier to fuse in more cases.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+
+-- | @newArray count@ allocates an empty array that can hold 'count' items.
+--
+-- /Pre-release/
+{-# INLINE newArray #-}
+newArray :: forall m a. MonadIO m => Int -> m (Array a)
+newArray n@(I# n#) =
+    liftInIO
+        $ \s# ->
+              case newArray# n# bottomElement s# of
```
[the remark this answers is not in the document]
adithyaov: `newUninitializedArray#` would be the same? What lifted element would it point to?

src/Streamly/Internal/Data/Array/Mut/Type.hs
adithyaov: 1. I shouldn't have renamed it to `arrFree`. I don't know what I was thinking at that point of time.
2. `arrFree` (`arrSpace` initially) is the true length of the array contents itself. The first unacceptable index.

src/Streamly/Internal/Data/Array/Mut/Type.hs
adithyaov: Both are incorrect, fixed now.

src/Streamly/Internal/Data/Array/Mut/Type.hs
adithyaov: See comment above

src/Streamly/Internal/Data/Array/Mut/Type.hs
adithyaov: Except one field all other fields are in use.

src/Streamly/Internal/Data/Array/Mut/Type.hs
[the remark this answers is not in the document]
adithyaov: All previous slices will become invalid if `resizeMutableByteArray#` is used.

src/Streamly/Internal/Data/Array/Mut/Type.hs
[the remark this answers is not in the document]
harendra-kumar: The point is, we should not be writing anything to it. Currently we are explicitly writing undefined to it which looks like a waste of time.

src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -0,0 +1,535 @@
...
+                              let (a1, b) = f a
+                                  s2# = writeArray# arrContents# n# a1 s1#
+                               in (# s2#, b #)
+
+-- | Modify a given index of an array using a modifier function.
+--
+-- /Pre-release/
+modifyIndex :: MonadIO m => Array a -> Int -> (a -> (a, b)) -> m b
+modifyIndex arr@(Array {..}) i f = do
+    if i < arrLen && i > 0
```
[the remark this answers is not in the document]
harendra-kumar: The condition is still the same.

src/Streamly/Internal/Data/Array/Mut/Type.hs
harendra-kumar: Yes, I know, and sometimes that one field itself makes a difference.

harendra-kumar (changes requested):

harendra-kumar (approved):
Can we add some more tests to arrays so that the kind of bugs that we caught in the review can be caught by the tests?

Remember to squash the commits before merge.

adithyaov:
See https://github.com/composewell/streamly/issues/1380

PR #1328, opened 2021-11-10
Fsevent testing

harendra-kumar:
Windows and Darwin test modules are done, Linux is still to be done.


test/Streamly/Test/FileSystem/Event/Common.hs
```diff
@@ -0,0 +1,294 @@
...
+      , [ (dir1, dirEvent Event.isDeleted) ]
+      )
+    , ( desc <> ", isMoved"
+      , createDir1
+      , \fp -> renameDirectory (fp </> dir1) (fp </> dir2)
+      , [ (dir1, dirEvent Event.isMoved)
+        , (dir2, dirEvent Event.isMoved)
+        ]
+      )
+    ]
```
rnjtranjan, line 226: Do we test for events like "root unwatched and root moved"

harendra-kumar: That would go in rootTests if we do. There are more tests that can be added. You can add more once this PR is committed.

PR #1346, opened 2021-11-23
More changes for the release

adithyaov:


docs/CONTRIBUTORS.md
```diff
@@ -4,6 +4,15 @@ please see https://github.com/composewell/streamly/issues.
 Use `git shortlog -sn tag1...tag2` on the git repository to get a list of
 contributors between two repository tags.
 
+## 0.8.1
+
+* Harendra Kumar
+* Adithya Kumar
+* Ranjeet Kumar Ranjan
+* Julian Ospald
+* Ranjeet Ranjan
```
harendra-kumar, line 13: Repeated name.

configure
```diff
@@ -1,12 +1,11 @@
 #! /bin/sh
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.71 for streamly 0.8.0.
+# Generated by GNU Autoconf 2.69 for streamly 0.8.1.
```
harendra-kumar, line 3: Why use an older autconf version? I guess if we use the same version, we will not have many changes in this file.

adithyaov: Ah yes.

harendra-kumar (approved):

Issue #1307, opened 2021-10-28
Event.watchPaths throws: addToWatch: /some/example.csv/: inappropriate type (Not a directory)


Unisay:
**Expected behavior:**

As [documentation](https://streamly.composewell.com/haddocks/streamly-0.8.0/Streamly-Internal-FileSystem-Event-Linux.html#v:watchPathsWith) says:
> The paths could be files or directories. 

**Actual behavior:**
```haskell
  filePath <- toUtf8 "/some/example.csv"
  Event.watchPaths (pure filePath)
```
throws:
```
addToWatch: /some/example.csv/: inappropriate type (Not a directory)
```

harendra-kumar:
@Unisay sorry for the delay in answering this. We missed this issue somehow.

The behavior of fsevents is highly platform specific. Which platform are seeing this behavior on? We have to test this particular aspect better and correct the documentation if necessary or fix the code.

BTW We have made several changes to the fsevents subsystem recently and also created a common fsevents module that tries to capture the platform independent behavior. We are going to release 0.8.1 soon which will have these changes.

harendra-kumar:
A file root works on macOS, I added tests for it in the darwin test suite a while ago. There are TODOs for the same in Linux and Windows test suites, let's add those and fix if anything needs to be fixed or documented as part of this issue. 

The working darwin tests are [here](https://github.com/composewell/streamly/blob/fb5cc5401cfa2cdc4f03aa71b82e277d06c4b2d8/test/Streamly/Test/FileSystem/Event/Darwin.hs#L68:L85).

For Linux failing tests are [here](https://github.com/composewell/streamly/blob/fb5cc5401cfa2cdc4f03aa71b82e277d06c4b2d8/test/Streamly/Test/FileSystem/Event/Linux.hs#L150:L152).

For Windows we have a [TODO here](https://github.com/composewell/streamly/blob/fb5cc5401cfa2cdc4f03aa71b82e277d06c4b2d8/test/Streamly/Test/FileSystem/Event/Windows.hs#L38:L40). Should be easy to copy the tests from darwin test suite.

We need to fix the Linux code and get the Windows tests working.

harendra-kumar:
@Unisay let me know if you would be in a position to contribute to this.

Unisay:
Sure, what can I do to help with this issue?

harendra-kumar:
[...]

PR #1357, opened 2021-11-27
Fix inappropriate type (Not a directory) error in Linux for File path…

rnjtranjan:
…s in recursive mode

Fix for #1307

### DIFF
--- a/src/Streamly/Internal/FileSystem/Event/Windows.hs
+++ b/src/Streamly/Internal/FileSystem/Event/Windows.hs
@@ -447,16 +447,15 @@ closePathHandleStream = S.mapM_ (\(h, _, _) -> closeHandle h)
 -- based on the configured code page?
 -- https://docs.microsoft.com/en-us/windows/uwp/design/globalizing/use-utf8-code-page
 --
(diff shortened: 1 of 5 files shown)


src/Streamly/Internal/FileSystem/Event/Windows.hs
```diff
@@ -12,7 +12,7 @@
 --
 -- =Overview
 --
--- Use 'watchTrees'or 'watchPaths' with a list of file system dir paths you 
+-- Use 'watchTrees'or 'watchPaths' with a list of file system dir paths you
```
harendra-kumar, line 15: We no longer have these APIs. We have `watch` and `watchRecursive`.

harendra-kumar (changes requested):
Can you also enable the commented test for the case when root path is a file?

rnjtranjan:
> 

For windows test for the case when root path is a file is not valid as windows API doesn't support it.
For Linux nested file create event times out due to the event is lost.

rnjtranjan (commented):
Looks good

PR #1359, opened 2021-11-28
Implement some undefined methods

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -668,10 +716,12 @@ roundUpLargeArray size =
...
     else size
 
-{-
+
+
+
 roundUpToPower2 :: Int -> Int
-roundUpToPower2 = undefined
--}
+roundUpToPower2 x = 2 ^ ceiling (logBase 2 (fromIntegral x))
```
harendra-kumar, line 723: We should do this smartly using bitwise operations if possible.

PR #1362, opened 2021-12-01
Immutable fromUnsafeForeignPointer + nullPtr check

adithyaov:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -403,6 +403,8 @@ fromForeignPtrUnsafe ::
     Storable a =>
 #endif
     ForeignPtr a -> Ptr a -> Ptr a -> Array a
+fromForeignPtrUnsafe (ForeignPtr start _) _ _
+    | Ptr start == nullPtr = Array nilArrayContents nullPtr nullPtr nullPtr
```
harendra-kumar, line 407: We can add a `nil` array API as we have in `src/Streamly/Internal/Data/Array/Foreign/Type.hs`.

harendra-kumar (approved):

PR #1364, opened 2021-12-01
Enhance the ring array

adithyaov:


harendra-kumar:
Can you compare the perf benchmarks?

PR #1370, opened 2021-12-04
Comment out incorrect test

harendra-kumar:
The test is unpredictable as we create the watch dir before starting the
watch, we may miss the event.

### DIFF
--- a/test/Streamly/Test/FileSystem/Event/Darwin.hs
+++ b/test/Streamly/Test/FileSystem/Event/Darwin.hs
@@ -57,8 +57,9 @@ main = do
             -- The watch root create event always seems to come even though the
             -- root is created before the watch is started. That may be because
             -- of batching?
-            -- XXX Need to create watch root after adding the watch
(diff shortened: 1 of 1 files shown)


test/Streamly/Test/FileSystem/Event/Darwin.hs
```diff
@@ -57,8 +57,9 @@ main = do
             -- The watch root create event always seems to come even though the
             -- root is created before the watch is started. That may be because
             -- of batching?
-            -- XXX Need to create watch root after adding the watch
-            : dirCreate "" (\dir -> [(dir, dirEvent Event.isCreated)])
+            -- XXX Need to create watch root after adding the watch, otherwise
+            -- it fails intermittently.
+            -- : dirCreate "" (\dir -> [(dir, dirEvent Event.isCreated)])
```
adithyaov, line 62: Maybe we can move this under `dev`.

adithyaov (approved):

PR #1374, opened 2021-12-07
Fix and test for concurrent applicative

adithyaov:
This is a breaking change as it modifies the behavior of `<*>` in `ZipAsyncM`

### DIFF
--- a/benchmark/Streamly/Benchmark/Prelude/ZipAsync.hs
+++ b/benchmark/Streamly/Benchmark/Prelude/ZipAsync.hs
@@ -41,6 +41,12 @@ zipAsyncAp count n =
     S.fromZipAsync $
         (,) <$> sourceUnfoldrM count n <*> sourceUnfoldrM count (n + 1)
 
+fromZipAsyncTraverse :: String -> Int -> Benchmark
+fromZipAsyncTraverse name count =
+    bench name
+        $ nfIO
+        $ S.drain $ S.fromZipAsync $ traverse S.fromPure [0 :: Int .. count]
+
 o_1_space_joining :: Int -> [Benchmark]
 o_1_space_joining value =
     [ bgroup "joining"
@@ -50,6 +56,7 @@ o_1_space_joining value =
                                                        (value `div` 2))
         , benchIOSrc fromSerial "zipAsyncAp (2,x/2)" (zipAsyncAp (value `div` 2))
         , benchIOSink value "fmap zipAsyncly" $ fmapN S.fromZipAsync 1
+        , fromZipAsyncTraverse "ZipAsync Applicative (x/100)" (value `div` 100)
         ]
     ]
 
--- a/docs/ConcurrentStreams.hs
+++ b/docs/ConcurrentStreams.hs
@@ -723,19 +723,20 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- 'fromZipAsync' type combinator can be used to switch to parallel applicative
 -- zip composition:
 --
--- This takes 7 seconds to zip, which is max (1,3) + max (2,4) because 1 and 3
--- are produced concurrently, and 2 and 4 are produced concurrently:
---
 -- >>> d n = delay n >> return n
--- >>> s1 = Stream.fromSerial $ d 1 <> d 2
--- >>> s2 = Stream.fromSerial $ d 3 <> d 4
+-- >>> s1 = Stream.fromSerial $ d 2 <> d 4
+-- >>> s2 = Stream.fromSerial $ d 3 <> d 1
 -- >>> (Stream.toList $ Stream.fromZipAsync $ (,) <$> s1 <*> s2) >>= print
--- ThreadId ...: Delay 1
 -- ThreadId ...: Delay 2
 -- ThreadId ...: Delay 3
+-- ThreadId ...: Delay 1
 -- ThreadId ...: Delay 4
--- [(1,3),(2,4)]
+-- [(2,3),(4,1)]
 --
+-- The actions within each stream are executed serially, however, the two
+-- streams are run concurrently with respect to each other. Therefore, it takes
+-- 6 seconds to zip the two streams, which is the maximum of the time to
+-- evaluate stream s1 (2 + 4 = 6 seconds) and stream s2 (3 + 1 = 4 seconds).
 
 -- $concurrent
 --
--- a/src/Streamly/Internal/Data/Stream/Zip.hs
+++ b/src/Streamly/Internal/Data/Stream/Zip.hs
@@ -54,14 +54,17 @@ import Streamly.Internal.Data.Maybe.Strict (Maybe'(..), toMaybe)
 import Streamly.Internal.Data.Stream.Serial (SerialT(..))
 import Streamly.Internal.Data.Stream.StreamK.Type (Stream)
 
-import qualified Streamly.Internal.Data.Stream.Parallel as Par
 import qualified Streamly.Internal.Data.Stream.Prelude as P
     (cmpBy, eqBy, foldl', foldr, fromList, toList)
 import qualified Streamly.Internal.Data.Stream.StreamK.Type as K
+import qualified Streamly.Internal.Data.Stream.StreamK as K
 import qualified Streamly.Internal.Data.Stream.StreamD as D
 import qualified Streamly.Internal.Data.Stream.Serial as Serial
+import qualified Streamly.Internal.Data.Stream.SVar.Eliminate as SVar
+import qualified Streamly.Internal.Data.Stream.SVar.Generate as SVar
 
 import Prelude hiding (map, repeat, zipWith, errorWithoutStackTrace)
+import Streamly.Internal.Data.SVar
 
 #include "Instances.hs"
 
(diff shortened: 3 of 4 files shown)


harendra-kumar:
> This is a breaking change as it modifies the behavior of <*> in ZipAsyncM

It only fixes the buggy behavior, we cannot characterize every bug fix as a breaking change.

harendra-kumar:
Can you add a test for this?

harendra-kumar:
Can you also add a benchmark like this:

```
    bench "ZipAsync Applicative" $ nfIO $ Stream.drain
        $ Stream.fromZipAsync $ traverse Stream.fromPure [0 :: Int ..99]
```

Also, it would be a good idea to extract the code snippets from the md files in `docs` dir and run docspec on those so that we know that the examples are working.

src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -784,6 +792,22 @@ zipWithM f = go
...
+zipAsyncWithM :: MonadAsync m =>
+    (a -> b -> m c) -> Stream m a -> Stream m b -> Stream m c
+zipAsyncWithM f m1 m2 = mkStream $ \st yld sng stp -> do
+    sv1 <- newParallelVar StopNone (adaptState st)
+    sv2 <- newParallelVar StopNone (adaptState st)
+    SVar.toSVarParallel (adaptState st) sv1 $ D.fromStreamK m1
+    SVar.toSVarParallel (adaptState st) sv2 $ D.fromStreamK m2
+    let pm1 = getSerialT $ SVar.fromSVar sv1
+        pm2 = getSerialT $ SVar.fromSVar sv2
+    foldStream st yld sng stp $ zipWithM f pm1 pm2
```
harendra-kumar, line 809: Can we create an SVar only for the second stream? So that we evaluate the first stream in the current thread and the second stream in a child thread. That way we do not have to create an additional thread.

adithyaov:
```
Generating reports for Prelude.ZipAsync...
Prelude.ZipAsync(cpuTime)
Benchmark                                                    default(0)(μs) default(1) - default(0)(%)
------------------------------------------------------------ -------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        8147.03                     +18.69
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           9350.73                     +11.55
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         8457.43                     +11.01
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly               541.25                      +6.00
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp           58.34                     -17.96

Prelude.ZipAsync(Allocated)
Benchmark                                                    default(0)(KiB) default(1) - default(0)(%)
------------------------------------------------------------ --------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         28136.49                     +46.02
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        28161.23                     +45.89
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           33624.88                     +34.75
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly               6246.38                      +0.03
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp           265.27                      -3.45

Prelude.ZipAsync(bytesCopied)
Benchmark                                                    default(0)(Bytes) default(1) - default(0)(%)
------------------------------------------------------------ ----------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        3530069.00                     +68.80
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         3709184.00                     +53.52
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           4631803.00                     +20.14
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp            8322.00                      +1.92
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly                  834.00                      -1.46
```

adithyaov:
```
Generating reports for Prelude.ZipAsync...
Prelude.ZipAsync(cpuTime)
Benchmark                                                    default(0)(μs) default(1) - default(0)(%)
------------------------------------------------------------ -------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly               527.33                     -12.19
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         8293.65                     -64.05
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        8116.46                     -67.56
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           9356.93                     -69.78
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp           57.39                     -87.16

Prelude.ZipAsync(Allocated)
Benchmark                                                    default(0)(KiB) default(1) - default(0)(%)
------------------------------------------------------------ --------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         28135.67                      +9.80
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        28160.90                      +9.70
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           33568.09                      +9.31
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly               6246.38                      +0.03
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp           265.23                     -33.31

Prelude.ZipAsync(bytesCopied)
Benchmark                                                    default(0)(Bytes) default(1) - default(0)(%)
------------------------------------------------------------ ----------------- --------------------------
All.Prelude.ZipAsync/o-1-space.joining.fmap zipAsyncly                  834.00                      -1.46
All.Prelude.ZipAsync/o-1-space.monad-outer-product.toNullAp            8330.00                    -106.85
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWithM (2,x/2)        3540197.00                    -165.41
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncWith (2,x/2)         3775020.00                    -182.14
All.Prelude.ZipAsync/o-1-space.joining.zipAsyncAp (2,x/2)           4809149.00                    -219.39
```

src/Streamly/Internal/Data/Stream/Zip.hs
```diff
@@ -96,9 +99,10 @@ zipWithK f = zipWithMK (\a b -> return (f a b))
 {-# INLINE zipAsyncWithMK #-}
 zipAsyncWithMK :: MonadAsync m
     => (a -> b -> m c) -> Stream m a -> Stream m b -> Stream m c
-zipAsyncWithMK f m1 m2 = D.toStreamK $
-    D.zipWithM f (Par.mkParallelD $ D.fromStreamK m1)
-                 (Par.mkParallelD $ D.fromStreamK m2)
```
adithyaov, line 101: This seems cleaner! We can achieve this using unfold and the composition of unfolds?

adithyaov: Composition of `inject` would do the trick of initializing the parallel svars.

docs/ConcurrentStreams.hs
```diff
@@ -723,18 +723,21 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
 -- 'fromZipAsync' type combinator can be used to switch to parallel applicative
 -- zip composition:
 --
--- This takes 7 seconds to zip, which is max (1,3) + max (2,4) because 1 and 3
--- are produced concurrently, and 2 and 4 are produced concurrently:
+-- This takes 6 seconds to zip, which is max (2 + 4, 3 + 1) because both the
+-- streams are run concurrently but the elements in the stream are produced
+-- serially. There is intra-stream concurrency but no inter-stream
+-- concurrency. Intra-stream concurrency is due to 'fromZipAsync' whereas the
+-- absence of inter-stream concurrency is due to 'fromSerial'.
```
harendra-kumar, line 730: Let's avoid using the inter/intra stream terms it may be confusing. Also, keep it terse but clear. How about: 

"The actions within each stream are executed serially, however, the two streams are run concurrently with respect to each other. Therefore, it takes 6 seconds to zip the two streams, which is the maximum of the time to evaluate stream s1 (2 + 4 = 6 seconds) and stream s2 (3 + 1 = 4 seconds)."

docs/ConcurrentStreams.hs
```diff
@@ -723,18 +723,21 @@ import Control.Monad.Trans.Class   (MonadTrans (lift))
...
 --
 -- >>> d n = delay n >> return n
--- >>> s1 = Stream.fromSerial $ d 1 <> d 2
--- >>> s2 = Stream.fromSerial $ d 3 <> d 4
+-- >>> s1 = Stream.fromSerial $ d 2 <> d 4
+-- >>> s2 = Stream.fromSerial $ d 3 <> d 1
```
harendra-kumar, line 734: Was there some specific reason to change the values?

benchmark/Streamly/Benchmark/Prelude/ZipAsync.hs
```diff
@@ -41,6 +41,11 @@ zipAsyncAp count n =
     S.fromZipAsync $
         (,) <$> sourceUnfoldrM count n <*> sourceUnfoldrM count (n + 1)
 
+fromZipAsyncTraverse100 :: Benchmark
+fromZipAsyncTraverse100 =
+    bench "ZipAsync Applicative"
+        $ nfIO $ S.drain $ S.fromZipAsync $ traverse S.fromPure [0 :: Int .. 99]
```
harendra-kumar, line 47: Why 100? For time consuming benchmarks we usually scale them by a factor e.g. `n/10`. See benchmarks that do something like this and do the same here as well.

docs/ConcurrentStreams.hs
adithyaov: The previous test fails, there is no consistent way for it to pass.
`1 + 2`
`3 + 4`
1 comes first
No way to know whether 2 or 3 comes next

benchmark/Streamly/Benchmark/Prelude/ZipAsync.hs
adithyaov: Even for 100 elements, this takes like a few `ms`. Need to check what the problem is. Probably the use of `traverse`.

harendra-kumar: Yes, I know that is perf issue. But irrespective of that we can use a scaling factor to reduce the timing appropriately, instead of using a constant value.

harendra-kumar (approved):

PR #1382, opened 2021-12-20
Add Stream.Mut.Fold.Foreign

adithyaov:
- There is a lot of common code
- Can use CPP and extract out commonality
- This module is required as mutable arrays are important first-class citizens
- The only problem is the common code
- The logic is exactly the same and hence the only required review is how or whether to add this module


harendra-kumar:
Can you please fix the docspec CI?

PR #1383, opened 2021-12-21
implement rollWindow

specdrake:


harendra-kumar (changes requested):
Would be nice to have a benchmark and a test for this.

The Ring module does not have a corresponding benchmark module. You can create one. See the similar Fold module benchmarks to know how to benchmark this.

src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -506,3 +509,48 @@ unsafeFoldRingNM count rh f z rb@Ring {..} =
...
+rollWindow :: forall m a b. (MonadIO m, Storable a) 
+    => Int -> Fold m (a, Maybe a) b -> Fold m a b
+rollWindow n (Fold step1 initial1 extract1)= Fold step initial extract
+
+    where
+
+    initial = do 
+        s <- initial1
+        (rb, rh) <- liftIO $ new n
+        return (Fold.Partial $ Tuple4' rb rh (0 :: Int) s)
```
harendra-kumar, line 534: We should not just tuck in `s` in the state. We have to examine `s` and if it returns `Done` we have to return `Done` right away. If it is Partial we have to store the state inside Partial in the tuple, do not store the state boxed inside Partial/Done constructor. You delayed inspection of the result of `initial1` to the `step` function which is incorrect.

See how other such combinators are implemented in the Fold module.

harendra-kumar (changes requested):

specdrake:
> Would be nice to have a benchmark and a test for this.
> 
> The Ring module does not have a corresponding benchmark module. You can create one. See the similar Fold module benchmarks to know how to benchmark this.

Where should I write its test? Are there tests for other ``Ring`` functions?

harendra-kumar:
@adithyaov can you advise regarding the tests.

src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -506,3 +509,48 @@ unsafeFoldRingNM count rh f z rb@Ring {..} =
...
+        (rb, rh) <- liftIO $ new n
+        return $
+            case s of
+                Fold.Partial x -> Fold.Partial $ Tuple4' rb rh (0 :: Int) x
+                Fold.Done b -> Fold.Done b
+
+    step (Tuple4' rb rh i x') a 
+        | i < n = do
+            rh1 <- liftIO $ unsafeInsert rb rh a
+            s <- step1 x' (a, Nothing)
```
harendra-kumar, line 542: Use `r` as return value, and `s` can be used for the state in Partial.

src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -506,3 +509,48 @@ unsafeFoldRingNM count rh f z rb@Ring {..} =
...
+            return $
+                case s of
+                    Fold.Partial x -> Fold.Partial $ Tuple4' rb rh1 (i + 1) x
+                    Fold.Done b -> Fold.Done b
+        | otherwise = do
+            a' <- liftIO $ peek rh
```
harendra-kumar, line 548: We can use `old` instead of `a'` for better cue about what it is. We usually do not use prime on variable names, to represent new generation of a variable we use x1, x2 etc instead.

harendra-kumar: Also, see the coding guidelines here https://github.com/composewell/streamly/blob/master/CONTRIBUTING.md .

src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -506,3 +508,48 @@ unsafeFoldRingNM count rh f z rb@Ring {..} =
...
+-- out the oldest element and the new element along with the old element are 
+-- supplied to the collector fold. 
+--
+-- The Maybe is for the case when initially the window is filling and 
+-- there is no old element.
+--
+{-# INLINE rollWindow #-}
+rollWindow :: forall m a b. (MonadIO m, Storable a) 
+    => Int -> Fold m (a, Maybe a) b -> Fold m a b
+rollWindow n (Fold step1 initial1 extract1)= Fold step initial extract
```
harendra-kumar, line 526: We can name it as `slidingWindow`, that's a more common term for such processing, also it is a noun and folds are usually nouns.

src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -506,3 +508,48 @@ unsafeFoldRingNM count rh f z rb@Ring {..} =
         if ptr == rh || n == 0
             then return acc'
             else go (n - 1) acc' ptr
+
+data Tuple4' a b c d = Tuple4' !a !b !c !d deriving Show
+
+-- | @rollWindow n collector@ is a simple, incremental sliding window
```
harendra-kumar, line 514: We can omit the word `simple`.

harendra-kumar (commented):
Looks good. Just two little things to change.

harendra-kumar:
There is a failure of "trailing spaces" check in the hlint CI. Can you do two things:

1. Remove the trailing spaces from those lines
2. In the hlint CI where the trailing spaces check is being done emit a message on console saying "Checking trailing spaces..." so that in the CI messages we know why it is failing.

harendra-kumar:
Also, raise an issue about adding tests and benchmarks for slidingWindow function, so that we can track it.

adithyaov (changes requested):
Changes requested.

harendra-kumar (approved):
Just squash the commits.

PR #1386, opened 2021-12-23
Extend the behaviour of Utf8 to replicate that of strict Text

adithyaov:
* Few undefined APIs.
* Few APIs need to be improved, need to be made more performant.

This PR adds the general structure and some basic implementation.

For reviewer:
- The only thing that needs to be reviewed are the `XXX` comments and the export list.
- We might not need a few APIs. The docs are copied with minor modification, they don't really need to be reviewed.
- I wanted to add individual small commits but at one point it became too late.


src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,501 @@
...
+"TEXT last -> unfused" [1] forall t.
+    S.last (stream t) = last t
+  #-}
+-}
+
+-- | /O(1)/ Returns all characters after the head of a 'Utf8', or 'Nothing' if
+-- empty.  Subject to fusion.
+{-# INLINE_NORMAL tail #-}
+tail :: Utf8 -> Maybe Utf8
+tail = fmap unstream . unsafePerformIO . Stream.tail . stream
```
harendra-kumar, line 398: Use `uncons` to get the tail slice.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,501 @@
...
+"TEXT tail -> unfused" [1] forall t.
+    unstream (S.tail (stream t)) = tail t
+ #-}
+-}
+
+-- | /O(1)/ Returns all but the last character of a 'Utf8', or 'Nothing' if
+-- empty.  Subject to fusion.
+{-# INLINE_NORMAL init #-}
+init :: Utf8 -> Maybe Utf8
+init = fmap unstream . unsafePerformIO . Stream.init . stream
```
harendra-kumar, line 413: If we can write a routine to decode utf8 in reverse then we can just decode the last char from the end of the array and then slice it.

Otherwise, use `last` on the stream, get the encoded length of the last char and use that to slice it.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,501 @@
...
+-- | /O(1)/ Tests whether a 'Utf8' contains exactly one character.  Subject to
+-- fusion.
+{-# INLINE isSingleton #-}
+isSingleton :: Utf8 -> Bool
+isSingleton = undefined
+
+-- | /O(n)/ Returns the number of characters in a 'Utf8'.  Subject to fusion.
+{-# INLINE length #-}
+length :: Utf8 -> Int
+length = unsafePerformIO . Stream.length . stream
```
harendra-kumar, line 453: We could possibly determine the length faster by using a custom routine that counts the starting chars from the utf8 encoded bytes without decoding the chars.

harendra-kumar (commented):
I only reviewed the Type module. Will review others later.

1. Can you arrange the code in define before use order? For example the `pack` operation is used before it is defined. Similarly, `toArray`.
2. Operations that apply to generic array modules can be moved to the array module.

src/Streamly/Internal/Unicode/Utf8/Eliminate.hs
```diff
@@ -0,0 +1,414 @@
...
+    , splitAt
+    , tail
+    , take
+    , takeWhile
+    , unlines
+    , unwords
+    , words
+    , zip
+    , zipWith
+    )
```
harendra-kumar, line 110: This list can be cut short.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,474 @@
...
+empty :: Utf8
+empty = Utf8 Array.nil
+
+--------------------------------------------------------------------------------
+-- Helpers
+--------------------------------------------------------------------------------
+
+{-# INLINE toStream #-}
+toStream :: Utf8 -> SerialT IO Char
+toStream = Unicode.decodeUtf8' . Array.toStream . toArray
```
harendra-kumar, line 139: Why IO? Can't we use Identity or `Monad m` here? It will require MonadIO constraint everywhere which is not good.

src/Streamly/Internal/Unicode/Utf8/Eliminate.hs
```diff
@@ -0,0 +1,414 @@
...
+-- >>> :set -XOverloadedStrings
+-- >>> import qualified Streamly.Internal.Unicode.Utf8 as Utf8
+
+--------------------------------------------------------------------------------
+-- Reducing Streams (folds)
+--------------------------------------------------------------------------------
+
+{-# INLINE fold #-}
+fold :: MonadIO m => Fold m Char b -> Utf8 -> m b
+fold f = Stream.fold f . hoist liftIO . toStream
```
harendra-kumar, line 122: Can we remove the `hoist` and MonadIO constraint?

src/Streamly/Internal/Unicode/Utf8/Eliminate.hs
```diff
@@ -0,0 +1,414 @@
...
+{-# INLINE foldl #-}
+foldl :: (a -> Char -> a) -> a -> Utf8 -> a
+foldl = undefined
+
+-- | A strict version of 'foldl'.
+--
+-- /Time complexity:/ O(n)
+{-# INLINE foldl' #-}
+foldl' :: (a -> Char -> a) -> a -> Utf8 -> a
+foldl' f z t = unsafePerformIO $ Stream.foldl' f z (toStream t)
```
harendra-kumar, line 139: Can we use Identity monad instead of IO?

src/Streamly/Internal/Unicode/Utf8/Eliminate.hs
```diff
@@ -0,0 +1,414 @@
...
+_isSuffixOf a b = unsafePerformIO (Stream.isSuffixOf (toStream a) (toStream b))
+
+-- XXX This specific API uses a lot of memory to compile
+-- XXX Use domain specific knowledge to implement it efficiently!
+-- | The 'isInfixOf' function takes two 'Utf8's and returns
+-- 'True' iff the first is contained, wholly and intact, anywhere
+-- within the second.
+--
+-- In (unlikely) bad cases, this function's time complexity degrades
+-- towards /O(n*m)/.
```
harendra-kumar, line 321: `Stream.isInfixOf` doc says `O(n)`, why would this be any different? If it is true then we should change that as well.

src/Streamly/Internal/Unicode/Utf8/Eliminate.hs
```diff
@@ -0,0 +1,414 @@
...
+
+-- | 'foldl', applied to a binary operator, a starting value
+-- (typically the left-identity of the operator), and a 'Utf8',
+--
+-- /Time complexity:/ O(n)
+-- reduces the 'Utf8' using the binary operator, from left to right.
+--
+{-# INLINE foldl #-}
+foldl :: (a -> Char -> a) -> a -> Utf8 -> a
+foldl = undefined
```
harendra-kumar, line 132: Can you add the `/Unimplemented/` in the docs for undefined functions so that it appears in the docs?

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
```diff
@@ -0,0 +1,334 @@
...
+{-
+-- | Returns the /array/ index (in units of 'Word16') at which a
+-- character may be found.  This is /not/ the same as the logical
+-- index returned by e.g. 'findIndex'.
+findAIndexOrEnd :: (Char -> Bool) -> Utf8 -> Int
+findAIndexOrEnd q t@(Utf8 _arr _off len) = go 0
+    where go !i | i >= len || q c       = i
+                | otherwise             = go (i+d)
+                where Iter c d          = iter t i
+-}
```
harendra-kumar, line 146: We can remove this, or just keep the signature if you really want to add it later.

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
```diff
@@ -0,0 +1,334 @@
...
+--
+-- and
+--
+-- > intercalate s . splitOn s         == id
+-- > splitOn (singleton c)             == split (==c)
+--
+-- (Note: the string @s@ to split on above cannot be empty.)
+--
+-- In (unlikely) bad cases, this function's time complexity degrades
+-- towards /O(n*m)/.
```
harendra-kumar, line 197: ??

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
```diff
@@ -0,0 +1,334 @@
...
+--
+-- > append prefix match == haystack
+-- >   where (prefix, match) = breakOn needle haystack
+--
+-- If you need to break a string by a substring repeatedly (e.g. you
+-- want to break on every instance of a substring), use 'breakOnAll'
+-- instead, as it has lower startup overhead.
+--
+-- In (unlikely) bad cases, this function's time complexity degrades
+-- towards /O(n*m)/.
```
harendra-kumar, line 270: ??

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
```diff
@@ -0,0 +1,334 @@
...
+-- | The 'partition' function takes a predicate and a 'Utf8',
+-- and returns the pair of 'Utf8's with elements which do and do not
+-- satisfy the predicate, respectively; i.e.
+--
+-- > partition p t == (filter p t, filter (not . p) t)
+--
+-- /Time complexity:/ O(n)
+{-# INLINE partition #-}
+partition :: (Char -> Bool) -> Utf8 -> (Utf8, Utf8)
+partition p t = (filter p t, filter (not . p) t)
```
harendra-kumar, line 306: Instead of reading it twice, can we use a partition fold to write both the Utf8 segments in one go?

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
```diff
@@ -0,0 +1,334 @@
...
+-- Examples:
+--
+-- >> breakOnAll "::" ""
+-- []
+--
+-- >> breakOnAll "/" "a/b/c/"
+-- [("a","/b/c/"),("a/b","/c/"),("a/b/c","/")]
+--
+-- In (unlikely) bad cases, this function's time complexity degrades
+-- towards /O(n*m)/.
```
harendra-kumar, line 325: ??

src/Streamly/Internal/Unicode/Utf8/Transform.hs
```diff
@@ -0,0 +1,584 @@
...
+-- "foo"
+--
+-- In cases where several instances of @needle@ overlap, only the
+-- first one will be replaced:
+--
+-- >> replace "ofo" "bar" "ofofo"
+-- "barfo"
+--
+-- In (unlikely) bad cases, this function's time complexity degrades
+-- towards /O(n*m)/.
```
harendra-kumar, line 212: ??

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,474 @@
...
+
+{-# INLINE fromStream #-}
+fromStream :: SerialT IO Char -> Utf8
+fromStream =
+    Utf8 . unsafePerformIO . Stream.fold Array.write . Unicode.encodeUtf8'
+
+{-# INLINE second #-}
+second :: (b -> c) -> (a, b) -> (a, c)
+second f (a, b) = (a, f b)
```
harendra-kumar, line 148: Why not use `second` from Bifunctor?

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,474 @@
...
+
+{-
+-- | Apply a function to the first element of an optional pair.
+{-# INLINE firstf #-}
+firstf :: (a -> c) -> Maybe (a,b) -> Maybe (c,b)
+firstf f (Just (a, b)) = Just (f a, b)
+firstf _  Nothing      = Nothing
+-}
```
harendra-kumar, line 156: Remove it if its not used.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,474 @@
...
+{-# INLINE read #-}
+read :: Unfold m Utf8 Char
+read = undefined
+
+--------------------------------------------------------------------------------
+-- Basic functions
+--------------------------------------------------------------------------------
+
+singleton :: Char -> Utf8
+singleton x = pack [x]
```
harendra-kumar, line 339: `pack` may be inefficient because it may not fuse. We can use `fromStream . Stream.fromPure`.

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
adithyaov: Now that I read it again, this seems like a weird statement. It should be `O(n*m)`.

src/Streamly/Internal/Unicode/Utf8/Reduce.hs
adithyaov: My implementation is the bare minimum first thing in my mind kind of a thing. I'll encode the improvement suggesstions. A lot of improvement is required for sure.

src/Streamly/Internal/Unicode/Utf8/Type.hs
adithyaov: Ah, I've written IO keeping mutable arrays in mind. It can be removed for immutable stuff.

src/Streamly/Internal/Unicode/Utf8.hs
```diff
@@ -6,62 +6,160 @@
...
+        [t] -> t
+        xs -> Prelude.foldl1 append xs
+
+-- | Map a function over a 'Utf8' that results in a 'Utf8', and
+-- concatenate the results.
+--
+-- /Time complexity:/ O(n)
+{-# INLINE concatMap #-}
+concatMap :: (Char -> Utf8) -> Utf8 -> Utf8
+concatMap f = concat . foldr ((:) . f) []
```
harendra-kumar, line 109: I would prefer to use stream based implementation instead of using list.

src/Streamly/Internal/Unicode/Utf8.hs
```diff
@@ -6,62 +6,160 @@
...
+lines = split (== '\n')
 
-    len = length s
+-- | Joins lines, after appending a terminating newline to
+-- each.
+--
+-- /Time complexity:/ O(n)
+{-# INLINE unlines #-}
+unlines :: [Utf8] -> Utf8
+unlines = concat . List.map (`snoc` '\n')
```
harendra-kumar, line 158: using snoc may not be most efficient, we should implement it in a streaming way, see unlines in Unicode.Stream module.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,422 @@
...
+
+{-# INLINE toStream #-}
+toStream :: Monad m => Utf8 -> SerialT m Char
+toStream = Unicode.decodeUtf8 . Array.toStream . toArray
+
+-- XXX Try removing IO
+-- XXX This would require rewriting Array.Foreign.write-ish functions. We can't
+-- use the underlying Array.Foreign.Mut.write-ish
+{-# INLINE fromStream #-}
+fromStream :: SerialT IO Char -> Utf8
```
harendra-kumar, line 106: This must be `fromStream :: SerialT m Char -> m Utf8`. You cannot have the result pure when the stream is not pure.

src/Streamly/Internal/Unicode/Utf8/Type.hs
```diff
@@ -0,0 +1,422 @@
...
+{-# INLINE_NORMAL pack #-}
+pack :: String -> Utf8
+pack = fromStream . Stream.fromList
+
+-- | Convert a 'Utf8' into a 'String'.
+--
+-- /Time complexity:/ O(n)
+{-# INLINE_NORMAL unpack #-}
+unpack :: Utf8 -> String
+unpack = unsafePerformIO . Stream.toList . toStream
```
harendra-kumar, line 134: Use identity monad, do not use unsafePerformIO anywhere.

harendra-kumar (changes requested):
1. Remove _all_ unsafePerformIO. If you retain any there should be a good reason for that.
2. I think we should use a `SerialT m a` wherever we are using a `[a]`. That will provide a more efficient way of composing the APIs. I do not think we need a drop in replacement for text. We can add the missing APIs in a compat module if we feel that is needed.

PR #1389, opened 2021-12-27
Implement putIndices

rnjtranjan:


harendra-kumar (approved):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -547,10 +547,26 @@ withNewArrayUnsafe count f = do
...
 -- error if any index is out of bounds.
 --
--- /Unimplemented/
+-- /Pre-release/
 {-# INLINE putIndices #-}
-putIndices :: Array a -> Fold m (Int, a) ()
-putIndices = undefined
+putIndices :: forall m a. (MonadIO m, Storable a)
+    => Array a -> Fold m (Int, a) ()
+putIndices Array{..} = Fold step initial extract
```
harendra-kumar, line 554: This would just be `Fold.drainBy (uncurry (putIndex arr))`. Should not make any difference to perf as each element is anyway independently modified.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -610,6 +583,19 @@ putIndex arr i x =
     unsafeWithArrayContents (arrContents arr) (arrStart arr)
         $ \p -> putIndexPtr p (aEnd arr) i x
 
+-------------------------------------------------------------------------------
+-- Random writes
+-------------------------------------------------------------------------------
```
harendra-kumar, line 588: This header should remain at its original place. It is for all the random write functions.

PR #1390, opened 2021-12-27
Implement modifyIndices

rnjtranjan:


harendra-kumar (approved):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -626,12 +626,23 @@ modifyIndex Array{..} i f = do
...
+-- | Modify the array indices generated by the supplied stream.
 --
 -- /Pre-release/
-modifyIndices :: -- forall m a b. (MonadIO m, Storable a) =>
-    Unfold m (Array a) Int -> Array a -> (a -> a) -> m ()
-modifyIndices = undefined
+modifyIndices :: forall m a. (MonadIO m, Storable a)
+    => D.Stream m Int -> Array a -> (a -> a) -> m ()
+modifyIndices str Array{..} f = do
+    D.foldlM' func (return ()) str
```
harendra-kumar, line 635: We can use `modifyIndex` itself to implement this. If you can write a benchmark we can check if there is any difference in perf if at all, I do not see any reason for that though.

harendra-kumar (changes requested):

rnjtranjan (commented):
looks good.

PR #1391, opened 2021-12-27
Implement modify

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -635,10 +635,21 @@ modifyIndices = undefined
...
+    where
+
+    go i = do
+        unsafeWithArrayContents arrContents arrStart $ \ptr -> do
+            let elemSize = sizeOf (undefined :: a)
+                elemPtr = ptr `plusPtr` (elemSize * i)
+            when (elemPtr `plusPtr` elemSize <= aEnd) $ do
+                r <- peek elemPtr
+                poke elemPtr (f r)
+                go (i+1)
```
harendra-kumar, line 652: Instead of passing the index in the loop we can pass the ptr itself, so that we do not have to multiply with the index each time. We can just keep incrementing the ptr.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -638,18 +638,17 @@ modifyIndices = undefined
...
 modify :: forall m a. (MonadIO m, Storable a)
     => Array a -> (a -> a) -> m ()
-modify Array{..} f = liftIO $ go 0
+modify Array{..} f = liftIO $ go arrStart
 
     where
 
     go i = do
-        unsafeWithArrayContents arrContents arrStart $ \ptr -> do
+        unsafeWithArrayContents arrContents arrStart $ \_ -> do
```
harendra-kumar, line 646: This should not be in the loop, should be outside the loop.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -638,18 +638,17 @@ modifyIndices = undefined
...
             let elemSize = sizeOf (undefined :: a)
-                elemPtr = ptr `plusPtr` (elemSize * i)
-            when (elemPtr `plusPtr` elemSize <= aEnd) $ do
-                r <- peek elemPtr
-                poke elemPtr (f r)
-                go (i+1)
+            when (i `plusPtr` elemSize <= aEnd) $ do
```
harendra-kumar, line 648: rename `i` to `ptr`.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -642,13 +642,16 @@ modify Array{..} f = liftIO $ go arrStart
...
-                r <- peek i
-                poke i (f r)
-                go (i `plusPtr` elemSize)
+            go' ptr0 elemSize
+            where
+                go' ptr elemSize = do
+                    when (ptr `plusPtr` elemSize <= aEnd) $ do
+                        r <- peek ptr
+                        poke ptr (f r)
+                        go' (ptr `plusPtr` elemSize) elemSize
```
harendra-kumar, line 654: We do not need another `go` for this. We can do:

```
modify Array{..} f = 
    unsafeWithArrayContents arrContents arrStart $ \_ ->
        liftIO $ go arrStart
```

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -638,20 +638,18 @@ modifyIndices = undefined
 -- /Pre-release/
 modify :: forall m a. (MonadIO m, Storable a)
     => Array a -> (a -> a) -> m ()
-modify Array{..} f = liftIO $ go arrStart
+modify Array{..} f = liftIO $
+    unsafeWithArrayContents arrContents arrStart $ \ptr -> do
+        let elemSize = sizeOf (undefined :: a)
+        go ptr elemSize
```
harendra-kumar, line 644: We do not need to pass `elemSize` as it is statically determined, you can compute it inside `go` itself.

harendra-kumar (changes requested):

harendra-kumar (approved):
squash the commits.

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -635,10 +635,21 @@ modifyIndices = undefined
...
 --
--- /Unimplemented/
-modify :: -- forall m a b. (MonadIO m, Storable a) =>
-    Array a -> (a -> a) -> m ()
-modify = undefined
+-- /Pre-release/
+modify :: forall m a. (MonadIO m, Storable a)
+    => Array a -> (a -> a) -> m ()
+modify Array{..} f = liftIO $
+    unsafeWithArrayContents arrContents arrStart $ \ptr -> do
```
harendra-kumar, line 642: This can be simplified to `unsafeWithArrayContents arrContents arrStart go`

PR #1392, opened 2021-12-27
Implement swapIndices

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -643,9 +643,19 @@ modify = undefined
...
-swapIndices = undefined
+swapIndices :: forall m a. (MonadIO m, Storable a)
+    => Array a -> Int -> Int -> m ()
+swapIndices Array{..} i1 i2 = liftIO $ do
+    unsafeWithArrayContents arrContents arrStart $ \ptr -> do
+        let elemSize = sizeOf (undefined :: a)
+            elemPtr1 = ptr `plusPtr` (elemSize * i1)
+            elemPtr2 = ptr `plusPtr` (elemSize * i2)
+        when (i1 < 0 ) $ invalidIndex "swapIndices" i1
+        when (i2 < 0 ) $ invalidIndex "swapIndices" i2
```
harendra-kumar, line 654: 1. We need to check if we are writing beyond the end of the array
2. we can implement `unsafeSwapIndices` to swap without any index checks and use that here. It can be useful in perf critical higher level combinators that need to swap indices.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -643,20 +644,31 @@ modify = undefined
...
         r2 :: a <- peek elemPtr2
         poke elemPtr1 r2
         poke elemPtr2 r1
 
+swapIndices :: forall m a. (MonadIO m, Storable a)
+    => Array a -> Int -> Int -> m ()
+swapIndices arr@Array{..} i1 i2 = liftIO $ do
+        let elemSize = sizeOf (undefined :: a)
+            elemPtr1 = arrStart `plusPtr` (elemSize * i1)
+            elemPtr2 = arrStart `plusPtr` (elemSize * i2)
```
harendra-kumar, line 664: I hope this duplication is removed by GHC. We can check the core, if it does.

harendra-kumar (approved):

rnjtranjan (commented):
looks good

PR #1393, opened 2021-12-27
Implement roundUpToPower2

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -677,10 +678,15 @@ roundUpLargeArray size =
...
-roundUpToPower2 = undefined
--}
+roundUpToPower2 x =
+    let z = x - 1
+        z1 = z .|. z `shiftR` 1
+        z2 = z1 .|. z1 `shiftR` 2
+        z3 = z2 .|. z2 `shiftR` 4
+        z4 = z3 .|. z3 `shiftR` 8
+        z5 = z4 .|. z4 `shiftR` 16
+        in z5 + 1
```
harendra-kumar, line 689: This assumes the word size to be 32-bit. 

We can use `#include "MachDeps.h"` and test `WORD_SIZE_IN_BITS` to be 64 or 32 and handle appropriately. For 64-bit word we will need a z6 as well.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -686,7 +686,10 @@ roundUpToPower2 x =
         z3 = z2 .|. z2 `shiftR` 4
         z4 = z3 .|. z3 `shiftR` 8
         z5 = z4 .|. z4 `shiftR` 16
-        in z5 + 1
+        z6 = if (maxBound :: Int) == 9223372036854775807
```
harendra-kumar, line 689: Its not a good idea to relay on maxBound. It also places a branch and comparison in this code which makes it less efficient. Let's do it statically, [see this for example](https://hackage.haskell.org/package/statistics-0.15.2.0/docs/src/Statistics.Function.html#nextHighestPowerOfTwo).

src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -677,10 +679,21 @@ roundUpLargeArray size =
             ((size + blockSize - 1) .&. negate blockSize)
     else size
 
-{-
 roundUpToPower2 :: Int -> Int
-roundUpToPower2 = undefined
--}
+roundUpToPower2 x =
+#if WORD_SIZE_IN_BITS == 64
+    1 + _z6
```
harendra-kumar, line 685: Why is it named `_z6`? We are testing on 64-bit platforms so it should not be unused.

harendra-kumar (approved):
Squash the commits.

PR #1394, opened 2021-12-27
Implement reverse

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1120,10 +1120,22 @@ getSlice index len (Array contents start e _) =
...
+    swap l h arr
+
+    where
+
+    swap l h arr0 = do
+        if l < h
+        then do
+            swapIndices arr0 l h
+            swap (l+1) (h-1) arr0
+        else return True
```
harendra-kumar, line 1138: To implement it more efficiently, we should use two pointers instead of the indices. We can keep incrementing the lower pointer and decrement the higher one. We do not need to do validations of indices and computing the pointer from the index which is being done in swapIndices.

harendra-kumar (changes requested):

harendra-kumar (approved):
Looks good. See if you can adapt the stream reverse test to test this.

PR #1395, opened 2021-12-27
Issue 1215 partitionBy

rnjtranjan:


harendra-kumar:
Can you rebase it and use unsafeSwapIndices?

harendra-kumar:
Can you add a benchmark for this? I want to try out a few perf experiments.

benchmark/streamly-benchmarks.cabal
```diff
@@ -483,6 +483,17 @@ benchmark Data.Array.Foreign
   if flag(limit-build-mem)
     ghc-options: +RTS -M1000M -RTS
 
+benchmark Data.Mut.Array
```
harendra-kumar, line 486: Benchmark modules are usually named after the library modules. This should be named `Data.Array.Foreign.Mut`.

benchmark/streamly-benchmarks.cabal
```diff
@@ -483,6 +483,17 @@ benchmark Data.Array.Foreign
...
+  import: bench-options
+  type: exitcode-stdio-1.0
+  hs-source-dirs: Streamly/Benchmark/Data/Mut
```
harendra-kumar, line 489: This should be in Streamly/Benchmark/Data/Array/Foreign/Mut .

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
```diff
@@ -21,6 +21,8 @@ module Streamly.Internal.Data.Array.Stream.Mut.Foreign
     , compactLE
     , compactEQ
     , compactGE
+    , Array(..)
+    , partitionBy
```
harendra-kumar, line 25: Why is this added?

harendra-kumar (changes requested):

rnjtranjan (commented):
looks good.

harendra-kumar:
@rnjtranjan can you push a commit fixing the above comments.

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
rnjtranjan: used in Streamly/Benchmark/Data/Array/Foreign/Mut/Array.hs

harendra-kumar: Why is a mutable array API being exported via ArrayStream module and then used from there? Why aren't we using it directly from the mutable array module?

benchmark/Streamly/Benchmark/Data/Mut/Array.hs
```diff
@@ -0,0 +1,96 @@
...
+    main
+  ) where
+
+import Control.DeepSeq (NFData(..))
+import System.Random (randomRIO)
+import Prelude hiding ()
+
+import qualified Streamly.Prelude  as Stream
+import qualified Streamly.Internal.Data.Fold as Fold
+import qualified Streamly.Internal.Data.Array.Stream.Mut.Foreign as MArray
```
harendra-kumar, line 32: We should import the array APIs from `Data.Array.Foreign.Mut`.

benchmark/Streamly/Benchmark/Data/Mut/Array.hs
```diff
@@ -0,0 +1,96 @@
...
+    :: NFData b
+    => String -> (Int -> t IO a) -> (t IO a -> IO b) -> Benchmark
+benchIO name src sink =
+    bench name $ nfIO $ randomRIO (1,1) >>= sink . src
+
+{-# INLINE partitionBy #-}
+partitionBy :: Int -> SerialT IO (MArray.Array Int) -> IO ()
+partitionBy value s = Stream.fold Fold.drain $ do
+    a <- s
+    MArray.partitionBy ( < value) a
```
harendra-kumar: This is unnecessarily complicated. We should run the benchmark on the array itself instead of running it on a stream of arrays.

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
harendra-kumar: This is the wrong module to export these from. The right module is `Data.Array.Foreign.Mut` and these are already exported from that module. So we can just remove these from here.

benchmark/Streamly/Benchmark/Data/Mut/Array.hs
rnjtranjan: arraysOf :: (IsStream t, MonadIO m, Storable a)
    => Int -> t m a -> t m (Array a) is not defined here (Data.Array.Foreign.Mut). Do we need to redefine here?

harendra-kumar: We should not be using arraysOf, see my other comment. And even if we need that we can import that from this module and import other stuff from mutable array module.

benchmark/Streamly/Benchmark/Data/Mut/Array.hs
rnjtranjan: agreed.

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut/Array.hs
```diff
@@ -57,20 +56,19 @@ sourceUnfoldrM value n = Stream.unfoldrM step n
...
+    => String -> (Int -> [a]) -> ([a] -> IO b) -> Benchmark
 benchIO name src sink =
     bench name $ nfIO $ randomRIO (1,1) >>= sink . src
 
 {-# INLINE partitionBy #-}
-partitionBy :: Int -> SerialT IO (MArray.Array Int) -> IO ()
-partitionBy value s = Stream.fold Fold.drain $ do
-    a <- s
-    MArray.partitionBy ( < value) a
+partitionBy :: Int -> [MArray.Array Int] -> IO ()
```
harendra-kumar, line 64: Why list of arrays? We need just a single array.

harendra-kumar (commented):
You can use change this:

```
Stream.toList $ MArray.arraysOf value $ sourceUnfoldrM value 0
```

to something like

```
Array.fromStream $ sourceUnfoldrM value 0
```

PR #1396, opened 2021-12-27
Implement mergeOrdered

rnjtranjan:


harendra-kumar:
This does not look useful. We can achieve this by converting the arrays to streams, use stream merge and then covert back to array. In this module we are concerned mostly with in-place mutation algorithms.

PR #1400, opened 2021-12-28
Implement compactGE idiomatically

adithyaov:


src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
```diff
@@ -246,6 +246,55 @@ compactLEParserD n = ParserD.Parser step initial extract
     functionPath =
         "Streamly.Internal.Data.Array.Stream.Mut.Foreign.compactLEParserD"
 
+-- | Coalesce adjacent arrays in incoming stream to form bigger arrays of a
+-- minimum specified size. Note that if a all the arrays in the stream together
```
harendra-kumar: typo "a all".

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
```diff
@@ -246,6 +246,55 @@ compactLEParserD n = ParserD.Parser step initial extract
...
+    => Int -> FL.Fold m (Array a) (Array a)
+compactGEFold n = Fold step initial extract
+
+    where
+
+    nBytes = n * sizeOf (undefined :: a)
+
+    initial =
+        return
+            $ if n <= 0
```
harendra-kumar, line 268: `n == 0` should be a valid case in this.

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
```diff
@@ -246,6 +246,55 @@ compactLEParserD n = ParserD.Parser step initial extract
...
+    step Nothing arr =
+        return
+            $ let len = MArray.byteLength arr
+               in if len >= nBytes
+                  then FL.Done arr
+                  else FL.Partial (Just arr)
+    step (Just buf) arr = do
+        let len = MArray.byteLength buf + MArray.byteLength arr
+        buf1 <-
+            if MArray.byteCapacity buf < len
```
harendra-kumar, line 284: Since we have collect `>= n` we can allocate at least `n` right away.

harendra-kumar: In compactLE as well we can allocate equal to `n` to avoid repeated allocations.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Array/Stream/Mut/Foreign.hs
adithyaov: Yeah.

adithyaov: In compactLE, we do it already.

harendra-kumar (approved):

PR #1407, opened 2022-01-02
Bump the upper bound of unicode-data

adithyaov:
- Used in normalization combinators
- Essentially the normalization helpers did not change over the versions
- Admit 0.3.*

### DIFF
--- a/benchmark/streamly-benchmarks.cabal
+++ b/benchmark/streamly-benchmarks.cabal
@@ -114,7 +114,7 @@ common bench-depends
     -- Core libraries shipped with ghc, the min and max
     -- constraints of these libraries should match with
     -- the GHC versions we support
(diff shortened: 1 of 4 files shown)


harendra-kumar:
We can also change base and filepath to lowest major version.

adithyaov:
`filepath` is already constrained, I believe.

harendra-kumar (approved):

harendra-kumar:
> `filepath` is already constrained, I believe.

I meant restricting the constraint to major version. Instead of

```
, filepath >= 1.2.0.0 && < 1.4.3.0
```

Use

```
, filepath >= 1.2 && < 1.5
```

adithyaov:
We should just make a revision on hackage right?

PR #1422, opened 2022-01-09
Move some code to use define-before-use order

harendra-kumar:


adithyaov (approved):
Assuming you've just moved stuff around without changing any logic. Approved.

PR #1423, opened 2022-01-16
Fail on exceeding threshold

adithyaov:


benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -18,6 +18,7 @@ print_help () {
   echo "       [--benchmarks <"bench1 bench2 ..." | help>]"
   echo "       [--prefix <benchmark name prefix to match>"
   echo "       [--fields <"field1 field2 ..." | help>]"
+  echo "       [--failure-field]"
```
harendra-kumar, line 21: We can call it `--fail-on`.

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -40,6 +41,8 @@ print_help () {
   echo "--benchmarks: benchmarks to run, use 'help' for list of benchmarks"
   echo "--compare: compare the specified benchmarks with each other"
   echo "--fields: measurement fields to report, use 'help' for a list"
+  echo "--threshold: threshold percentage to decide regression"
```
harendra-kumar, line 44: We already have `--diff-cutoff-percent` that serves a similar purpose.

benchmark/bench-report/bin/bench-runner.sh
adithyaov: We probably don't need this. It should fail on all the fields. We can filter fields accordingly though.

adithyaov:
See #1431

PR #1424, opened 2022-01-18
Add Fold.postscan

specdrake:


src/Streamly/Internal/Data/Fold.hs
```diff
@@ -416,6 +417,46 @@ scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
...
+-- another 'Fold'.
+-- /Pre-release/
+{-# INLINE postscan #-}
+postscan :: Monad m => Fold m a b -> Fold m b c -> Fold m a c
+postscan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
+    Fold step initial extract
+
+    where
+
+    runStep actionL sR = do
```
harendra-kumar, line 430: Since this is common in `scan` and `postscan`, can we take this out and reuse in both?

harendra-kumar: It is a good idea to add an INLINE annotation on this function.

harendra-kumar (changes requested):
Can you also add a benchmark for this? We can just copy the `scan` benchmark if it exists.

src/Streamly/Internal/Data/Fold.hs
specdrake: > Since this is common in `scan` and `postscan`, can we take this out and reuse in both?

I'm facing some difficulties in making the `runStep` function (taken outside) to typecheck.
```
 Couldn't match expected type ‘s3’ with actual type ‘s1’
      ‘s3’ is a rigid type variable bound by
        a pattern with constructor:
          Fold :: forall (m :: * -> *) a b s.
                  (s -> a -> m (Step s b))
                  -> m (Step s b) -> (s -> m b) -> Fold m a b,
        in an equation for ‘runStep’
        at src/Streamly/Internal/Data/Fold.hs:388:41-68
      ‘s1’ is a rigid type variable bound by
        the type signature for:
          runStep :: forall (m :: * -> *) a b c s s1.
                     Monad m =>
                     Fold m a b
                     -> Fold m b c -> m (Step s b) -> s1 -> m (Step (s, s1) c)
        at src/Streamly/Internal/Data/Fold.hs:387:1-111
        ```

Comment (harendra-kumar) [reply src/Streamly/Internal/Data/Fold.hs:430]: You can leave it like that, just add a to-do comment saying that these two are the same can use a common definition.

File: benchmark/Streamly/Benchmark/Data/Fold.hs [hunk 2]
```diff
@@ -303,6 +303,16 @@ o_1_space_serial_transformation value =
...
+        , benchIOSink
+            value
+            "fold-scan"
+            (S.fold
+                (FL.scan FL.drain FL.drain))
+        , benchIOSink
+            value
+            "fold-postscan"
+            (S.fold
+                (FL.postscan FL.drain FL.drain))
```
harendra-kumar, line 315: FL.drain is too trivial. Let's use `FL.scan FL.sum FL.drain` instead.

harendra-kumar (commented):
Can you also post the benchmark results?

While we are at it we can add a `scan1` as well, in a separate PR.

specdrake:
```
All 1 tests passed (0.16s)
All.Data.Fold/o-1-space.transformation.fold-scan:   +RTS -T -K36K -M16M    -RTS
All
  Data.Fold/o-1-space
    transformation
      fold-scan: OK (0.15s)
        37.7 μs ± 3.1 μs,   0 B  allocated,   0 B  copied, 2.0 MB peak memory

All 1 tests passed (0.16s)
All.Data.Fold/o-1-space.transformation.fold-postscan:   +RTS -T -K36K -M16M    -RTS
All
  Data.Fold/o-1-space
    transformation
      fold-postscan: OK (0.15s)
        38.1 μs ± 3.7 μs,   0 B  allocated,   0 B  copied, 2.0 MB peak memory
```

specdrake:
Is this PR mergeable?

PR #1429, opened 2022-01-25
Update the CIs

harendra-kumar:


.github/workflows/haskell.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    name: GHC ${{ matrix.name }}
+    name: GHC ${{matrix.variant}}
```
adithyaov, line 11: Why this change? Why not just use `name`?

.github/workflows/haskell.yml
```diff
@@ -47,97 +47,111 @@ jobs:
...
-          - 8.8.4
-          - 8.8.4+inspection+fusion-plugin+Werror
-          - 8.6.5+fusion-plugin
-          - 8.6.5+streamk
-          - 8.4.4+debug
+        # The order is important to optimize fail-fast.
+        variant:
+          - 8.10.7-sdist-fusion-inspection-Werror
+          - 9.2.1-docspec
+          - 8.10.7-coverage
```
adithyaov, line 58: It would be nice to have all the `variants` to be listed here. More informative and easier to read through.

.github/workflows/haskell.yml
```diff
@@ -47,97 +47,111 @@ jobs:
...
+
+        # Note: if cabal.project is not specified benchmarks and tests won't
+        # run. But we need at least one test where we test without
+        # cabal.project because that is how hackage would build it.
         include:
-          - name: 9.2.1
+          - variant: head
```
adithyaov, line 64: Can you separate all variants with a new line?

.github/workflows/haskell.yml
adithyaov: Or none of them. Does `include` not respect the order?

.github/workflows/haskell.yml
harendra-kumar: I did more changes earlier where name was derived from variant, ghc version and other things. Those changes were reverted but I kept the variant.

.github/workflows/haskell.yml
harendra-kumar: Not sure what that means. Do you want a blank line after the variant line?

adithyaov: Yes

Issue #1428, opened 2022-01-25
GHC-9: invert INLINE pragmas on all mutually recursive loops


harendra-kumar:
From performance benchmark regressions it seems like the definition of which function in a mutually recursive loop would be a loop breaker has changed. We need to invert the INLINE pragmas on all mutually recursive loops.

Issue #1068, opened 2021-05-24
GHC-9: FileSystem.Handle benchmark regressions


harendra-kumar:
After including the GHC rewrite rule fix:

```
FileSystem.Handle(Allocated)
Benchmark                                                                                    default(0)(MiB) default(1) - default(0)(%)
-------------------------------------------------------------------------------------------- --------------- --------------------------
All.FileSystem.Handle/o-1-space.copy.FH.writeWithBufferOf . FH.readWithBufferOf                       202.59                   +9477.62
All.FileSystem.Handle/o-1-space.copy.FH.write . FH.read                                               202.62                   +9476.21
All.FileSystem.Handle/o-1-space.copy/read.rawToNull                                                   202.57                   +7898.66
All.FileSystem.Handle/o-1-space.copy/read.rawToFile                                                   202.70                   +7893.67
All.FileSystem.Handle/o-1-space.reduce/read.SS.decodeUtf8 (1/10)                                       10.45                   +1139.24
```

harendra-kumar:
Two of the benchmarks are fixed by removing NOINLINE on the benachmarks which was required earlier:

```
FileSystem.Handle(cpuTime)
Benchmark                                                                                    default(0)(ms) default(1) - default(0)(%) default(2) - default(0)(%)
-------------------------------------------------------------------------------------------- -------------- -------------------------- --------------------------
All.FileSystem.Handle/o-1-space.copy.FH.writeWithBufferOf . FH.readWithBufferOf                      103.67                   +1268.08                      +2.32
All.FileSystem.Handle/o-1-space.copy.FH.write . FH.read                                              104.43                   +1264.89                      +1.45
```

harendra-kumar:
All allocation regressions are gone now, but some cpu time regressions remain:

```
FileSystem.Handle(cpuTime)
Benchmark                                                                                    default(0)(ms) default(1) - default(0)(%)
-------------------------------------------------------------------------------------------- -------------- --------------------------
All.FileSystem.Handle/o-1-space.reduce/read.US.lines . SS.decodeLatin1                                69.47                     +56.20
All.FileSystem.Handle/o-1-space.reduce/read/chunks.S.chunksOf 1 (1/10)                               104.65                     +48.91
All.FileSystem.Handle/o-1-space.reduce/read/chunks.S.arraysOf 1 (1/10)                               120.09                     +41.56
[...]

Issue #1067, opened 2021-05-24
GHC-9: Unicode stream regressions


harendra-kumar:
All Unicode.Stream benchmarks have regressed. These results are after including the GHC rewrite rule fix:

```
Unicode.Stream(Allocated)
Benchmark                                                                                          default(0)(MiB) default(1) - default(0)(%)
-------------------------------------------------------------------------------------------------- --------------- --------------------------
All.Unicode.Stream/o-1-space.decode-encode.SS.encodeLatin1' . SS.decodeLatin1                               202.55                   +7899.55
All.Unicode.Stream/o-1-space.decode-encode.SS.encodeLatin1 . SS.decodeLatin1                                202.61                   +7897.03
All.Unicode.Stream/o-1-space.decode-encode/toChunks.decodeEncodeUtf8Lenient (1/10)                           26.63                   +7307.98
All.Unicode.Stream/o-1-space.decode-encode.SS.encodeUtf8 . SS.decodeUtf8Lax (1/10)                           28.64                   +6910.65
All.Unicode.Stream/o-1-space.ungroup-group.S.interposeSuffix . S.splitOnSuffix(Array Word8) (1/10)           82.05                   +3013.24
All.Unicode.Stream/o-1-space.ungroup-group.UA.unlines . UA.lines (Array Char) (1/10)                        231.36                   +1068.76
All.Unicode.Stream/o-1-space.ungroup-group.US.unlines . S.splitOnSuffix ([Word8]) (1/10)                    661.96                    +144.90
All.Unicode.Stream/o-1-space.ungroup-group.US.unwords . S.wordsBy ([Char]) (1/10)                           681.31                    +140.49
All.Unicode.Stream/o-1-space.ungroup-group.S.interposeSuffix . S.wordsBy ([Word8]) (1/10)                   685.53                    +133.46
All.Unicode.Stream/o-1-space.ungroup-group.UA.unwords . UA.words (Array Char) (1/10)                       1647.77                     +76.55
```

harendra-kumar:
The situation on this looks like this as of now:

```
Unicode.Stream(Allocated)
Benchmark                                                                                            default(0)(MiB) default(1) - default(0)(%)
---------------------------------------------------------------------------------------------------- --------------- --------------------------
All.Unicode.Stream/o-1-space.ungroup-group.interposeSuffix . splitOnSuffix (Array Word8) (1/10)                80.81                   +1867.45
[...]

Issue #1066, opened 2021-05-24
GHC-9: Fold and Parser perf regressions


harendra-kumar:
```
Data.Fold(Allocated)
Benchmark                                                              default(0)(Bytes) default(1) - default(0)(%)
---------------------------------------------------------------------- ----------------- --------------------------
All.Data.Fold/o-1-space.elimination.drainN                                      13335.00                  +46790.86
All.Data.Fold/o-1-space.composition.many drain (take 1)                         13396.00                  +15406.93
```

```
Data.Parser(Allocated)
Benchmark                                                                   default(0)(Bytes) default(1) - default(0)(%)
--------------------------------------------------------------------------- ----------------- --------------------------
All.Data.Parser/o-1-space.groupBy                                                    13318.00                  +78495.46
All.Data.Parser/o-1-space.groupByRolling                                             13363.00                  +78230.40
All.Data.Parser/o-1-space.sliceBeginWith                                           6503438.00                    +118.39
All.Data.Parser/o-n-heap.choice                                                   21477672.00                     +74.27
```

```
Data.Parser(cpuTime)
Benchmark                                                                     default(0)(s) default(1) - default(0)(%)
--------------------------------------------------------------------------- --------------- --------------------------
All.Data.Parser/o-1-space.groupBy                                               64707893.00                   +2550.05
All.Data.Parser/o-1-space.groupByRolling                                        76729618.00                   +1977.47
All.Data.Parser/o-1-space.sliceBeginWith                                      1107494400.00                    +207.97
All.Data.Parser/o-1-space.parseMany/Unfold/1000 arrays/take 1                  359844000.00                     +69.49
All.Data.Parser/o-1-space/filesystem.S.parseMany (FL.take 1 FL.sum)          83482449000.00                     +22.57
All.Data.Parser/o-n-heap.choice                                              12841884800.00                     +14.25
```

harendra-kumar:
After fixing INLINE and Fuse annotations, all regressions in allocations in folds or parsers benchmarks are fixed, some cpu time regressions remain in parser benchmarks. First run is GHC-8.10.7 and the second run is GHC-9.3.20211119 with SPJ's rewrite rules patch:

```
Data.Parser.ParserD(cpuTime)
Benchmark                                                                      default(0)(μs) default(1) - default(0)(%)
------------------------------------------------------------------------------ -------------- --------------------------
All.Data.Parser.ParserD/o-1-space.spanByRolling                                         61.04                     +35.76
[...]

PR #1430, opened 2022-01-26
Fix performance regressions in GHC-9

harendra-kumar:
See #1061, #1066, #1067, #1068, #1428 .

There were several fixes in GHC mentioned in the issues above.

The type of fixes required in streamly are:

* INLINE annotation added in a few cases, or INLINABLE converted to INLINE
* INLINE pragma was swapped in mutually recursive cases (deciding the loop breaker)
* Remove NOINLINE from some benchmarks
* Fuse annotation added in one case

There are still some cases remaining to be fixed.

### DIFF
--- a/src/Streamly/Internal/Data/Parser/ParserD.hs
+++ b/src/Streamly/Internal/Data/Parser/ParserD.hs
@@ -268,7 +268,7 @@ fromFold (Fold fstep finitial fextract) = Parser step initial fextract
 --
 -- /Pre-release/
 --
-{-# INLINABLE peek #-}
+{-# INLINE peek #-}
 peek :: MonadThrow m => Parser m a a
 peek = Parser step initial extract
 
@@ -284,7 +284,7 @@ peek = Parser step initial extract
 --
 -- /Pre-release/
 --
-{-# INLINABLE eof #-}
+{-# INLINE eof #-}
 eof :: Monad m => Parser m a ()
 eof = Parser step initial return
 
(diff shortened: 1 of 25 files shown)


src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -744,7 +743,7 @@ splitMany (Parser step1 initial1 extract1) (Fold fstep finitial fextract) =
                     IError _ -> done <$> fextract fs
             FL.Done fb -> return $ done fb
 
-    -- Do not inline this
+    {-# INLINE handleCollect #-}
```
adithyaov, line 746: This should be `runCollectorWith`?

src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -866,7 +863,7 @@ splitSome (Parser step1 initial1 extract1) (Fold fstep finitial fextract) =
...
```
adithyaov, line 866: `runCollectorWith`?

src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -806,7 +804,7 @@ splitManyPost (Parser step1 initial1 extract1) (Fold fstep finitial fextract) =
...
```
adithyaov, line 807: `runCollectorWith`?

src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1176,7 +1176,7 @@ manyTill (Fold fstep finitial fextract)
 
     -- Caution: Mutual recursion
 
-    -- Don't inline this
+    {-# INLINE scrutR #-}
```
adithyaov, line 1179: `scrutL`?

adithyaov (changes requested):
Can fix the comments? The rest looks good.

harendra-kumar:
Got closed accidentally, I will open another one.

PR #1431, opened 2022-01-26
Fail on exceeding threshold

adithyaov:


benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -664,8 +669,29 @@ done
...
+    then
+        echo "$targets_reports"
+    else
+        echo "No reports generated for the following, $TARGETS"
+    fi
+  fi
+
+  if [ -n "$COMPARISON_REPORTS" ]
+  then
+    comparision_reports=$(run_reports "$COMPARISON_REPORTS")
```
harendra-kumar, line 692: typo `comparison_reports`

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -436,15 +436,20 @@ run_reports() {
...
             --benchmark $i \
             $(test "$USE_GAUGE" = 1 && echo "--use-gauge") \
             $(test "$GRAPH" = 1 && echo "--graphs") \
             $(test "$SORT_BY_NAME" = 1 && echo "--sort-by-name") \
             $(test -n "$BENCH_DIFF_STYLE" && echo "--diff-style $BENCH_DIFF_STYLE") \
             $(test -n "$BENCH_CUTOFF_PERCENT" && echo "--diff-cutoff-percent $BENCH_CUTOFF_PERCENT") \
-            --fields "$FIELDS"
+            --fields "$FIELDS")
+
+        if [ -n "$res" ]
```
harendra-kumar, line 454: I usually use `test -n "$res"` rather than using `[`. I find `[` a bit weird, it is not some shell syntax, it is a command with the name `[`, you can type `which [` and see. `test` is more explicit.

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -436,15 +436,20 @@ run_reports() {
...
+        then
+            echo "Generating reports for ${i}..."
+            echo $res
```
harendra-kumar, line 457: If you store the output in a variable, all the newlines will be lost.

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -664,8 +669,29 @@ done
...
 then
-  run_reports "$TARGETS"
-  run_reports "$COMPARISON_REPORTS"
+
+  if [ -n "$TARGETS" ]
+  then
+    targets_reports=$(run_reports "$TARGETS")
+    if [ -n "$targets_reports" ]
+    then
+        echo "$targets_reports"
```
harendra-kumar, line 684: The newlines will be lost.

benchmark/bench-report/bin/bench-runner.sh
harendra-kumar: I guess if you use `echo "$res"` it might work.

benchmark/bench-report/bin/bench-runner.sh
harendra-kumar: This might work as the variable is quoted.

benchmark/bench-report/bin/bench-runner.sh
harendra-kumar: Instead of buffering the output, we can just omit the "Generating reports..." message.

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -664,8 +669,29 @@ done
...
+  if [ -n "$COMPARISON_REPORTS" ]
+  then
+    comparision_reports=$(run_reports "$COMPARISON_REPORTS")
+    if [ -n "$comparision_reports" ] && [ -n "$COMPARISON_REPORTS" ]
+    then
+        echo "$comparision_reports"
+    else
+        echo "No reports generated for the following, $COMPARISON_REPORTS"
+    fi
+  fi
```
harendra-kumar, line 699: I don't like buffering all the output. Why do we even need the "No reports generated" message from this script? The user of this script can examine the output themselves. If you want to detect per target output then you can invoke this script supplying only one target at a time.

bin/bench-with-exit.sh
```diff
@@ -0,0 +1,18 @@
...
+# If no reports are generated, the last line of the output will look like:
+# No reports generated for the following, ...
+
+SCRIPT_DIR=$(cd `dirname $0`; pwd)
+
+OUTPUT=$($SCRIPT_DIR/bench.sh "$@" | tail -1)
+PREFIX_FOR_SUCCESS="No reports generated for the following,"
+LEN_PREFIX=${#PREFIX_FOR_SUCCESS}
+
+[ "${OUTPUT:0:$LEN_PREFIX}" = "$PREFIX_FOR_SUCCESS" ]
```
harendra-kumar, line 18: Why can't we just check that the output of `bench.sh` is empty?

benchmark/bench-report/BenchReport.hs
```diff
@@ -224,10 +224,7 @@ selectBench Options{..} f =
     cutOffMultiples = 1 + cutOffPercent / 100
 
     filterPred x =
-        case benchType of
-            Just (Compare _) ->
-                x <= negate cutOffPercent || x >= cutOffPercent
-            _ -> True
+        x <= negate cutOffPercent || x >= cutOffPercent
```
harendra-kumar: If we are not comparing, then how does cutOffPercent make sense? I am missing something.

benchmark/bench-report/bin/bench-runner.sh
```diff
@@ -39,6 +39,12 @@ print_help () {
   echo
   echo "--benchmarks: benchmarks to run, use 'help' for list of benchmarks"
   echo "--compare: compare the specified benchmarks with each other"
+
+  echo "--diff-cutoff-percent: Diff percentage used for benchmark selection."
+  echo "The diff is considered only between the first and second column of the"
```
harendra-kumar, line 44: Not very clear description. How about something like;

"When comparing benchmarks with a baseline, show only those benchmarks that changed by more than the cutoff percentage. When more than two sets of benchmarks are present, this applies to only the first two."

benchmark/bench-report/BenchReport.hs
harendra-kumar: Ok, I guess Compare is for comparison groups, the diff applies in the other case as well. I have not looked at the code.

benchmark/bench-report/BenchReport.hs
```diff
@@ -209,25 +209,25 @@ selectBench
...
-      sortFunc
-      $ f (ColumnIndex 1) (Just PercentDiff)
+    -- Apply filterPred only if at least 2 columns exist
+    let colVals =
+            case f (ColumnIndex 1) (Just PercentDiff) of
+                Left _ -> either error id $ f (ColumnIndex 0) (Just PercentDiff)
+                Right xs -> filter (\(_,y) -> filterPred y) xs
+    in reverse
+           $ fmap fst
+           $ filter (\(_,y) -> filterPred y)
```
adithyaov, line 219: This should not be there

adithyaov:
Something like this should be used in the CI,
```
bin/bench.sh --benchmarks Data.Array --no-measure --diff-cutoff-percent 2 --silent | grep '\n'
```

harendra-kumar:
> Something like this should be used in the CI,
> 
> ```
> bin/bench.sh --benchmarks Data.Array --no-measure --diff-cutoff-percent 2 --silent | grep '\n'
> ```

Or 
`
```
... | grep -v '^$' > /dev/null
```

Should return exit code 1 on success.

harendra-kumar (approved):

Issue #1426, opened 2022-01-19
CI failure in `Prelude.Top`


adithyaov:
https://github.com/composewell/streamly/runs/4857174772?check_suite_focus=true#step:6:16056

harendra-kumar:
I got this failure again https://github.com/composewell/streamly/runs/4955445252?check_suite_focus=true:

```
Failures:

  Streamly/Test/Prelude/Top.hs:81:9: 
  1) Prelude.Top joinInnerMap
       Assertion failed (after 88 tests):
         [3379,1635,6454,7244,303,8306,3697,8993,2526,705,4339,6024,9524,4601,2572,7365,2280,1497,1850,5750,8150,9542,284,6867,2262,8691,9449,4580,707,3721,4216,8454,5575,5633,72,5051,7558,4553,1077,9477,4362,2703,7711,9435,3170,4837,7344,3918,9590,3002,6411,1110,4340,8160,5664,682,3087,8461,8120,458,1558,2641,2005,356,7046,7295,4306,5937,9512,9319,6065,4048,1152,1168,7649,319,1681,9045,784,8979,9778,7388,7392,3756,6971,2286,2548]
         [6057,4379,7115,545,3394,5328,4612,6850,9048,5287,9511,1884,7964,6187,1077,9637,9516,5381,2834,2738,2299,8572,9147,396,3684,5141,765,5792,3282,7691,3616,6353,8397,9384,1635,470,8746,1386,4686,7668,3288,1077,5673,2799,2562,1613,4837,7727,4136,7509,7836,2622,9681,502,7438,8637,872,5412,8823,6404,5108,524,9511,9834,6187,545,9222,5269]
```

rnjtranjan:
Fixed: https://github.com/composewell/streamly/pull/1432

PR #1432, opened 2022-01-27
Fix test case failure

rnjtranjan:


rnjtranjan:
Fixes: #1426

harendra-kumar (changes requested):
Looks good. hlint CI needs to be fixed.

PR #1433, opened 2022-01-27
Add a github workflow for checking regressions

adithyaov:


adithyaov:
This PR needs to be merged into master to check how it works.

You can do the following to verify:
1. Make a fork of this repo
2. Merge this branch into master
3. Follow **Regression checking** in `dev/MAINTAINING.md`

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,91 @@
...
+
+    - name: Checkout the current branch
+      uses: actions/checkout@v2
+      with:
+        clean: false
+
+    - name: Run benchmarks and append
+      run: |
+        chmod +x $BENCH_SH
+        $BENCH_SH --benchmarks "$BENCHMARKS" --raw --quick --append
```
harendra-kumar, line 77: Perhaps do not need to use `--quick` as there is not much difference. We can check how much we gain with this.

dev/MAINTAINING.md
```diff
@@ -319,3 +319,66 @@ dependencies. May show additional ghc boot libraries.
...
+gh workflow run regression-check.yml \
+    -f benchmarks=<benchmarks> \
+    -f fields=<fields> \
+    -f diff-cutoff-percent=<diff-cutoff-percent> \
+    -r <branch>
+```
+
+Examples,
+
+The following command run the workflow for `Data.Array` on branch `array-dev`,
```
harendra-kumar, line 340: "runs".

dev/MAINTAINING.md
```diff
@@ -319,3 +319,66 @@ dependencies. May show additional ghc boot libraries.
...
+scrutinizing the field `cuptime`, for any benchmarks that have regressed beyond
```
harendra-kumar, line 341: "cputime"

dev/MAINTAINING.md
```diff
@@ -319,3 +319,66 @@ dependencies. May show additional ghc boot libraries.
...
+```
+gh workflow run regression-check.yml \
+    -f benchmarks="Data.Array" \
+    -f fields="cputime" \
+    -f diff-cutoff-percent=20 \
+    -r array-dev
+```
+
+The following command run the workflow for `Data.Array and Data.Array.Foreign`
+on branch `array-refactor`, scrutinizing the fields `cuptime` and `bytescopied`,
```
harendra-kumar, line 353: run => runs, cuptime => cputime.

dev/MAINTAINING.md
```diff
@@ -319,3 +319,66 @@ dependencies. May show additional ghc boot libraries.
...
+for any benchmarks that have regressed beyond 35 percent.
+
+```
+gh workflow run regression-check.yml \
+    -f benchmarks="Data.Array Data.Array.Foreign" \
+    -f fields="cputime bytescopied" \
+    -f diff-cutoff-percent=35 \
+    -r array-refactor
+```
```
harendra-kumar, line 362: In the name of `brevity` we can just keep one of these examples, just this one is general enough.

.github/workflows/regression-check.yml
```diff
@@ -15,6 +15,11 @@ on:
         description: '--diff-cutoff-percent'
         required: false
         default: 50
+  pull_request:
+    with:
+      benchmarks: ""
+      fields: "cputime allocated"
```
harendra-kumar, line 21: `cputime` is not reliable especially on CI systems. It will fail all the time. We should use only allocations and the cutoff percent should be small like 3% because allocations usually remain exactly the same if there is no regression.

PR #1435, opened 2022-02-01
Add character based unicode decoding

adithyaov:
The two benchmarks doing the same are below,
```
decodeUtf8 = parseManyD parseCharUtf8D
```

```
Generating reports for Unicode.Stream...
Unicode.Stream(cpuTime)
Benchmark                                                                                            default(ms)
---------------------------------------------------------------------------------------------------- -----------
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseManyD parseCharUtf8D (1/10)                  589.66
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . decodeUtf8 (1/10)                                 141.05

Unicode.Stream(Allocated)
Benchmark                                                                                            default(MiB)
---------------------------------------------------------------------------------------------------- ------------
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseManyD parseCharUtf8D (1/10)                  1748.73
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . decodeUtf8 (1/10)                                   28.57

Unicode.Stream(bytesCopied)
Benchmark                                                                                            default(KiB)
---------------------------------------------------------------------------------------------------- ------------
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseManyD parseCharUtf8D (1/10)                 11986.93
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . decodeUtf8 (1/10)                                    9.44
```

### DIFF
--- a/src/Streamly/Internal/Unicode/Stream.hs
+++ b/src/Streamly/Internal/Unicode/Stream.hs
@@ -15,6 +15,9 @@ module Streamly.Internal.Unicode.Stream
       decodeLatin1
 
     -- ** UTF-8 Decoding
+    , CodingFailureMode(..)
+    , writeCharUtf8'
+    , parseCharUtf8With
     , decodeUtf8
     , decodeUtf8'
     , decodeUtf8_
@@ -78,6 +81,7 @@ where
 
 #include "inline.hs"
 
+import Control.Monad.Catch (MonadThrow, MonadCatch)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Data.Bits (shiftR, shiftL, (.|.), (.&.))
 import Data.Char (chr, ord)
@@ -101,6 +105,9 @@ import Streamly.Internal.Data.Unfold.Type (Unfold(..))
 import Streamly.Internal.System.IO (unsafeInlineIO)
 
 import qualified Streamly.Internal.Data.Unfold as Unfold
+import qualified Streamly.Internal.Data.Parser as Parser
+import qualified Streamly.Internal.Data.Parser.ParserD as ParserD
+import qualified Streamly.Internal.Data.Parser.ParserK.Type as ParserK
 import qualified Streamly.Internal.Data.Stream.Serial as Serial
 import qualified Streamly.Internal.Data.Array.Foreign as Array
 import qualified Streamly.Internal.Data.Array.Foreign.Type as A
@@ -422,6 +429,96 @@ data CodingFailureMode
 {-# INLINE replacementChar #-}
 replacementChar :: Char
 replacementChar = '\xFFFD'
+
+data UTF8CharDecodeState a
+    = UTF8CharDecodeInit
+    | UTF8CharDecoding !DecodeState !CodePoint
+
+{-# INLINE parseCharUtf8WithD #-}
+parseCharUtf8WithD ::
+       Monad m => CodingFailureMode -> ParserD.Parser m Word8 Char
+parseCharUtf8WithD cfm =
+    let A.Array _ ptr _ = utf8d
+    in ParserD.Parser (step' ptr) initial extract
+
+    where
+
+    prefix = "Streamly.Internal.Data.Stream.parseCharUtf8WithD:"
+
+    {-# INLINE initial #-}
+    initial = return $ ParserD.IPartial UTF8CharDecodeInit
+
+    handleError err souldBackTrack =
+        case cfm of
+            ErrorOnCodingFailure -> ParserD.Error err
+            TransliterateCodingFailure ->
+                case souldBackTrack of
+                    True -> ParserD.Done 1 replacementChar
+                    False -> ParserD.Done 0 replacementChar
+            DropOnCodingFailure ->
+                case souldBackTrack of
+                    True -> ParserD.Continue 1 UTF8CharDecodeInit
+                    False -> ParserD.Continue 0 UTF8CharDecodeInit
+
+    {-# INLINE step' #-}
+    step' table UTF8CharDecodeInit x =
+        -- Note: It is important to use a ">" instead of a "<=" test
+        -- here for GHC to generate code layout for default branch
+        -- prediction for the common case. This is fragile and might
+        -- change with the compiler versions, we need a more reliable
+        -- "likely" primitive to control branch predication.
+        return $ case x > 0x7f of
+            False -> ParserD.Done 0 $ unsafeChr $ fromIntegral x
+            True ->
+                let (Tuple' sv cp) = decode0 table x
+                 in case sv of
+                        12 ->
+                            let msg = prefix
+                                    ++ "Invalid first UTF8 byte" ++ show x
+                             in handleError msg False
+                        0 -> ParserD.Error $ prefix ++ "unreachable state"
+                        _ -> ParserD.Continue 0 (UTF8CharDecoding sv cp)
+
+    step' table (UTF8CharDecoding statePtr codepointPtr) x = return $
+        let (Tuple' sv cp) = decode1 table statePtr codepointPtr x
+         in case sv of
+            0 -> ParserD.Done 0 $ unsafeChr cp
+            12 ->
+                let msg = prefix
+                        ++ "Invalid subsequent UTF8 byte"
+                        ++ show x
+                        ++ "in state"
+                        ++ show statePtr
+                        ++ "accumulated value"
+                        ++ show codepointPtr
+                 in handleError msg True
+            _ -> ParserD.Continue 0 (UTF8CharDecoding sv cp)
+
+    {-# INLINE extract #-}
+    extract _ = error $ prefix ++ "Not enough input"
+
+-- XXX This should ideally accept a "CodingFailureMode" and perform appropriate
+-- error handling. This isn't possible now as "TransliterateCodingFailure"'s
+-- workflow requires backtracking 1 element. This can be revisited once "Fold"
+-- supports backtracking.
+{-# INLINE writeCharUtf8' #-}
+writeCharUtf8' :: MonadThrow m => Fold m Word8 Char
+writeCharUtf8' =  ParserD.toFold (parseCharUtf8WithD ErrorOnCodingFailure)
+
+-- XXX The initial idea was to have "parseCharUtf8" and offload the error
+-- handling to another parser. So, say we had "parseCharUtf8'",
+--
+-- >>> parseCharUtf8Smart = parseCharUtf8' <|> Parser.fromPure replacementChar
+--
+-- But unfortunately parseCharUtf8Smart used in conjunction with "parseMany" -
+-- that is "parseMany parseCharUtf8Smart" on a stream causes the heap to
+-- overflow. Even a heap size of 500 MB was not sufficient.
+--
+-- This needs to be investigated futher.
+{-# INLINE parseCharUtf8With #-}
+parseCharUtf8With ::
+       MonadCatch m => CodingFailureMode -> Parser.Parser m Word8 Char
+parseCharUtf8With = ParserK.toParserK . parseCharUtf8WithD
 
 -- XXX write it as a parser and use parseMany to decode a stream, need to check
 -- if that preserves the same performance. Or we can use a resumable parser
--- a/benchmark/Streamly/Benchmark/Unicode/Stream.hs
+++ b/benchmark/Streamly/Benchmark/Unicode/Stream.hs
@@ -267,6 +267,23 @@ inspect $ hasNoTypeClasses 'copyStreamUtf8
 -- inspect $ 'copyStreamUtf8Lax `hasNoType` ''D.ConcatMapUState
 #endif
 
+{-# NOINLINE _copyStreamUtf8'Fold #-}
+_copyStreamUtf8'Fold :: Handle -> Handle -> IO ()
+_copyStreamUtf8'Fold inh outh =
+   Stream.fold (Handle.write outh)
+     $ Unicode.encodeUtf8
+     $ Stream.foldMany Unicode.writeCharUtf8'
+     $ Stream.unfold Handle.read inh
+
+{-# NOINLINE _copyStreamUtf8Parser #-}
+_copyStreamUtf8Parser :: Handle -> Handle -> IO ()
+_copyStreamUtf8Parser inh outh =
+   Stream.fold (Handle.write outh)
+     $ Unicode.encodeUtf8
+     $ Stream.parseMany
+           (Unicode.parseCharUtf8With Unicode.TransliterateCodingFailure)
+     $ Stream.unfold Handle.read inh
+
 o_1_space_decode_encode_read :: BenchEnv -> [Benchmark]
 o_1_space_decode_encode_read env =
     [ bgroup "decode-encode"
@@ -280,7 +297,11 @@ o_1_space_decode_encode_read env =
         -- Requires valid unicode input
         , mkBench "encodeUtf8' . decodeUtf8'" env $ \inh outh ->
             _copyStreamUtf8' inh outh
+        , mkBench "encodeUtf8' . foldMany writeCharUtf8'" env $ \inh outh ->
+            _copyStreamUtf8'Fold inh outh
 #endif
+        , mkBenchSmall "encodeUtf8 . parseMany parseCharUtf8" env
+              $ \inh outh -> _copyStreamUtf8Parser inh outh
         , mkBenchSmall "encodeUtf8 . decodeUtf8" env $ \inh outh ->
             copyStreamUtf8 inh outh
         ]


src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
...
+                        ++ show x
+                        ++ "in state"
+                        ++ show statePtr
+                        ++ "accumulated value"
+                        ++ show codepointPtr
+                 in handleError table msg (Just x)
+            _ -> return $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
+
+writeCharUtf8' :: MonadThrow m => Fold m Word8 Char
+writeCharUtf8' =  ParserD.toFold $ parseCharUtf8WithD ErrorOnCodingFailure
```
harendra-kumar, line 499: The low level fold should take an argument to determine how to handle error. The signature would be:

```
writeCharUtf8With :: Monad m => CodingFailureMode -> Fold m Word8 Char
```

It should not use MonadThrow constraint.

We can then write `writeCharUtf8` and `writeCharUtf8'` in terms of the above fold.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
...
+
+parseCharUtf8D :: MonadThrow m => ParserD.Parser m Word8 Char
+parseCharUtf8D =  parseCharUtf8WithD TransliterateCodingFailure
```
harendra-kumar, line 502: The parser should not use `TransliterateCodingFailure` for failure handling. It should not do any error handling, it should always use `Error` constructor on failure. We can then run an alternative parser to handle error.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
 replacementChar :: Char
 replacementChar = '\xFFFD'
 
+data UTF8CharDecodeState a
+    = UTF8CharDecodeInit
+    | UTF8CharDecoding !DecodeState !CodePoint
+
+parseCharUtf8WithD
```
harendra-kumar, line 434: Use an INLINE pragma.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
...
+                        ++ "Invalid subsequent UTF8 byte"
+                        ++ show x
+                        ++ "in state"
+                        ++ show statePtr
+                        ++ "accumulated value"
+                        ++ show codepointPtr
+                 in handleError table msg (Just x)
+            _ -> return $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
+
+writeCharUtf8' :: MonadThrow m => Fold m Word8 Char
```
harendra-kumar, line 498: Use an INLINE pragma.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
...
+writeCharUtf8' =  ParserD.toFold $ parseCharUtf8WithD ErrorOnCodingFailure
+
+parseCharUtf8D :: MonadThrow m => ParserD.Parser m Word8 Char
```
harendra-kumar, line 501: Use an INLINE pragma.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,80 @@ data CodingFailureMode
 replacementChar :: Char
 replacementChar = '\xFFFD'
 
+data UTF8CharDecodeState a
+    = UTF8CharDecodeInit
+    | UTF8CharDecoding !DecodeState !CodePoint
```
harendra-kumar, line 432: If it does not fuse you can try ANN Fuse on it.

harendra-kumar (changes requested):

adithyaov:
```
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8' . foldMany writeCharUtf8'                    17.03
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . decodeUtf8                                  16.22
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8' . decodeUtf8'                                15.78
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseManyD parseCharUtf8D'                  12.03
```

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
+{-# INLINE parseCharUtf8D' #-}
+parseCharUtf8D' :: Monad m => ParserD.Parser m Word8 Char
+parseCharUtf8D' =
+    let A.Array _ ptr _ = utf8d
+    in ParserD.Parser (step' ptr) initial extract
+
+    where
+
+    {-# INLINE prefix #-}
+    prefix = "Streamly.Internal.Data.Stream.parseCharUtf8WithD:"
```
harendra-kumar, line 445: INLINE pragma is not needed on this.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
 replacementChar :: Char
 replacementChar = '\xFFFD'
 
+data UTF8CharDecodeState a
+    = UTF8CharDecodeInit
+    | UTF8CharDecoding !DecodeState !CodePoint
+
+-- XXX We can chain this with another parser to propagate the error and act
+-- accordingly.
```
harendra-kumar, line 435: This comment can be removed. If it is informative only comment then XXX must not be used.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
+    where
+
+    {-# INLINE prefix #-}
+    prefix = "Streamly.Internal.Data.Stream.parseCharUtf8WithD:"
+
+    {-# INLINE initial #-}
+    initial = return $ ParserD.IPartial UTF8CharDecodeInit
+
+    {-# INLINE extract #-}
+    extract _ = error $ prefix ++ "Not enough input"
```
harendra-kumar, line 451: Move `extract` after `step`, we always write these as per the flow, initial, then step, then extract.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
+            True ->
+                let (Tuple' sv cp) = decode0 table x
+                 in case sv of
+                        12 ->
+                            let msg = prefix
+                                    ++ "Invalid first UTF8 byte" ++ show x
+                             in return $ ParserD.Error msg
+                        0 -> error $ prefix ++ "unreachable state"
+                        _ ->
+                            return $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
```
harendra-kumar, line 471: return can be factored out before `case`.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
+            12 ->
+                let msg = prefix
+                        ++ "Invalid subsequent UTF8 byte"
+                        ++ show x
+                        ++ "in state"
+                        ++ show statePtr
+                        ++ "accumulated value"
+                        ++ show codepointPtr
+                 in return $ ParserD.Error msg
+            _ -> return $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
```
harendra-kumar, line 486: return can be factored out before `case`.

In fact, since initial, step, extract all are pure we can use our new `parser` smart constructor to make this parser.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -15,6 +15,8 @@ module Streamly.Internal.Unicode.Stream
       decodeLatin1
 
     -- ** UTF-8 Decoding
+    , writeCharUtf8'
+    , parseCharUtf8D'
```
harendra-kumar, line 19: Need to export parseCharUtf8 i.e. Parser type and not a ParserD type routine.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
+                        ++ show x
+                        ++ "in state"
+                        ++ show statePtr
+                        ++ "accumulated value"
+                        ++ show codepointPtr
+                 in return $ ParserD.Error msg
+            _ -> return $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
+
+-- XXX We cannot implement a "writeCharUtf8" as "TransliterateCodingFailure"
+-- requires backtracking of an element.
```
harendra-kumar, line 489: `XXX` comments are for fixing something broken, so we need to suggest what is to be done instead of making a statement. We can write "This should accept CodingFailure argument but as of now TransliterateCodingFailure requires backtracking so cannot be implemented as a Fold, once folds support backtracking we can revisit and implement that."

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +427,70 @@ data CodingFailureMode
...
 
+data UTF8CharDecodeState a
+    = UTF8CharDecodeInit
+    | UTF8CharDecoding !DecodeState !CodePoint
+
+-- XXX We can chain this with another parser to propagate the error and act
+-- accordingly.
+{-# INLINE parseCharUtf8D' #-}
+parseCharUtf8D' :: Monad m => ParserD.Parser m Word8 Char
+parseCharUtf8D' =
```
harendra-kumar, line 438: We can make the parser accept CodingFailure as an argument and do error handling appropriately. The higher level parser wrapping this routine can choose an appropriate value for that argument.

harendra-kumar (changes requested):

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: I like to write initial and extract together as in most cases they are rather small. But yes, keeping the flow is much better and cleaner

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: I've had to introduce another constructor in the state for this. But that shouldn't be a problem I guess.

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: Will do. I'm exporting `parseCharUtf8D'` as well for benchmarking.

adithyaov:
This is a test comment made from magit

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: I find this to be more elegant.
```
parseCharUtfD = parseCharUtfD' <|> fromPure replacementChar
```

harendra-kumar: All I want is to reuse the same implementation for fold. The low level routine `parserCharInternal options -> Parser m Word8 Char` can be used to express both.

```
parserCharUtf8 = parserCharInternal ErrorOnCodingFailure
foldCharUtf8 = toFold $ parserCharInternal DropOnCodingFailure
```

Now you can use parseCharUtf8 in an alternative composition or whatever. Not that parserCharUtf8 would use the Error constructor in all the error cases and not "error" so that it is always a proper parser, the toFold routine then converts the Error to error.

adithyaov:
```
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8' . foldMany writeCharUtf8'                    16.85
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8' . decodeUtf8'                                15.89
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseMany parseCharUtf8'                    13.37
```

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: It fuses without the `ANN`

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +429,88 @@ data CodingFailureMode
...
+        case cfm of
+            ErrorOnCodingFailure -> return $ ParserD.Error e
+            TransliterateCodingFailure ->
+                return $ case s of
+                    Just _ -> ParserD.Done 1 replacementChar
+                    Nothing -> ParserD.Continue 0 UTF8CharDecodeInit
+            DropOnCodingFailure ->
+                case s of
+                    Just x -> step' table UTF8CharDecodeInit x
+                    Nothing -> return $ ParserD.Partial 0 UTF8CharDecodeInit
```
harendra-kumar, line 460: This should be `Continue 0 UTF8CharDecodeInit` as there is no usable partial result being produced.

harendra-kumar: after removing `step'` the handleError function should no longer be monadic.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +429,88 @@ data CodingFailureMode
...
+                let (Tuple' sv cp) = decode0 table x
+                 in case sv of
+                        12 ->
+                            let msg = prefix
+                                    ++ "Invalid first UTF8 byte" ++ show x
+                             in handleError table msg Nothing
+                        0 -> return
+                                 $ ParserD.Error $ prefix ++ "unreachable state"
+                        _ -> return
+                                 $ ParserD.Continue 0 (UTF8CharDecoding sv cp)
```
harendra-kumar, line 481: When handleError is not  monadic, `return` can be factored out at top level.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +429,88 @@ data CodingFailureMode
...
+-- error handling. This isn't possible now as "TransliterateCodingFailure"'s
+-- workflow requires backtracking 1 element. This can be revisited once "Fold"
+-- supports backtracking.
+{-# INLINE writeCharUtf8' #-}
+writeCharUtf8' :: MonadThrow m => Fold m Word8 Char
+writeCharUtf8' =  ParserD.toFold (parseCharUtf8WithD' ErrorOnCodingFailure)
+
+{-# INLINE parseCharUtf8' #-}
+parseCharUtf8' :: MonadCatch m => Parser.Parser m Word8 Char
+parseCharUtf8' = ParserK.toParserK (parseCharUtf8WithD' ErrorOnCodingFailure)
```
harendra-kumar, line 512: In case, the alternative composition is not be very efficient we may have to pass the error option to this parser which is guaranteed to be efficient. You can try measuring the performance of alternative to decide.

harendra-kumar (changes requested):

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +429,88 @@ data CodingFailureMode
...
+    handleError table e s =
+        case cfm of
+            ErrorOnCodingFailure -> return $ ParserD.Error e
+            TransliterateCodingFailure ->
+                return $ case s of
+                    Just _ -> ParserD.Done 1 replacementChar
+                    Nothing -> ParserD.Continue 0 UTF8CharDecodeInit
+            DropOnCodingFailure ->
+                case s of
+                    Just x -> step' table UTF8CharDecodeInit x
```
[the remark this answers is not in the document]
adithyaov: Ah ofcourse, this was initially implemented as a fold.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +429,96 @@ data CodingFailureMode
...
+--
+-- But unfortunately parseCharUtf8Smart used in conjunction with "parseMany" -
+-- that is "parseMany parseCharUtf8Smart" on a stream causes the heap to
+-- overflow. Even a heap size of 500 MB was not sufficient.
+--
+-- This needs to be investigated futher.
+{-# INLINE parseCharUtf8 #-}
+parseCharUtf8 :: MonadCatch m => Parser.Parser m Word8 Char
+parseCharUtf8 =
+    ParserK.toParserK (parseCharUtf8WithD TransliterateCodingFailure)
```
harendra-kumar, line 520: For now, we can pass the error handling option argument to this parser so that it is flexible.

src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -423,6 +430,96 @@ data CodingFailureMode
...
+    {-# INLINE extract #-}
+    extract _ = error $ prefix ++ "Not enough input"
+
+-- XXX This should ideally accept a "CodingFailureMode" and perform appropriate
+-- error handling. This isn't possible now as "TransliterateCodingFailure"'s
+-- workflow requires backtracking 1 element. This can be revisited once "Fold"
+-- supports backtracking.
+{-# INLINE writeCharUtf8' #-}
+writeCharUtf8' :: MonadThrow m => Fold m Word8 Char
+writeCharUtf8' =  ParserD.toFold (parseCharUtf8WithD ErrorOnCodingFailure)
```
harendra-kumar, line 506: Do we need this? It is just `Parser.toFold (parseCharUtf8With ErrorOnCodingFailure)`.

harendra-kumar (approved):

src/Streamly/Internal/Unicode/Stream.hs
adithyaov: For completion, it's nice to have this. We have a `readCharUtf8` I believe.

Issue #1402, opened 2021-12-30
Change the signature of array getIndices


harendra-kumar:
The current signature is:

```
getIndices :: (MonadIO m, Storable a) => Unfold m (Array a) Int -> Unfold m (Array a) a
```

Make it:

```
getIndices :: (MonadIO m, Storable a) => SerialT m Int -> Unfold m (Array a) a
```

Usually where we generate a stream the return type should be Unfold, and where we consume a stream the argument type should be a stream type. In generation, unfold is more general as it can be converted to stream, and in consumption stream is more general for the same reason, unfold can be converted to stream.

PR #1437, opened 2022-02-02
Change the signature of array getIndices

rnjtranjan:


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1102,18 +1102,17 @@ data GetIndicesState contents start end st =
 -- /Pre-release/
 {-# INLINE getIndices #-}
 getIndices :: (MonadIO m, Storable a) =>
-    Unfold m (Array a) Int -> Unfold m (Array a) a
-getIndices (Unfold stepi injecti) = Unfold step inject
+    D.Stream m Int -> Unfold m (Array a) a
```
harendra-kumar, line 1105: This should be `SerialT m Int`.

src/Streamly/Internal/Data/Array/Foreign.hs
```diff
@@ -459,8 +459,9 @@ getIndex arr i =
 --
 -- /Unimplemented/
 {-# INLINE getIndices #-}
-getIndices :: Unfold m (Array a) Int -> Unfold m (Array a) a
-getIndices = undefined
+getIndices :: (MonadIO m, Storable a)
+    => SerialT m Int -> Unfold m (MA.Array a) a
+getIndices (SerialT stream) = MA.getIndices $ D.fromStreamK stream
```
harendra-kumar, line 464: This is immutable array module, this API should use the immutable array type, not `MA.Array`.

harendra-kumar (changes requested):

rnjtranjan:
Fixes : #1402

src/Streamly/Internal/Data/Array/Foreign.hs
```diff
@@ -457,10 +462,34 @@ getIndex arr i =
...
+            D.Stop -> do
+                liftIO $ touch (arrContents arr)
+                return D.Stop
+
 {-# INLINE getIndices #-}
-getIndices :: Unfold m (Array a) Int -> Unfold m (Array a) a
-getIndices = undefined
+getIndices :: (MonadIO m, Storable a)
+    => SerialT m Int -> Unfold m (Array a) a
+getIndices (SerialT stream) = getIndices' $ D.fromStreamK stream
```
harendra-kumar, line 492: Just use unsafeFreeze/unsafeThaw to convert mutable/immutable arrays, do not duplicate the code.

harendra-kumar (changes requested):

harendra-kumar:
It would be nice to have a test as well, we can copy the takeWhile test if it exists.

Issue #1387, opened 2021-12-26
Add a takeEndBy transformation


harendra-kumar:
Like `takeWhile` but with an inverted condition and also taking the matching element.

```haskell
takeEndBy :: (a -> Bool) -> SerialT m a -> SerialT m a
```

We have a `takeEndBy` fold but a stream transformation would be more convenient when we want such an operation in a stream transformation code.

PR #1438, opened 2022-02-02
Add a takeEndBy transformation

rnjtranjan:


src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -638,6 +639,19 @@ takeWhile p = go
                        | otherwise = stp
          in foldStream st yieldk single stp m1
 
+{-# INLINE takeEndBy #-}
+takeEndBy :: (a -> Bool) -> Stream m a -> Stream m a
+takeEndBy p = go Nothing
+    where
+    go Nothing m1 = mkStream $ \st yld sng stp ->
+        let single a   | not $ p a = sng a
+                       | otherwise = stp
```
harendra-kumar, line 648: It should be `sng a` in both cases.

src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -638,6 +639,19 @@ takeWhile p = go
...
+            yieldk a r | not $ p a = yld a (go Nothing r)
+                       | otherwise = yld a (go (Just a) r)
```
harendra-kumar, line 650: This code is incorrect. 

There is no need to have the `Maybe` argument in `go`.

The code should be:

```
            yieldk a r | not $ p a = yld a (go r)
                       | otherwise = sng a
```

harendra-kumar (changes requested):

rnjtranjan:
Fixes: #1387

src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -645,22 +645,22 @@ takeWhile f = takeWhileM (return . f)
...
-                    then Yield x (s, Nothing)
-                    else Yield x (s, Just x)
-            Skip s -> return $ Skip (s, Nothing)
+                    then Yield x (Just s)
+                    else Yield x Nothing
+            Skip s -> return $ Skip (Just s)
             Stop   -> return Stop
 
-    step' _ (_, _) = return Stop
+    step' _ _ = return Stop
```
harendra-kumar, line 663: An explicit `Nothing` is better than a wildcard to understand the case.

src/Streamly/Internal/Data/Stream/IsStream/Common.hs
```diff
@@ -430,6 +431,10 @@ take n m = fromStreamS $ S.take n $ toStreamS
 takeWhile :: (IsStream t, Monad m) => (a -> Bool) -> t m a -> t m a
 takeWhile p m = fromStreamS $ S.takeWhile p $ toStreamS m
 
+{-# INLINE takeEndBy #-}
+takeEndBy :: (IsStream t, Monad m) => (a -> Bool) -> t m a -> t m a
+takeEndBy p m = fromStreamS $ S.takeEndBy p $ toStreamS m
```
harendra-kumar, line 436: Since you have removed StreamK impl, you can use D.takeEndBy .

src/Streamly/Internal/Data/Stream/IsStream/Common.hs
```diff
@@ -433,7 +433,7 @@ takeWhile p m = fromStreamS $ S.takeWhile p $ toStreamS m
 
 {-# INLINE takeEndBy #-}
 takeEndBy :: (IsStream t, Monad m) => (a -> Bool) -> t m a -> t m a
-takeEndBy p m = fromStreamS $ S.takeEndBy p $ toStreamS m
+takeEndBy p m = fromStreamS $ D.takeEndBy p $ toStreamS m
```
harendra-kumar, line 436: You need to use fromStreamD and toStreamD instead of fromStreamS et al. Streamk CI is failing because of this.

harendra-kumar (changes requested):

harendra-kumar:
It would be nice to have a test as well, we can copy the takeWhile test if it exists.

harendra-kumar:
Can you squash the changes?

Issue #1224, opened 2021-09-08
Add a test for wAsync behavior check


harendra-kumar:
WAsync broken behavior was caught by doctests but not by tests. We can add a sanity test to check it:

```
[ 180.79421] error: Stream.toList $ Stream.fromWAsync $ Stream.maxThreads 1 $ stream1 `wAsync` stream2
[ 180.79468] error: in comment at 497:1
expected: [1,4,2,5,3,6]
 but got: [1,2,3,4,5,6]
             ^
```

There are some tests in the WAsync test module but they are commented out.

rnjtranjan:
It's not reproducible. is it a random behavior?

harendra-kumar:
Now it has been fixed, it was during a change that broke the code.

PR #1447, opened 2022-02-03
Add a test for wAsync behavior check

rnjtranjan:


rnjtranjan:
Fixes: #1224

test/lib/Streamly/Test/Prelude/Common.hs
```diff
@@ -290,6 +291,18 @@ constructWithFromIndicesM op len =
         streamEffect <- run $ readIORef mvl
         listEquals (==) streamEffect list
 
+constructfromWAsync ::
+    S.WAsyncT IO Int -> S.WAsyncT IO Int-> [Int] -> Property
+constructfromWAsync s1 s2 res =
```
adithyaov, line 296: Can you also check the asyncness of the code?
An `IORef` to capture the effects would work.
```
gen i = threadDelay i >> addToIORefStuff >> return i
```
Then you can use `fromListM` and check the IORef for the expected order of effects.

test/Streamly/Test/Prelude/Async.hs
```diff
@@ -56,6 +56,14 @@ main = hspec
...
         asyncOps $ prop "asyncly (.:)" . constructWithCons (S..:)
         asyncOps $ prop "asyncly (|:)" . constructWithConsM (S.|:) sort
+        prop "asyncSingleThreaded" $
+            constructfromAsyncSingleThread
+            (S.fromList [1,2,3,4,5])
+            (S.fromList [6,7,8,9,10])
+            [1,2,3,4,5,6,7,8,9,10]
+
+        prop "asyncMultiThreaded" $
+            constructfromAsyncNThread 4 2 [2,4]
```
harendra-kumar, line 66: This is already being at line 127 below using `parallelCheck`. We can remove it.

test/lib/Streamly/Test/Prelude/Common.hs
```diff
@@ -290,6 +293,41 @@ constructWithFromIndicesM op len =
...
+    withMaxSuccess maxTestCount $
+    monadicIO $ do
+        x <-  run
+            $ S.toList
+            $ S.fromWAsync
+            $ S.maxThreads 1
+            $ s1 `S.wAsync` s2
+        equals (==) x res
```
harendra-kumar, line 306: We can move this code to WAsync test module itself as this is specific to wAsync and not common.

test/lib/Streamly/Test/Prelude/Common.hs
```diff
@@ -290,6 +293,41 @@ constructWithFromIndicesM op len =
...
+    S.AsyncT IO Int -> S.AsyncT IO Int-> [Int] -> Property
+constructfromAsyncSingleThread s1 s2 res =
+    withMaxSuccess maxTestCount $
+    monadicIO $ do
+        x <-  run
+            $ S.toList
+            $ S.fromAsync
+            $ S.maxThreads 1
+            $ s1 `S.async` s2
+        equals (==) x res
```
harendra-kumar, line 318: We can move this code to Async test module itself as this is specific to Async and not common.

harendra-kumar (changes requested):

PR #1449, opened 2022-02-03
Replace unsafeWithArrayContents with unsafeWithArray

adithyaov:
This makes the API less ambiguous


src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -333,12 +333,10 @@ arrayToFptrContents (ArrayContents contents) = PlainPtr contents
...
 
 -- | Similar to unsafeWithForeignPtr.
-{-# INLINE unsafeWithArrayContents #-}
-unsafeWithArrayContents :: MonadIO m =>
-    ArrayContents -> Ptr a -> (Ptr a -> m b) -> m b
-unsafeWithArrayContents contents ptr f = do
-  r <- f ptr
-  liftIO $ touch contents
+unsafeWithArray :: MonadIO m => Array a -> (Ptr a -> m b) -> m b
+unsafeWithArray Array{..} f = do
```
harendra-kumar, line 337: We should rename this to `asPtrUnsafe` and remove the current `asPtrUnsafe`.

src/Streamly/Internal/Data/Array/Foreign/Type.hs
```diff
@@ -158,6 +159,13 @@ data Array a =
 foreign import ccall unsafe "string.h strlen" c_strlen
     :: CString -> IO CSize
 
+-- | Similar to unsafeWithForeignPtr.
+unsafeWithArray :: MonadIO m => Array a -> (Ptr a -> m b) -> m b
+unsafeWithArray Array{..} f = do
+  r <- f arrStart
+  liftIO $ touch arrContents
+  return r
```
harendra-kumar, line 167: This should be `unsafeAsPtr`, I know there is a cast in unsafeAsPtr but the users of the API can do the cast themselves when needed.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #1451, opened 2022-02-04
Make applicative code in ParserD.serialWith monadic

adithyaov:


src/Streamly/Internal/Data/Parser/ParserD/Type.hs
```diff
@@ -426,28 +426,30 @@ serialWith func (Parser stepL initialL extractL)
 
     -- Note: For the composed parse to terminate, the left parser has to be
     -- a terminating parser returning a Done at some point.
-    step (SeqParseL st) a =
-      (\resL initR ->
+    step (SeqParseL st) a = do
```
harendra-kumar, line 429: We should write a comment here saying "Please do not change this to applicative code, see issue xyz".

harendra-kumar:
Benchmarks look good, there is no change at all.

harendra-kumar (approved):

PR #1452, opened 2022-02-04
Implement zipWith in terms of zipWithM in StreamK

adithyaov:


src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -751,19 +751,8 @@ mapMaybe f = go
...
-
-    go mx my = mkStream $ \st yld sng stp -> do
-        let merge a ra =
-                let single2 b = sng (f a b)
-                    yield2 b rb = yld (f a b) (go ra rb)
-                 in foldStream (adaptState st) yield2 single2 stp my
-        let single1 a = merge a nil
-            yield1 = merge
-        foldStream (adaptState st) yield1 single1 stp mx
+zipWith :: Monad m => (a -> b -> c) -> Stream m a -> Stream m b -> Stream m c
```
adithyaov, line 754: This required adding a `Monad` constraint here.

harendra-kumar:
In the benchmarks it would be nice to run `bench.sh --no-measure` once to create the diff comparison output of all the benchmarks. it would be convenient to see the comparisons. I had to go find both the runs of zipWith and compare the timings.

src/Streamly/Internal/Data/Stream/StreamK.hs
harendra-kumar: This is an internal change only, so should be fine in minor release.

adithyaov:
> In the benchmarks it would be nice to run bench.sh --no-measure once to create the diff comparison output of all the benchmarks. it would be convenient to see the comparisons. I had to go find both the runs of zipWith and compare the timings.

Yes, I wanted to make this change in the regression checking CI. Quickly list all the benchmarks, check for regressions after.

harendra-kumar (approved):

PR #1459, opened 2022-02-04
Add a dumb script to repl all the executables and library

adithyaov:
Running this in a loop in a smart way will work like `ghcid`.

The idea is to have an extremely light version of `run-ci.sh`
This can be very helpful in any refactoring change.

This script should be majorly improved but it solves my use case for the time being.

This need not even be merged if we think this will cause some maintenance burden. I'm not going to work on this script for the time being. All improvements are postponed to a later date. If anyone wants to just quickly use it, they can!


harendra-kumar:
Can you make the CI's pass and commit it?

PR #1464, opened 2022-02-05
Add some patches that need to be commited in the next release

adithyaov:


adithyaov:
As discussed, it is easier and better to use the commits with a label instead.
Add a commit labeled `NEXT-REL` and revert it.

PR #1466, opened 2022-02-06
Fix broken links and update MAINTAINENCE guide

adithyaov:


dev/utf8-decoder.md
```diff
@@ -404,7 +404,7 @@ the non-ASCII starter is swapped (approximately a 5% increase), but GCC
 I have experimented with various rearrangements of states and character
 classes. A seemingly promising one is the following:
 
-![Re-arranged DFA with class transitions](/design/dfa-rearr.png)
+![Re-arranged DFA with class transitions](./dfa-rearr.png)
```
harendra-kumar, line 407: We can omit the `./`?

harendra-kumar (approved):

dev/utf8-decoder.md
adithyaov: We can.

PR #1476, opened 2022-02-12
Add unfoldMany in Fold

adithyaov:


benchmark/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -210,6 +211,16 @@ unzipWithMinM = do
     let f = \a -> return (a + 1, a)
     IP.fold (FL.unzipWithMinM f FL.sum FL.length)
 
+-------------------------------------------------------------------------------
+-- Nested
+-------------------------------------------------------------------------------
+
+unfoldMany :: Int -> Benchmarkable
```
harendra-kumar, line 218: Its taking 1.88ms. INLINE it and see if it improves perf.

PR #1486, opened 2022-02-19
Add writeCharUtf8Rev'

adithyaov:


harendra-kumar:
We can use the same code as we are using in ParseCharUtf8WithD. We need to buffer at most 2 chars in the Fold state, as soon as we see the third char we can either decode or fail. As soon as we have sufficient number of chars required for decoding we can run the state machine as in ParseCharUtf8WithD.

adithyaov:
The motivation is the API `last`

adithyaov:
See https://github.com/composewell/streamly/issues/1527

PR #1490, opened 2022-02-20
Improve the documentation of intersperseM

adithyaov:


src/Streamly/Internal/Data/Stream/IsStream/Common.hs
```diff
@@ -463,6 +463,11 @@ findIndices p m = fromStreamS $ S.findIndices p (toStreamS m)
 -- >>> Stream.toList $ Stream.trace putChar $ Stream.intersperseM (putChar '.' >> return ',') $ Stream.fromList "hello"
 -- h.,e.,l.,l.,o"h,e,l,l,o"
 --
+-- Please note that the interspersed effect is offset by 1.
+--
+-- >>> Stream.toList $ Stream.intersperseM (putChar '.' >> return ',') $ Stream.mapM (\x -> putChar x >> return x) $ Stream.fromList "hello"
+-- he.l.l.o."h,e,l,l,o"
```
harendra-kumar, line 469: Not sure what additional information is this snippet trying to convey. The existing doc looks precise and correct.

This new snippet generates `he.l.l.o.` effect output because it is doing a mapM first which prints the stream element and only after that the effect introduced by intersperseM is executed. So the offset is an artifact of the way this code is written and not of intersperseM itself.

harendra-kumar: We can write like this - Be careful about the order of effects. In the above example we used `trace` after the intersperse, if we use it before the intersperse the output would be `he.l.l.o."h,e,l,l,o"`.

Issue #1471, opened 2022-02-10
uniqBy is broken


harendra-kumar:
intersectBy tests are failing, the root cause is in uniqBy which is using rollingMap.

```
> Stream.toList $ Stream.intersectBy (==) (Stream.fromList [8012,2486,409,3329,1519,4193]) (Stream.fromList [4193])
[]
> Stream.toListRev $ Stream.uniqBy (==) (Stream.fromList [4193])
[]
> Stream.toListRev $ Stream.rollingMap (\x y -> if x == y then Nothing else Just y) (Stream.fromList [4193])
[]
```

rollingMap docs say that it returns an empty stream when the input stream has a single element. Need to think if that should be changed or we should not using rollingMap in uniqBy. Also, how to avoid this pitfall again.

PR #1501, opened 2022-03-03
Fix uniqBy and rollingMap

rnjtranjan:


rnjtranjan:
Fixes: #1471

src/Streamly/Internal/Data/Stream/StreamD/Transform.hs
```diff
@@ -1203,28 +1203,28 @@ slicesBy p (Stream step1 state1) = Stream step (Just (state1, 0, 0))
 data RollingMapState s a = RollingMapInit s | RollingMapGo s a
 
 {-# INLINE rollingMapM #-}
-rollingMapM :: Monad m => (a -> a -> m b) -> Stream m a -> Stream m b
+rollingMapM :: Monad m => (a -> a -> m (Maybe a)) -> Stream m a -> Stream m (Maybe a)
```
harendra-kumar, line 1206: With this change rollingMapM just became useless for all other use cases, it is custom designed only for uniqBy use case. That's because, (1) the return type is restricted to be the same as the input type, (2)  the first element is always `Just`. Also the signature of rollingMap returns `a` whereas rollingMapM returns `Maybe a`, this is not expected.

Instead we can make the signature as:

```
rollingMapM :: Monad m => (Maybe a -> a -> m b) -> Stream m a -> Stream m b
```

The `Maybe` argument is the previous element, and the second argument is the current element. For the first element the previous element would be Nothing, for subsequent elements it would be Just.

harendra-kumar: We can keep the original function as well, we can call it rollingMapM2 because it requires minimum 2 elements. It could be defined as:

```
rollingMapM2 :: Monad m => (a -> a -> m b) -> Stream m a -> Stream m b
rollingMapM2 f = catMaybes . rollingMapM g

  where

   g Nothing _ = Nothing
   g (Just x) y = Just (f x y)
```

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -921,8 +922,13 @@ filterM p m = fromStreamD $ D.filterM p $ toStreamD m
...
     (a -> a -> Bool) -> t m a -> t m a
-uniqBy eq =
-    catMaybes . rollingMap (\x y -> if x `eq` y then Nothing else Just y)
+uniqBy eq = catMaybes . rollingMap f
+
+    where
+
+    f pre curr = case pre of
+                    Nothing -> Just curr
+                    Just x -> if x `eq` curr then Nothing else Just curr
```
harendra-kumar, line 931: Bring `case ...` on the next line and indent by 4 spaces.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -1519,17 +1525,20 @@ elemIndices a = findIndices (== a)
...
 -- /Pre-release/
 --
 {-# INLINE rollingMapM #-}
-rollingMapM :: (IsStream t, Monad m) => (a -> a -> m b) -> t m a -> t m b
+rollingMapM :: (IsStream t, Monad m) => (Maybe a -> a -> m b) -> t m a -> t m b
 rollingMapM f m = fromStreamD $ D.rollingMapM f $ toStreamD m
 
+{-# INLINE rollingMap2 #-}
+rollingMap2 :: (IsStream t, Monad m) => (a -> a -> b) -> t m a -> t m b
+rollingMap2 f m = fromStreamD $ D.rollingMap2 f $ toStreamD m
```
harendra-kumar, line 1541: Add a blank line after this.

src/Streamly/Internal/Data/Stream/StreamD/Transform.hs
```diff
@@ -1203,29 +1204,42 @@ slicesBy p (Stream step1 state1) = Stream step (Just (state1, 0, 0))
 data RollingMapState s a = RollingMapInit s | RollingMapGo s a
 
 {-# INLINE rollingMapM #-}
-rollingMapM :: Monad m => (a -> a -> m b) -> Stream m a -> Stream m b
+rollingMapM :: Monad m => (Maybe a -> a -> m b) -> Stream m a -> Stream m b
 rollingMapM f (Stream step1 state1) = Stream step (RollingMapInit state1)
+
```
harendra-kumar, line 1209: Just use `RollingMapGo state1 Nothing` here. The RollingMapInit state can be removed.

harendra-kumar (changes requested):

harendra-kumar:
Also, squash the commits once done.

harendra-kumar:
Documentation is not correct for rollingMap and absent for rollngMap2. I will fix it and push.

PR #1505, opened 2022-03-06
All major release changes

adithyaov:


src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
```diff
@@ -223,7 +223,7 @@ unfoldr step seed = fromStreamS (S.unfoldr step seed)
 --
 -- /Since: 0.1.0/
 {-# INLINE_EARLY unfoldrM #-}
-unfoldrM :: forall t m b a. (IsStream t, MonadAsync m) =>
+unfoldrM :: forall t m a b. (IsStream t, MonadAsync m) =>
```
harendra-kumar, line 226: We can skip this change. The order seems ok, we can swap the names of `a ` and `b` in the signature if we want `t m a b.` here.

adithyaov: This is only to satisfy `cabal-diff`

harendra-kumar (approved):

PR #1506, opened 2022-03-07
Extract the core modules into a seperate package

adithyaov:
Few files have been duplicated
The exposed core modules have been re-exposed

We probably want to re-export everything in streamly-core. We need to figure out
the best way to do it. Keeping the same module space is going to be tiresome as
all the internal files need to be duplicated. See `Streamly.Data.Fold` for
example.

I suggest, make another module `Streamly.Internal.Core` to re-export everything
in `streamly-core`? This again pollutes the namespace.

Need to think of a better way to do this?


core/src/Streamly/Internal/Data/Stream/PreludeCommon.hs
```diff
@@ -0,0 +1,139 @@
+------------------------------------------------------------------------------
```
harendra-kumar, line 1: Raise an issue to remove this module and the Data.Stream.Prelude module, and assimilate these into Serial module. Target the issue to 0.9.0.

We can also assimilate Instances.hs in the serial module, only those instances that are common with the Zip module or any other module in this package can be kept in the common CPP header.

.github/workflows/haskell.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     continue-on-error: ${{ matrix.ignore_error }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
harendra-kumar, line 53: We can revert this?

harendra-kumar (commented):
Can you raise an issue about moving relevant benchmarks and tests to streamly-core.

Are there any changes in `src/config.h.in`?

core/streamly-core.cabal
```diff
@@ -0,0 +1,379 @@
...
+
+    if os(darwin)
+      cpp-options:    -DCABAL_OS_DARWIN
+
+    if os(linux)
+      cpp-options:    -DCABAL_OS_LINUX
+
+    if os(windows)
+      cpp-options:    -DCABAL_OS_WINDOWS
+
```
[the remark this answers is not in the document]
adithyaov: Yes, I kept them as they are more generalized flags.

core/streamly-core.cabal
```diff
@@ -0,0 +1,379 @@
+cabal-version:      2.2
+name:               streamly-core
+version:            0.8.2
```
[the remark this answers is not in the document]
adithyaov: Yes.

adithyaov:
> Can you raise an issue about moving relevant benchmarks and tests to streamly-core.

Will do.

> Are there any changes in src/config.h.in?

Yes, you can see the diff. Was regenerated using `autoheaders`.

harendra-kumar:
Benchmarks are failing to run:

```
All.Unicode.Stream/o-1-space.decode-encode.encodeUtf8 . parseMany parseCharUtf8 (1/10):   +RTS -T -K36K -M16M    -RTS      
Using small input file: benchmark-tmp/in-10MB.txt
Using big input file: benchmark-tmp/in-100MB.txt
Using output file: benchmark-tmp/out.txt
All
  Unicode.Stream/o-1-space
    decode-encode
      encodeUtf8 . parseMany parseCharUtf8 (1/10): FAIL
        Exception: Streamly.Internal.Data.Stream.parseCharUtf8WithD:Not enough input
        CallStack (from HasCallStack):
          error, called at src/Streamly/Internal/Unicode/Stream.hs:498:17 in streamly-0.8.2-inplace:Streamly.Internal.Unicode.Stream

1 out of 1 tests failed (0.25s)
Error: Benchmark execution failed.
Error: Benchmark execution failed.
Error: Process completed with exit code 1.
```

streamly.cabal
```diff
@@ -391,13 +547,13 @@ library
...
-                     , Streamly.Internal.Data.Parser.ParserK.Type
-                     , Streamly.Internal.Data.Parser.ParserD.Type
-                     , Streamly.Internal.Data.Pipe.Type
+                     , streamly-core:Streamly.Internal.Data.Producer.Type
+                     , streamly-core:Streamly.Internal.Data.Producer
+                     , streamly-core:Streamly.Internal.Data.Producer.Source
+                     , streamly-core:Streamly.Internal.Data.Sink.Type
+                     , streamly-core:Streamly.Internal.Data.Parser.ParserK.Type
+                     , streamly-core:Streamly.Internal.Data.Parser.ParserD.Type
+                     , streamly-core:Streamly.Internal.Data.Pipe.Type
```
harendra-kumar, line 556: Why only these modules have a `streamly-core:` prefix, and not others?

harendra-kumar (commented):
Looks good. Need to check the benchmarks once fixed.

streamly.cabal
adithyaov: Ah, they should be removed. I forgot to remove all of them.

adithyaov:
The benchmark failure seems to be some temporary failure.
It does not happen on the second attempt
I am unable to reproduce it locally

harendra-kumar:
> The benchmark failure seems to be some temporary failure. It does not happen on the second attempt I am unable to reproduce it locally

There has to be a reason for it. We can figure it out. Otherwise it will keep troubling us.

harendra-kumar (approved):

PR #1507, opened 2022-03-08
Remove Deprecated Modules

rnjtranjan:


harendra-kumar:
We should remove only after at least two major releases of deprecation that means it has to be deprecated in 0.7 and 0.8, then we can remove in 0.9.

We can remove Streamly.Internal.Data.Time module for example because it remained deprecated in 0.7 and 0.8.

harendra-kumar:
As I said earlier, we should not remove too early. We should remove only if the modules/functions were deprecated in 0.7 or earlier. Many of the functions removed in this PR were deprecated only in 0.8.

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -705,29 +702,6 @@ postscan fld = fromStreamD . D.postscanOnce fld . toStreamD
...
 
--- XXX It may be useful to have a version of scan where we can keep the
--- accumulator independent of the value emitted. So that we do not necessarily
--- have to keep a value in the accumulator which we are not using. We can pass
--- an extraction function that will take the accumulator and the current value
--- of the element and emit the next value in the stream. That will also make it
--- possible to modify the accumulator after using it. In fact, the step function
--- can return new accumulator and the value to be emitted. The signature would
--- be more like mapAccumL. Or we can change the signature of scanx to
--- accommodate this.
```
harendra-kumar, line 716: We can keep this TODO comment.

harendra-kumar (changes requested):
I do not see Streamly or Streamly.Data.Unicode.Stream or Streamly.Memory.Array being deprecated in 0.7.0.

harendra-kumar:
Please rebase to resolve the conflicts.

src/Streamly/Memory/Array.hs
```diff
@@ -71,4 +71,4 @@ module Streamly.Memory.Array
     )
 where
 
-import Streamly.Internal.Data.Array.Foreign as A
+import Streamly.Internal.Data.Array.Foreign as A
```
harendra-kumar, line 74: This file should not have any change, should be reverted.

harendra-kumar (changes requested):

harendra-kumar:
Please squash the changes.

harendra-kumar:
Can we add a changelog entry, saying removal of function deprecated in or before 0.7.0?

rnjtranjan:
Fixes: #1477

docs/Changelog.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 <!-- See rendered changelog at https://streamly.composewell.com -->
+## 0.9.0 (Mar 2022)
```
harendra-kumar, line 4: We use "Unreleased" here instead of 0.9.0 .

harendra-kumar:
It seems many are still running 0.7.x, so let's remove only those that were deprecated in or before 0.6 .

rnjtranjan:
opened another PR : https://github.com/composewell/streamly/pull/1552

PR #1516, opened 2022-03-11
Move parCharUtf8 benchmark under the STRICT_UTF8 flag

adithyaov:


harendra-kumar:
Do we ever use this flag in any benchmarking workflow:

```
flag include-strict-utf8
  description: Include strict utf8 input benchmarks
  manual: True
  default: False
```

If we do not use it then it is as good as commenting out the benchmark.

adithyaov:
Comment out for the time being.

harendra-kumar (approved):

harendra-kumar:
Squash the commits.

PR #1523, opened 2022-03-14
Add/update doc/design notes of folds/parsers

harendra-kumar:


core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -12,45 +12,115 @@
...
 --
--- == Accumulators
+-- == Folds that never terminate
 --
--- These are the simplest folds that never fail and never terminate, they
--- accumulate the input values forever and can always accept new inputs (never
--- terminate) and always have a valid result value.  A
--- 'Streamly.Internal.Data.Fold.sum' operation is an example of an accumulator.
+-- An Accumulator is the simplest type of fold, it never fails and never
+-- terminates. It can always accept more inputs (never terminates) and the
```
adithyaov, line 18:
```suggestion
-- terminates. It can always accept more input (never terminates) and the
```

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -12,45 +12,115 @@
...
+--
+-- This is just a traditional left fold, compare with @foldl@. The driver of
+-- the fold would call @initial@ at the beginning and then keep accumulating
+-- inputs into its result using @step@ and finally extract the result using
+-- @extract@.
+--
+-- == Folds that terminate after one or more input
+--
+-- Terminating folds are accumulators that can terminate, but like accumulators
+-- they do not fail. Once a fold terminates it no longer accepts any more
```
adithyaov, line 44:
```suggestion
-- Terminating folds are accumulators that can terminate. Once a fold terminates it no longer accepts any more
```

`but` is rather confusing. I expected a contradiction. `but unlike` makes more sense.

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -12,45 +12,115 @@
...
+-- @
+--
+-- This is just a traditional left fold, compare with @foldl@. The driver of
+-- the fold would call @initial@ at the beginning and then keep accumulating
+-- inputs into its result using @step@ and finally extract the result using
+-- @extract@.
+--
+-- == Folds that terminate after one or more input
+--
+-- Terminating folds are accumulators that can terminate, but like accumulators
```
adithyaov, line 43: `but` is rather confusing for some reason. I expected a contradiction. `but unlike` makes more sense. `and like` makes more sense.

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -12,45 +12,115 @@
...
+-- to pass a fold that never consumes an input. However, we do not have a
+-- separate Fold1 type for the sake of simplicity of types/API.
+--
+-- Adding this capability adds a certain amount of complexity in the
+-- implementation of fold combinators. @initial@ has to always handle two cases
+-- now.  We could potentially not implement this in folds to keep fold
+-- implementation simpler, and these use cases can be transferred to the parser
+-- type. However, it would be a bit inconvenient to not have a `take` operation
+-- or to not be able to use `take 0` if we have it. Also, applicative and
+-- monadic composition of folds would not be possible.
```
adithyaov, line 113: This is more like developer notes.

adithyaov: It's fine I guess.

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -60,30 +130,74 @@
...
+--
+-- many :: Monad m => Fold m b c -> Fold m a b -> Fold m a c
+--
+-- If the inner fold itself returns a @Done1@ then we need to backtrack all
+-- the elements that have been consumed by the outer fold to generate that
+-- value. We need backtracking of more than one element.
+--
+-- Arbitrary backtracking requires arbitrary buffering. However, we do not want
+-- to buffer unconditionally, only if the buffer is needed. One way to do this
+-- is to use a "Continue" ocnstructor like parsers. When we have nested folds,
```
adithyaov, line 151:
```suggestion
-- is to use a "Continue" constructor like parsers. When we have nested folds,
```

core/src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -1156,6 +1156,8 @@ data ParseChunksState x inpBuf st pst =
     | ParseChunksBuf inpBuf st inpBuf !pst
     | ParseChunksYield x (ParseChunksState x inpBuf st pst)
 
+-- XXX This is in fact parseMany1 (a la foldMany1). Do we need a parseMany as
+-- well?
```
adithyaov, line 1160: Can't answer this right away. Need to think.

adithyaov (approved):

core/src/Streamly/Internal/Data/Fold/Type.hs
harendra-kumar: These are developer notes. It is doc of internal module, not supposed to be visible in user docs.

Issue #1168, opened 2021-07-30
Rename a few APIs in FileSystem.Handle module


harendra-kumar:
See the [container api doc](https://github.com/composewell/streamly/blob/master/dev/container-api.md).

Rename:

* toBytes => getBytes
* toBytesWithBufferOf => getBytesWith
* *WithBufferOf => *With

We need to do the same for the FileSystem.File module as well. However, in the File module we will use "toBytes" instead of "getBytes", because the operation is on a stateless object unlike the stateful Handle object.

PR #1524, opened 2022-03-14
Rename a few APIs in FileSystem.Handle module

rnjtranjan:


rnjtranjan:
Fixes: #1168

src/Streamly/Internal/FileSystem/Handle.hs
```diff
@@ -35,16 +35,16 @@ module Streamly.Internal.FileSystem.Handle
...
+    , getBytesWith
 
     -- * Chunked Stream Read
     , readChunks
-    , readChunksWithBufferOf
+    , readChunksWith
 
-    , toChunksWithBufferOf
+    , toChunksWith
     , toChunks
```
harendra-kumar, line 48: Should be `getChunks` and `getChunksWith`.

harendra-kumar (changes requested):

harendra-kumar (changes requested):
Deprecate the released functions instead of changing them.

harendra-kumar (changes requested):
Only those functions that have been released, should be deprecated. See https://streamly.composewell.com/haddocks/streamly-0.8.2/Streamly-FileSystem-Handle.html for released functions.

src/Streamly/FileSystem/Handle.hs
```diff
@@ -76,9 +76,9 @@ module Streamly.FileSystem.Handle
     -- encountered.
 
     , read
-    , readWithBufferOf
+    , readWith
```
harendra-kumar, line 79: `readBufferOf` should be kept with deprecation. And same for other released functions as well.

harendra-kumar (changes requested):

PR #1534, opened 2022-03-16
Add benchmark and Test for scan of Unfold

rnjtranjan:


rnjtranjan:
Fixes : #1313

test/Streamly/Test/Data/Unfold.hs
```diff
@@ -449,6 +450,17 @@ enumerateFromToFractional =
 -- Stream transformation
 -------------------------------------------------------------------------------
 
+scan :: Property
+scan =
+    property
+        $ \f _list ->
+              let _fA = apply f :: Int -> Int
+                  _x = _list :: [Int]
```
harendra-kumar, line 458: This looks ugly, why is `_list` even being passed as argument? Why is `_fA` required.

harendra-kumar (changes requested):

test/Streamly/Test/Data/Unfold.hs
```diff
@@ -449,6 +450,17 @@ enumerateFromToFractional =
...
+                  unf = UF.scan Fold.sum UF.fromList
+                  list = [1,2,3,4] :: [Int]
+                  mList = [1,3,6,10] :: [Int]
```
harendra-kumar, line 461: Can we use the passed list instead and compare it against the Data.List scan function?

test/Streamly/Test/Data/Unfold.hs
```diff
@@ -449,6 +450,14 @@ enumerateFromToFractional =
 -- Stream transformation
 -------------------------------------------------------------------------------
 
+scan :: Property
+scan =
+    property
+        $ \(ls :: [Int]) ->
+              let unf = UF.scan Fold.sum UF.fromList
+                  mList = scanl1 (+) ls
```
harendra-kumar, line 458: Looks like `UF.scan` should be renamed to `UF.postscan`.

Also, I noticed that `UF.scanlM'` is marked as `Unimplemented` even though it is implemented. We can fix that too. Also. UF.scanlM' should be renamed as `postscanlM'`.

Also, rename this test as `postscan`.

harendra-kumar (changes requested):

PR #1536, opened 2022-03-16
Add scanMany to Fold

rnjtranjan:


core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -434,6 +437,50 @@ scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
 
     extract = extractR . snd
 
+-- | Scan the input of a 'Fold' to change it in a stateful manner using another
+-- 'Fold'. If second fold terminates it will restart again from initial state.
+-- /Pre-release/
+scanMany :: Monad m => d -> Fold m a b -> Refold m d b c -> Fold m a c
```
harendra-kumar, line 443: I am not sure about what would be a useful way to implement this using Refold. Can you implement it using Fold only, that is much simpler to understand and use?

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -434,6 +435,50 @@ scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
...
+                case rR of
+                    Partial sR1 -> return $ Partial (sL, sR1, nxt)
+                    Done b1 -> do
+                        if isNothing nxt
+                        then return $ Partial (sL, sR, Just b1)
+                        else do
+                            r <- initialR
+                            case r of
+                                Partial sR1 -> runStep actionL sR1 Nothing
+                                Done b2 -> return $ Done b2
```
harendra-kumar, line 468: This is not correct. We need to scan the input of the right fold using the left fold. So in `Done bL` case we need to initialize the left fold again.

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -439,6 +439,43 @@ scan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
...
+    where
+
+    {-# INLINE runStep #-}
+    runStep actionL sR = do
+        rL <- actionL
+        case rL of
+            Done bL -> do
+                rR <- stepR sR bL
+                case rR of
+                    Partial sR1 -> runStep initialL sR1
```
harendra-kumar, line 460: Only this line is different compared to `scan`. Can we use a common function `scanWith` passing a Bool flag `restart`, and making this line dependent on the bool flag. That way we can reuse the code.

Then:

```
scan = scanWith False
scanMany = scanWith True
```

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -401,25 +401,23 @@ transform (Pipe pstep1 pstep2 pinitial) (Fold fstep finitial fextract) =
...
     runStep actionL sR = do
         rL <- actionL
         case rL of
             Done bL -> do
                 rR <- stepR sR bL
                 case rR of
-                    Partial sR1 -> Done <$> extractR sR1
+                    Partial sR1 -> if isMany
+                                   then runStep initialL sR1
+                                   else Done <$> extractR sR1
```
harendra-kumar, line 420: Bring the whole if block on the next line and indent with 4 spaces.

harendra-kumar (commented):
squash the commits.

harendra-kumar:
It is taking a lot of time compared to scan/postscan:

```
All.Data.Fold/o-1-space.transformation.fold-scan:   +RTS -T -K36K -M16M    -RTS      
All
  Data.Fold/o-1-space
    transformation
      fold-scan: OK (0.12s)
        29.6 μs ± 2.7 μs,   0 B  allocated,   0 B  copied, 2.0 MB peak memory

All 1 tests passed (0.13s)
All.Data.Fold/o-1-space.transformation.fold-postscan:   +RTS -T -K36K -M16M    -RTS      
All
  Data.Fold/o-1-space
    transformation
      fold-postscan: OK (0.12s)
        29.6 μs ± 2.7 μs,   0 B  allocated,   0 B  copied, 2.0 MB peak memory

All 1 tests passed (0.13s)
All.Data.Fold/o-1-space.transformation.fold-scanMany:   +RTS -T -K36K -M16M    -RTS      
All
  Data.Fold/o-1-space
    transformation
      fold-scanMany: OK (0.13s)
        2.11 ms ± 171 μs,  17 MB allocated, 2.7 KB copied, 2.0 MB peak memory
```

benchmark/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -345,6 +345,10 @@ o_1_space_serial_transformation value =
             value
             "fold-postscan"
             (S.fold $ FL.postscan FL.sum FL.drain)
+        , benchIOSink
+            value
+            "fold-scanMany"
+            (S.fold $ FL.scanMany (FL.take 2 FL.toList) FL.drain)
```
harendra-kumar, line 351: Change `toList` to `drain` for fair comparison with scan and postscan.

harendra-kumar:
Even with Fold.drain it is taking long, maybe because we are dividing the scan into too many parts:

```
All.Data.Fold/o-1-space.transformation.fold-scanMany:   +RTS -T -K36K -M16M    -RTS      
All
  Data.Fold/o-1-space
    transformation
      fold-scanMany: OK (0.15s)
        1.22 ms ±  87 μs, 8.0 MB allocated, 1.0 KB copied, 2.0 MB peak memory
```

Issue #1308, opened 2021-10-30
A scanMany combinator


harendra-kumar:
Like foldMany. The current scan combinator stops if the scanning fold terminates. The proposed `scanMany` would start fresh after terminating.

This combinator can be added to streams/folds/unfolds.

rnjtranjan:
What will be signatures for "scanMany" for Fold and UnFold?

harendra-kumar:
It will be same as scan. The only difference between scan and scanMany is that scanMany continues to scan again after it terminates.

harendra-kumar:
It may be useful to pass a state from one scan instance to the next. For example, when splitting a stream into slices we may have a global state running through all splits and a local state (length of the split) running through a single instance. This can possibly be implemented by using a `Refold` in `scanMany`.

harendra-kumar:
Done.

PR #1537, opened 2022-03-17
add scanMnay for Unfold

rnjtranjan:


rnjtranjan:
Fixes: #1308

core/src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -458,6 +462,47 @@ scan (Fold stepF initial extract) (Unfold stepU injectU) =
...
+                    FL.Done v -> do
+                        if isNothing nxt
+                        then return $ Yield v (Just (fs, s), Just v)
+                        else do
+                            r <- initial
+                            return $ case r of
+                                FL.Partial fs1 ->
+                                    Skip (Just (fs1, us), Nothing)
+                                FL.Done _ ->
+                                    Skip (Nothing, Nothing)
```
harendra-kumar, line 496: This is unnecessarily complicated, why not just say return `Stop` here? The `Maybe` is not required on the first element of the tuple.

harendra-kumar: Better would be to just change `scan` to `scanWith` and pass a flag to it to stop or restart the scan. Then use the same common function of scan and scanMany.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -475,6 +479,7 @@ either (Unfold step1 inject1) = Unfold step inject
             Skip s -> Skip (s, f)
             Stop -> Stop) <$> step1 st
 
+
```
harendra-kumar, line 482: Remove blank line.

core/src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -507,6 +512,64 @@ postscan (Fold stepF initial extract) (Unfold stepU injectU) =
...
+    where
+
+    inject a =  do
+        r <- initial
+        case r of
+            FL.Partial fs -> do
+                x <- injectU a
+                return (Just (fs, x), Nothing)
+            FL.Done _ -> return (Nothing, Nothing)
```
harendra-kumar, line 528: This is not correct, we need to yield the result before we stop.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -507,6 +511,66 @@ postscan (Fold stepF initial extract) (Unfold stepU injectU) =
...
+    Unfold step inject
+
+    where
+
+    inject a =  do
+        r <- initial
+        case r of
+            FL.Partial fs -> do
+                x <- injectU a
+                return (Just (fs, x), Nothing)
```
harendra-kumar, line 526: A scan always yields the initial value. This is not correct. Can you write a test for this?

harendra-kumar (changes requested):

PR #1539, opened 2022-03-17
Add ScanMany for Stream

rnjtranjan:


rnjtranjan:
Fixes: #1308

src/Streamly/Internal/Data/Stream/IsStream/Transform.hs
```diff
@@ -678,6 +679,11 @@ scan :: (IsStream t, Monad m) => Fold m a b -> t m a -> t m b
 -- scan = P.scanOnce
 scan fld m = fromStreamD $ D.scanOnce fld $ toStreamD m
 
+{-# INLINE scanMany #-}
+scanMany :: (IsStream t, Monad m) => Fold m a b -> t m a -> t m b
+-- scan = P.scanOnce
```
harendra-kumar, line 684: Remove this comment.

core/src/Streamly/Internal/Data/Stream/StreamD/Transform.hs
```diff
@@ -442,35 +443,58 @@ postscanOnce (FL.Fold fstep initial extract) (Stream sstep state) =
...
                         !b <- extract fs1
-                        return $ Yield b $ ScanDo s fs1
-                    FL.Done b -> return $ Yield b ScanDone
-            Skip s -> return $ Skip $ ScanDo s fs
+                        return $ Yield b (ScanDo s fs1, Nothing)
+                    FL.Done b ->
+                        if restart
+                        then do
+                            if isNothing nxt
+                            then return $ Yield b (ScanDo s fs, Just b)
```
harendra-kumar, line 475: I do not understand the point of keeping the second element of state tuple. This is unnecessarily complicated. All we need to do is unconditionally reinit the scan on done if the `restart` flag is true.

harendra-kumar (changes requested):

Issue #1538, opened 2022-03-17
Add a test for Fold.postscan


harendra-kumar:
Looking at this code:

```
    initial = do
        r <- initialR
        rL <- initialL
        case r of
            Partial sR ->
                case rL of
                    Done _ -> Done <$> extractR sR
                    Partial sL -> return $ Partial (sL, sR)
```

`Done _ -> Done <$> extractR sR`, is discarding the result and just extracting right fold without using the left result.

harendra-kumar:
Also, fix the bug described above.

harendra-kumar:
Ah, I forgot that this is postscan, it is not supposed to use the initial value.

PR #1540, opened 2022-03-17
Add a test for Fold.postscan

rnjtranjan:


rnjtranjan:
Fixes: #1538

test/Streamly/Test/Data/Fold.hs
```diff
@@ -519,6 +519,27 @@ unzip =
...
 
+
+postscan :: Property
+postscan =
+    monadicIO $ do
+    v1 :: [Int] <-
+        run
+            $ S.fold (F.postscan FL.sum FL.toList)
+            $ S.fromList [1,2,3,4]
+    let v2 = [1,3,6,10]
```
harendra-kumar, line 530: Can we generate a random list and compare against the list Data.List scan function, dropping the first element?

PR #1542, opened 2022-03-17
Add test case for Fold.scan

rnjtranjan:


rnjtranjan:
Benchmark for Fold.scan is already in master.

test/Streamly/Test/Data/Fold.hs
```diff
@@ -666,6 +666,26 @@ splitAt =
...
     ("Hello ","World!")
 
+scan :: Property
+scan =
+    monadicIO $ do
+    v1 :: [Int] <-
+        run
+            $ S.fold (F.scan FL.sum FL.toList)
+            $ S.fromList [1,2,3,4]
+    let v2 = [0,1,3,6,10]
```
harendra-kumar, line 676: Can generate a random list and compare against the Data.List `scan`.

harendra-kumar (changes requested):

Issue #927, opened 2021-02-09
Drop support for ghc-8.0 and 8.2


harendra-kumar:
1. custom setup requires compiling cabal for these versions
2. 8.2 has known memory hogging issues
3. We need code at several places which requires custom handling for these e.g. containers library use in Data.Fold

etc.

PR #1543, opened 2022-03-17
Drop 8.0 to 8.2.0 GHC Support

rnjtranjan:


rnjtranjan:
Fixes: #927

rnjtranjan:
GHC 8.2.2 build fails,  need help to fix it.

core/src/Streamly/Internal/BaseCompat.hs
```diff
@@ -19,10 +19,6 @@ module Streamly.Internal.BaseCompat
 where
 
 import Data.Coerce (Coercible, coerce)
-#if MIN_VERSION_base(4,10,0)
-import Data.Either (fromRight, fromLeft)
-import qualified GHC.Exts as GHCExt
```
harendra-kumar, line 24: These lines should be kept unconditionally.

core/src/Streamly/Internal/BaseCompat.hs
```diff
@@ -40,16 +36,13 @@ import Foreign.ForeignPtr (withForeignPtr)
...
-
-#if !(MIN_VERSION_base(4,10,0))
 fromLeft :: a -> Either a b -> a
 fromLeft _ (Left a) = a
 fromLeft a _        = a
 
 fromRight :: b -> Either a b -> b
 fromRight _ (Right b) = b
 fromRight b _         = b
-#endif
```
harendra-kumar, line 52: These should be removed.

core/src/Streamly/Internal/BaseCompat.hs
```diff
@@ -40,16 +36,13 @@ import Foreign.ForeignPtr (withForeignPtr)
 errorWithoutStackTrace :: String -> a
 errorWithoutStackTrace = error
 #endif
```
harendra-kumar, line 38: This should be removed.

harendra-kumar (changes requested):
This looks mostly incorrect. Looks like the code that was valid for newer compilers has been removed. Please review.

Also, the CIs for 8.0.x and 8.2.x should be removed.

harendra-kumar (changes requested):
Still looks mostly incorrect changes.

Consider code like this:

```
#if MIN_VERSION_base(4,10,0)
code
#endif
```

This means if base is at least 4.10.0 . After removing GHC 8.2 and lower we always have base at least 4.10. So the above code should be kept unconditionally instead of removing it.

benchmark/lib/Streamly/Benchmark/Prelude.hs
```diff
@@ -68,6 +68,9 @@ import Control.Applicative (liftA2)
 import Control.DeepSeq (NFData(..))
 import Control.Exception (try)
 import Data.Functor.Identity (Identity)
+#if !(MIN_VERSION_base(4,11,0))
+import Data.Semigroup (Semigroup((<>)))
+#endif
```
harendra-kumar, line 73: This has to be removed. This means GHC versions <= 8.2.x which we are not supporting any more.

core/src/Streamly/Internal/BaseCompat.hs
```diff
@@ -19,7 +19,10 @@ module Streamly.Internal.BaseCompat
 where
 
 import Data.Coerce (Coercible, coerce)
+#if MIN_VERSION_base(4,10,0)
 import Data.Either (fromRight, fromLeft)
+import qualified GHC.Exts as GHCExt
+#endif
```
harendra-kumar, line 25: The `#if` condition has to be removed, here and at other places as well. base is always >= 4.10 now.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/SmallArray/Type.hs
```diff
@@ -752,16 +706,6 @@ listLiftShowsPrec _ sl _ = sl
...
-#if MIN_VERSION_base(4,9,0) || MIN_VERSION_transformers(0,4,0)
--- | @since 0.6.4.0
-instance Show1 SmallArray where
-#if MIN_VERSION_base(4,9,0) || MIN_VERSION_transformers(0,5,0)
-  liftShowsPrec = smallArrayLiftShowsPrec
-#else
-  showsPrec1 = smallArrayLiftShowsPrec showsPrec showList
-#endif
-#endif
-
```
harendra-kumar, line 764: Should be kept.

test/Streamly/Test/Data/Parser.hs
```diff
@@ -92,7 +92,7 @@ dieM =
 parserFail :: Property
 parserFail =
     property $
-    case S.parse (fail err) (S.fromList [0 :: Int]) of
+    case S.parse (P.die err) (S.fromList [0 :: Int]) of
```
harendra-kumar, line 95: should be `fail`.

core/src/Streamly/Internal/Data/Parser/ParserK/Type.hs
```diff
@@ -238,13 +230,6 @@ instance Monad m => Monad (Parser m a) where
     {-# INLINE fail #-}
     fail = die
 #endif
-
-#if MIN_VERSION_base(4,9,0)
-instance Monad m => Fail.MonadFail (Parser m a) where
-    {-# INLINE fail #-}
-    fail = die
-#endif
-
```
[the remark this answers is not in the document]
rnjtranjan: MIN_VERSION_base will  be always >= (4,10,0)
then it will never execute?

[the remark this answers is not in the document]
harendra-kumar: It seems you still have an incorrect understanding of the macro. `#if MIN_VERSION_base` means - this code block is enabled as long as the `base` version is 4.9.0 or more. So if base version is 4.10, 4.11, 4.12... this code will be enabled.

PR #1545, opened 2022-03-20
Fix Top.outer join

rnjtranjan:


rnjtranjan:
Fixes: #1545

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -54,7 +54,8 @@ import Control.Monad.Catch (MonadCatch)
 import Control.Monad.IO.Class (MonadIO(..))
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.State.Strict (get, put)
--- import Data.Hashable (Hashable)
+import Data.Function ((&))
+import Data.Maybe (isJust)
```
harendra-kumar, line 58: Sort the imports.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -80,6 +81,7 @@ import qualified Streamly.Internal.Data.Stream.IsStream.Reduce as Stream
 import qualified Streamly.Internal.Data.Stream.IsStream.Transform as Stream
 import qualified Streamly.Internal.Data.Stream.IsStream.Type as IsStream
 import qualified Streamly.Internal.Data.Stream.StreamD as StreamD
+import qualified Streamly.Internal.Data.Array.Foreign.Mut.Type as MA
```
harendra-kumar, line 84: sort the imports.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -444,16 +446,28 @@ joinOuter :: MonadIO m =>
 joinOuter eq s1 s =
     Stream.concatM $ do
         arr <- Array.fromStream $ fmap (,False) s
-        return $ go arr <> leftOver arr
+        let len = length arr
+        arr2 <-
+            Stream.fold
+            (MA.writeN len)
+            (Stream.fromList (Prelude.replicate len False))
+        return $ go arr arr2  <> leftOver arr arr2
```
harendra-kumar, line 454: This looks unnecessarily complicated.

Why not just use the existing code with following modifications:

* Make the existing Array a mutable array
* Uncomment line 476 in the old code, and use `modifyIndex` to modify the flag in the tuple to True.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -443,17 +445,29 @@ joinOuter :: MonadIO m =>
...
 
-    leftOver =
-            fmap (\(x, _) -> (Nothing, Just x))
-                . Stream.filter (not . snd)
-                . Array.toStream
-
-    go arr = Stream.evalStateT (return False) $ do
+    leftOver arr arr2 =
+            let stream1 = IsStream.fromSerial $ Array.toStream arr
+                stream2 = IsStream.fromSerial $ fromStreamD (MA.toStreamD arr2)
```
harendra-kumar, line 460: Why not use `MA.toStream` instead?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
rnjtranjan: I don't find toStream API :
Not in scope: ‘MA.toStream’
Perhaps you meant one of these:
  ‘MA.toStreamD’ (imported from Streamly.Internal.Data.Array.Foreign.Mut.Type),
  ‘MA.toStreamK’ (imported from Streamly.Internal.Data.Array.Foreign.Mut.Type)
Module ‘Streamly.Internal.Data.Array.Foreign.Mut.Type’ does not export ‘toStream’.typecheck

harendra-kumar: If there is no `toStream` you can use an unfold, like `unfold MArray.read`. If we have to use `StreamD` to convert an array to stream then there is a major issue with the array API and it has to be fixed.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -82,6 +83,7 @@ import qualified Streamly.Internal.Data.Stream.IsStream.Type as IsStream
 import qualified Streamly.Internal.Data.Stream.StreamD as StreamD
 
 import Prelude hiding (filter, zipWith, concatMap, concat)
+import qualified Streamly.Internal.Data.Array as Array
```
harendra-kumar, line 86: Why is this import moved here? It can remain where it was.

src/Streamly/Internal/Data/Stream/IsStream/Top.hs
```diff
@@ -443,17 +445,29 @@ joinOuter :: MonadIO m =>
...
     Stream.concatM $ do
-        arr <- Array.fromStream $ fmap (,False) s
-        return $ go arr <> leftOver arr
+        arr <- Array.fromStream s
+        let len = length arr
+        arr2 <-
+            Stream.fold
+            (MA.writeN len)
+            (Stream.fromList (Prelude.replicate len False))
+        return $ go arr arr2 <> leftOver arr arr2
```
harendra-kumar, line 454: We can give better names to `arr` and `arr2` e.g. `inputArr`, `foundArr`.

harendra-kumar (commented):
Please squash the commits.

Issue #1182, opened 2021-08-05
Change the argument order of Unfold.many


harendra-kumar:
It is like this:

```
many :: Monad m => Unfold m a b -> Unfold m b c -> Unfold m a c
```

The second argument is the one that is applied many times. To read it more naturally we need that to be the first argument. Compare with `Fold.many`, it applies the first argument repeatedly.

harendra-kumar:
This will also be consistent with `unfoldMany`.

PR #1547, opened 2022-03-21
Reorder gbracket arguments

rnjtranjan:


rnjtranjan:
fixes: #1182

core/src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -272,9 +272,13 @@ bracket' :: (MonadAsync m, MonadCatch m) =>
     -> (b -> m e)
     -> (b -> Stream m a)
     -> Stream m a
-bracket' bef aft exc gc =
-    gbracket bef (inline MC.try) aft gc
+bracket' bef aft gc exc =
```
harendra-kumar, line 275: Why did the order change in this? This was not proposed. This is dangerous because this will silently introduce bugs in all current users of `bracket'` . If we change the order here we should fix all the callers of this. Because of this, IsStream.bracket' behaviour would become incorrect, streamly-process package may also be broken by this.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
harendra-kumar: Let's not change the order in this API.

Issue #1215, opened 2021-08-28
Add some array APIs


harendra-kumar:
```
-- | Truncate the array at the beginning and end as long as the predicate
-- holds true.
strip :: (a -> Bool) -> Array a -> Array a
--
-- | Update multiple indices of an array.
overlay :: Array a -> SerialT m (Int, a) -> m (Array a)
```

harendra-kumar:
In the `Unimplemented` APIs I have added `modifyIndices` which is a replacement of overlay:

```
-- | Modify the array indices generated by the supplied unfold.
--
-- /Pre-release/
modifyIndices :: -- forall m a b. (MonadIO m, Storable a) =>
    Unfold m (Array a) Int -> Array a -> (a -> a) -> m ()
modifyIndices = undefined
```

See `getIndices` in mutable foreign array module for implementation of a similar API.

harendra-kumar:
On second thoughts, I am wondering if an Unfold is necessary in this case as a Stream would be as efficient and more flexible because it allows both Stream and unfolds (after converting to stream).

Also, we can keep the `overlay` API as well along with modifyIndices for convenience.

harendra-kumar:
We can write some benchmarks to check if Stream would be as good as Unfold.

PR #1548, opened 2022-03-21
Add benchmark for MArray.modifyIndices

rnjtranjan:


rnjtranjan:
Fixes: #1215

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -63,6 +64,8 @@ o_1_space_serial_marray value array =
         $ MArray.strip (\x -> x < value `div` 2 || x > value `div` 2)
     , benchIO "strip (> 0)" (const array)
         $ MArray.strip (> 0)
+    , benchIO "modifyIndices id" (const array)
+        $ Stream.fold (MArray.modifyIndices id array) . (IS.fromStreamD . MArray.toStreamD)
```
harendra-kumar, line 68: Can we use `+1` instead of `id`, to ensure some change is being made to the array.

Is this line going beyond 80 columns?

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -25,6 +25,7 @@ import System.Random (randomRIO)
 
 import qualified Streamly.Prelude as Stream
 import qualified Streamly.Internal.Data.Array.Foreign.Mut as MArray
+import qualified Streamly.Internal.Data.Stream.IsStream as IS
```
harendra-kumar, line 28: We can import this as `Stream` itself.

harendra-kumar (changes requested):

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -52,21 +53,25 @@ benchIO name src sink =
...
+        , benchIO "partitionBy (> value/2)" (const array)
+            $ MArray.partitionBy (> (value `div` 2))
+        , benchIO "strip (< value/2 || > value/2)" (const array)
+            $ MArray.strip (\x -> x < value `div` 2 || x > value `div` 2)
+        , benchIO "strip (> 0)" (const array)
+            $ MArray.strip (> 0)
+        , benchIO "modifyIndices (+ 2)" (const idxarr)
+            $ Stream.fold
+            (MArray.modifyIndices (+ 2) array)
+            . (Stream.fromStreamD . MArray.toStreamD)
```
harendra-kumar, line 73: We should not use StreamD functions. StreamD is a purely internal module. Isn't there a `toStream` function in MArray?

harendra-kumar (changes requested):

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -52,21 +53,25 @@ benchIO name src sink =
...
+            $ MArray.partitionBy (> 0)
+        , benchIO "partitionBy (< value/2)" (const array)
+            $ MArray.partitionBy (< (value `div` 2))
+        , benchIO "partitionBy (> value/2)" (const array)
+            $ MArray.partitionBy (> (value `div` 2))
+        , benchIO "strip (< value/2 || > value/2)" (const array)
+            $ MArray.strip (\x -> x < value `div` 2 || x > value `div` 2)
+        , benchIO "strip (> 0)" (const array)
+            $ MArray.strip (> 0)
+        , benchIO "modifyIndices (+ 2)" (const idxarr)
```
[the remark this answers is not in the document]
rnjtranjan: earlier code was having problem after the first pass all the index get incremented so we are geting IndexOutofBound error.

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -78,7 +82,11 @@ main = do
 
     where
 
-    alloc value = MArray.fromStream (sourceUnfoldrM value 0 :: SerialT IO Int)
+    alloc value =
+        sequence
+        [ MArray.fromStream (sourceUnfoldrM value 0 :: SerialT IO Int)
+        , MArray.fromStream (sourceUnfoldrM value 0 :: SerialT IO Int)
+        ]
```
harendra-kumar, line 89: 1. A list is not needed, a tuple should be enough.
2. A mutable array is not needed for indices, an immutable array should be enough.

benchmark/Streamly/Benchmark/Data/Array/Foreign/Mut.hs
```diff
@@ -38,7 +39,7 @@ sourceUnfoldrM :: (IsStream t, MonadAsync m) => Int -> Int -> t m Int
 sourceUnfoldrM value n = Stream.unfoldrM step n
     where
     step cnt =
-        if cnt > n + value
+        if cnt > n + (value - 1)
```
rnjtranjan, line 42: why do we need -1 here?

harendra-kumar: You added that, so you need to explain :-)

rnjtranjan: ahh I thought initially array index generated by unfold has 0 to n causing indexoutofBound.
But it generates index 0 to n-1 so -1 is not needed.

Issue #1565, opened 2022-03-28
slidingWindow: use error when windowsize n <= 0


harendra-kumar:
Otherwise `step` will crash.

harendra-kumar:
fixed by #1569

PR #1569, opened 2022-03-29
Use error when windowsize n <= 0

rnjtranjan:


rnjtranjan:
fixes : #1565

harendra-kumar (changes requested):
squash the commits

PR #1570, opened 2022-03-29
Add bubbleAsc function

rnjtranjan:


core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,26 @@ strip eq arr@Array{..} = do
...
+-- | Bubble sorts the elements.
+--
+-- /Pre-release/
+{-# INLINE bubbleAsc #-}
+bubbleAsc :: (MonadIO m, Storable a) =>
+    Int -> (a -> a -> Ordering) -> Array a -> m ()
+bubbleAsc idx cmp0 arr =
+    do
+    let l = length arr
+    go idx l
```
harendra-kumar, line 2348: Why `l` is being passed to `go` if it never changes? It can be defined as a `where` binding.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,26 @@ strip eq arr@Array{..} = do
...
             else return cur
         else return cur
+
+-- | Bubble sorts the elements.
+--
+-- /Pre-release/
+{-# INLINE bubbleAsc #-}
+bubbleAsc :: (MonadIO m, Storable a) =>
+    Int -> (a -> a -> Ordering) -> Array a -> m ()
+bubbleAsc idx cmp0 arr =
```
harendra-kumar, line 2345: We will usually be appending elements to the array. In this API we can assume that the out-of-order element is the last element of the array. So we need not pass idx.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,26 @@ strip eq arr@Array{..} = do
...
+    do
+    let l = length arr
+    go idx l
+
+    where
+    go i l =
+        when ((i + 1) < l) $ do
```
harendra-kumar, line 2352: Assuming grow by appends as I assumed in an earlier comment, we should keep going back to the first index.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,26 @@ strip eq arr@Array{..} = do
...
+        x1 <- getIndexUnsafe i arr
```
harendra-kumar, line 2353: indent this against `when` by 4 spaces.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,26 @@ strip eq arr@Array{..} = do
...
+
+    go i =
+        when (i >= 0) $ do
+            x1 <- getIndexUnsafe i arr
+            x2 <- getIndexUnsafe (i+1) arr
+            case x1 `cmp0` x2 of
+                GT -> do
+                    _ <- putIndexUnsafe i x2 arr
+                    _ <- putIndexUnsafe (i+1) x1 arr
+                    go (i - 1)
```
harendra-kumar, line 2359: We can pass `x2` to `go` so that we do not have to read it again in the next iteration of the loop.

harendra-kumar (changes requested):
Also, squash the commits.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,27 @@ strip eq arr@Array{..} = do
...
         else return cur
+
+-- | Bubble sorts the elements.
+--
+-- /Pre-release/
+{-# INLINE bubbleAsc #-}
+bubbleAsc :: (MonadIO m, Storable a) =>
+    (a -> a -> Ordering) -> Array a -> m ()
+bubbleAsc cmp0 arr = do
+    x <- getIndexUnsafe (l - 1) arr
```
harendra-kumar, line 2346: This would crash on empty array.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,27 @@ strip eq arr@Array{..} = do
...
+    where
+
+    l = length arr
+
+    go x i =
+        when (i >= 0) $ do
+            x1 <- getIndexUnsafe i arr
+            case x1 `cmp0` x of
+                GT -> do
+                    _ <- putIndexUnsafe i x arr
```
harendra-kumar, line 2358: We can avoid writing this, and write it only in the end when we exit the loop on line 2361.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,27 @@ strip eq arr@Array{..} = do
...
         else return cur
+
+-- | Bubble sorts the elements.
+--
+-- /Pre-release/
+{-# INLINE bubbleAsc #-}
+bubbleAsc :: (MonadIO m, Storable a) =>
+    (a -> a -> Ordering) -> Array a -> m ()
+bubbleAsc cmp0 arr =
+    when (l > 0) $ do
```
harendra-kumar, line 2346: We can make this `l > 1`

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -2334,3 +2335,27 @@ strip eq arr@Array{..} = do
...
+        where
+
+        l = length arr
+
+        go x i =
+            when (i >= 0) $ do
+                x1 <- getIndexUnsafe i arr
+                case x1 `cmp0` x of
+                    GT -> do
+                        _ <- putIndexUnsafe (i + 1) x1 arr
```
harendra-kumar, line 2359: Why do we need `_ <- ...`?

harendra-kumar (changes requested):
We can write a simple test for this. Just use a quickcheck property test to generate a random list, append the list elements one by one to an array and call bubbleAsc each time. At the end compare the array with the sorted list.

PR #1571, opened 2022-03-29
Categorize docs in a more elaborate way

adithyaov:


harendra-kumar (commented):
* Rename "Explanation" to "Explanatory"
* performance.md should be in explanatory docs
* resources.md should be at top level perhaps or in explanatory

harendra-kumar (approved):

PR #1572, opened 2022-03-29
Complete doc reorder

adithyaov:


harendra-kumar (commented):
Why do we need these in the docs directory? Are the dev docs going to displayed on the same page along side user docs?

adithyaov:
> Why do we need these in the docs directory? Are the dev docs going to displayed on the same page along side user docs?

As they are structured, yes. I can however add another node called `Guides`.

adithyaov:
There is some CI glitch. Look here for the Github CI https://github.com/composewell/streamly/actions/runs/2141709667

PR #1574, opened 2022-03-30
Rename Socket BufferOf APIs

rnjtranjan:


src/Streamly/Network/Socket.hs
```diff
@@ -101,16 +101,16 @@ module Streamly.Network.Socket
...
+    , readChunksWith
     , readChunk
 
     -- * Write
     , write
-    , writeWithBufferOf
+    , writeWith
     , writeChunks
-    , writeChunksWithBufferOf
+    , writeChunksWith
```
harendra-kumar, line 113: We need to deprecate the old APIs instead of removing them altogether.

harendra-kumar (changes requested):

src/Streamly/Network/Socket.hs
harendra-kumar: We need to address this. Also, add a changelog entry about renaming.

harendra-kumar:
I do not see any changes since last review.

src/Streamly/Internal/Network/Socket.hs
```diff
@@ -486,32 +517,40 @@ writeChunksWithBufferOf n h = lpackArraysChunksOf n (writeChunks h)
...
-putBytesWithBufferOf :: MonadIO m => Int -> Socket -> SerialT m Word8 -> m ()
-putBytesWithBufferOf n h m = putChunks h $ AS.arraysOf n m
+{-# INLINE putBytesWith #-}
+putBytesWith :: MonadIO m => Int -> Socket -> SerialT m Word8 -> m ()
+putBytesWith n h m = putChunks h $ AS.arraysOf n m
 
 -- | Write a byte stream to a socket. Accumulates the input in chunks of
 -- specified number of bytes before writing.
 --
 -- @since 0.7.0
```
harendra-kumar, line 527: This should be since 0.9.0, the same applies to other deprecated functions as well.

src/Streamly/Internal/Network/Socket.hs
```diff
@@ -486,32 +517,40 @@ writeChunksWithBufferOf n h = lpackArraysChunksOf n (writeChunks h)
...
+{-# INLINE writeWith #-}
+writeWith :: MonadIO m => Int -> Socket -> Fold m Word8 ()
+writeWith n h = FL.chunksOf n (A.writeNUnsafe n) (writeChunks h)
+
+-- | Same as 'writeWith'
+--
+-- @since 0.9.0
```
harendra-kumar, line 534: This should be since 0.7.0, the same applies to other deprecated functions as well.

harendra-kumar (changes requested):

PR #1578, opened 2022-04-05
Seperate benchmark core package

rnjtranjan:


harendra-kumar (changes requested):
* There is too much duplication of code between streamly-core-benchmarks and streamly-benchmarks. If some modules are common then stream-benchmarks can depend on streamly-core-benchmarks and use the modules from it. Maintaining exact copies of the same code in two packages would be a nightmare.
* bench-report code can be used from the bench-report package that adithya is creating, instead of copying it, we can wait for adithya to commit the new changes.
* Files should be moved using `git mv` instead of copying, we need to maintain the history.

harendra-kumar:
Closing in favor of #1593 .

PR #1583, opened 2022-04-13
Implement slidingWindowWith in Streamly.Internal.Data.Ring.Foreign module

rnjtranjan:


core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -351,3 +360,82 @@ mean = Fold.teeWith (/) sum length
...
+        return
+            $ Partial (rb0, rh0, 0)
+
+    step (_, _, i) ((_, _), (rb, rh)) =
+        return $ Partial (rb, rh, i+1)
+
+    extract = return
+
+md ::  MonadIO m => Int -> Fold m Double Double
+md n = slidingWindowWith n $ Fold.teeWith getMD meanWith foldRingBuff
```
harendra-kumar, line 431: This is not good.

As I explained earlier we need to pass (a, Maybe a) to the `mean` fold and (Ring a, Ptr a) to the `last` fold and then zip the results. Use `lmap fst` and `lmap snd` to pass only selected input to `mean` and `last`. We do not need meanWith  or foldRingBuff, implementing such custom functions goes against all the modularity and makes it ugly.

harendra-kumar: `md` should be implemented in the streamly-statistics package.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold/Window.hs
rnjtranjan: how do we get the length of the ring buffer from last fold?
we need it to decide between using unsafeFoldRingM and unsafeFoldRingMFull.

core/src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -559,3 +560,42 @@ slidingWindow n (Fold step1 initial1 extract1) = Fold step initial extract
                     Done b -> Done b
 
     extract (Tuple4' _ _ _ st) = extract1 st
+
+{-# INLINE slidingWindowWith #-}
+slidingWindowWith :: forall m a b. (MonadIO m, Storable a)
+    => Int -> Fold m ((a, Maybe a), (Ring a, Ptr a, Int)) b -> Fold m a b
+slidingWindowWith n (Fold step1 initial1 extract1) = Fold step initial extract
```
harendra-kumar, line 567: Can we define slidingWindow as follows?

```
slidingWindow n f = slidingWindowWith n (Fold.lmap fst f)
```

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Ring/Foreign.hs
```diff
@@ -446,6 +449,24 @@ unsafeFoldRing ptr f z Ring{..} =
...
+unsafeFoldRingFull rh f z rb@Ring {..} =
+    unsafeInlineIO $ withForeignPtrM ringStart $ \_ -> go z rh
+  where
+    go !acc !start = do
+        let !x = unsafeInlineIO $ peek start
+            acc' = f acc x
+            ptr = advance rb start
+        if ptr == rh
+            then return acc'
+            else go acc' ptr
```
harendra-kumar, line 468: We can implement this using unsafeFoldRingM.

harendra-kumar (changes requested):

Issue #1562, opened 2022-03-26
takeBetween parser test failure


harendra-kumar:
https://github.com/composewell/streamly/runs/5705388063?check_suite_focus=true

```
Failures:

  Streamly/Test/Data/Parser/ParserD.hs:742:9: 
  1) Data.Parser.ParserD, test for sequence parser, P.takeBetween m n = Prelude.take when len >= m and len <= n and                ail otherwise
       Falsified (after 68 tests):
         4
         19
         [2340,5816,4025,2150,2399,8755,5383,5923,330,5297,1674,696,4118,4738,7152,7603,7129,3027,9308,9594,4330,964,8392]

  To rerun use: --match "/Data.Parser.ParserD/test for sequence parser/P.takeBetween m n = Prelude.take when len >= m and len <= n and                \fail otherwise/"
```

rnjtranjan:
fixed in PR : #1585

PR #1585, opened 2022-04-19
Fix test case failure in CI

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1562

PR #1590, opened 2022-04-21
Fix default.nix for VS code

rnjtranjan:


default.nix
```diff
@@ -129,7 +129,6 @@ let haskellPackages =
         # Use a better prompt
         shellHook = ''
           export CABAL_DIR="$(pwd)/.cabal.nix"
-          cabal user-config update -a "jobs: 1"
```
harendra-kumar, line 132: Can we make it `"jobs: 2"` instead?

rnjtranjan: No, It doesn't work with 2.

harendra-kumar: That's strange. Then comment out this line instead of removing it. And add a comment above it saying that this is commented for hls to work with VSCode so that we do not re-introduce this in future.

PR #1597, opened 2022-05-05
Export isPower2

rnjtranjan:


rnjtranjan:
this function is not required to be exported.

PR #1606, opened 2022-05-09
Refactor parser helper functions type signatures for better currying

### LINKED ISSUE #1594: Consider changing signatures of some parsers that use a fold for collecting output
For example:

```
many :: MonadCatch m => Parser m a b -> Fold m b c -> Parser m a c
```

Can be changed to:

```
many :: MonadCatch m => Fold m b c -> Parser m a b -> Parser m a c
```

That the fold can be used to curry the function to get a signature that is more like the standard `many` function.

twitu:
Closes #1594


twitu:
Closes #1594

twitu:
There are two key arguments for and against this change.
1. Currently the convention in different parts of the code base is to have `Parser m a b -> Fold m b c -> Parser m a c`. This style makes the flow of types intuitive as the type flows from a to b and then to c.
2. The proposed change makes `Fold m b c -> Parser m a b -> Parser m a c` currying easier. After currying the function looks like a transformation on the parser as in `Parser m a b -> Parser m a c`.

However there is strong evidence in favor of either, so the conclusion was to keep the current API until a better reason for the change comes up.

PR #1621, opened 2022-05-19
Replace SerialT to Stream

rnjtranjan:


rnjtranjan:
a new PR is created: https://github.com/composewell/streamly/pull/1642

PR #1623, opened 2022-05-19
Refactor some parser APIs

harendra-kumar:


core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -409,12 +412,17 @@ eof = D.toParserK D.eof
...
+-- >>> toMaybe f x = if f x then Just x else Nothing
+-- >>> satisfy f = Parser.maybe (toMaybe f)
+--
 -- /Pre-release/
 --
 {-# INLINE satisfy #-}
 satisfy :: MonadCatch m => (a -> Bool) -> Parser m a a
 satisfy = D.toParserK . D.satisfy
 
+-- XXX Change this to "next".
```
adithyaov, line 424: What?

adithyaov: To `next` or to `one`?

adithyaov: `Parser.fromFold one` would have `Parser m a (Maybe a)`
Should we have a combinator that converts failable folds to failure in parsers?
`fromFoldMaybe :: Fold m a (Maybe b) -> Parser m a b`?
Can you see any other use case for this?

adithyaov (approved):

PR #1633, opened 2022-05-21
Implement `sequence` for ParserD

twitu:


core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+    where
+    initial = do
+        fres <- finitial
+        case fres of
+            FL.Partial fs -> return $ IPartial (Nothing', sstate, fs)
+            FL.Done c -> return $ IDone c
+
+    -- state does not contain any parser
+    -- yield a new parser from the stream
+    step (Nothing', ss, fs) _ = do
```
adithyaov, line 1917: You can probably create a custom state for being more elaborate,

```
data SequenceState = SSInit ss fs | SSParserInit ...
```

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+        sres <- sstep defState ss
+        case sres of
+            D.Yield p ss' -> return $ Continue 1 (Just' $ Left' p, ss', fs)
+            D.Stop -> do
+                c <- fextract fs
+                return $ Done 1 c
+            D.Skip ss' -> return $ Continue 1 (Nothing', ss', fs)
+
+    -- state contains a parser that has not yet been initialized
+    step (Just' (Left' (Parser pstep pinit pextr)), ss, fs) _ = do
```
adithyaov, line 1927: Why is `Left` required? This state is superseded by the one below, `Right`.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+        ps <- pinit
+        case ps of
+            IPartial ps' -> return $ Continue 1 (Just' $ Right' $ Parser pstep (return $ IPartial ps') pextr, ss, fs)
+            IDone b -> do
+                fres <- fstep fs b
+                case fres of
+                    FL.Partial fs' ->
+                        return $ Continue 1 (Nothing', ss, fs')
+                    FL.Done c -> return $ Done 1 c
+            IError err -> throwM $ ParseError err
```
adithyaov, line 1937: You can use the `Error` constructor here.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+
+    -- state contains a parser that has been initialized
+    -- and that needs to be stepped forward through the input
+    step (Just' (Right' (Parser pstep pinit pextr)), ss, fs) a = do
+        ps <- pinit
+        case ps of
+            IPartial ps' -> do
+                pres <- pstep ps' a
+                case pres of
+                    Partial n ps'' -> return $ Continue n (Just' $ Right' $ Parser pstep (return $ IPartial ps'') pextr, ss, fs)
```
adithyaov, line 1947: `Partial` instead of `Continue`?

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+                    Continue n ps'' -> return $ Continue n (Just' $ Right' $ Parser pstep (return $ IPartial ps'') pextr, ss, fs)
+                    Done n b -> do
+                        fres <- fstep fs b
+                        case fres of
+                            FL.Partial fs' ->
+                                return $ Continue n (Nothing', ss, fs')
```
adithyaov, line 1953: This would probably be `Partial`?

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+                            FL.Done c -> return $ Done n c
+                    Error msg -> throwM $ ParseError  msg
+            IDone b -> do
+                fres <- fstep fs b
+                case fres of
+                    FL.Partial fs' ->
+                        return $ Continue 1 (Nothing', ss, fs')
```
adithyaov, line 1960: `Partial`?

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+                    FL.Done c -> return $ Done 1 c
+            IError err -> throwM $ ParseError err
+
+    extract (_, _, fs) = fextract fs
```
adithyaov, line 1964: You should use any intermediate parser state while extracting.
1. extract the parser state
2. run the fold

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1901,14 +1901,68 @@ sepBy
...
+        case fres of
+            FL.Partial fs -> return $ IPartial (Nothing', sstate, fs)
+            FL.Done c -> return $ IDone c
+
+    -- state does not contain any parser
+    -- yield a new parser from the stream
+    step (Nothing', ss, fs) _ = do
+        sres <- sstep defState ss
+        case sres of
+            D.Yield p ss' -> return $ Continue 1 (Just' $ Left' p, ss', fs)
```
adithyaov, line 1920: Use `ss1`, `ps1`, etc. instead of `ss'`, `ps'` etc.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
twitu: Aah yes that is true. I somehow got stuck on logically separating the initial parser and the modified parser state. I think this whole portion will be overhauled, when using the separate state for the parser that you mentioned in a previous comment.

twitu: I've removed the extra Either constructor now, since Just and nothing is enough.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
twitu: So the general pattern I here is that when the inner Parser returns a Partial then the step should also be a Partial. This would flow directly into the extract step where, as you mentioned, I should extract the Parser state check if it is Partial, apply it to the fold and then return the result.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
twitu: After the current changes, do we still need it?

harendra-kumar:
I see you have merged `master` into your branch. Instead of that you should rebase your branch on top of master. See https://github.com/composewell/git-intro/blob/master/git-intro.rst for the git wrokflows.

twitu:
Removed the merge commit and used rebase. Ready for a re-review.

test/Streamly/Test/Data/Parser/ParserD.hs
```diff
@@ -635,6 +637,17 @@ parseUnfold = do
 
             listEquals (==) xs ls
 
+parserSequence :: Property
+parserSequence =
+  forAll (vectorOf 11 (listOf (chooseAny :: Gen Int))) $ \ins ->
+    monadicIO $ do
+      let parsers = (fromStreamK . SK.fromList) $ fmap (\xs -> P.fromFold $ FL.take (length xs) FL.sum) ins
```
harendra-kumar, line 644: You can use StreamD.fromList instead of using `fromStreamK . SK.fromList`.

benchmark/Streamly/Benchmark/Data/Parser/ParserD.hs
```diff
@@ -32,6 +32,7 @@ import qualified Streamly.Internal.Data.Parser.ParserD as PR
 import qualified Streamly.Internal.Data.Producer as Producer
 import qualified Streamly.Internal.Data.Producer.Source as Source
 import qualified Streamly.Internal.Data.Stream.IsStream as IP
+import qualified Streamly.Internal.Data.Stream.StreamD.Generate as SD
```
harendra-kumar, line 35: We usually import it `as D`. Also you can import `StreamD` itself (which is more stable wrt code changes) rather than importing the lower level `StreamD.Generate`.

harendra-kumar (commented):
1. Restrict the line lengths to less than 80 columns.
2. ParserD is internal only module we never use it in user code. Parser module is the one that is to be used in user code. For this code to be actually used we need to add a wrapper of `Parser` type in the Parser module.

twitu:
I've made the changes based on previous comments.

I want to check that I've used the right types.
1. Internal sequence operates on ParserD and StreamD
2. Wrapper sequence takes ParserK and (Foldable t) which gets converted to StreamK.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1932,8 +1933,14 @@ sequence (Fold fstep finitial fextract) (D.Stream sstep sstate) =
             IPartial ps1 -> do
                 pres <- pstep ps1 a
                 case pres of
-                    Partial n ps2 -> return $ Partial n (Just' $ Parser pstep (return $ IPartial ps2) pextr, ss, fs)
-                    Continue n ps2 -> return $ Continue n (Just' $ Parser pstep (return $ IPartial ps2) pextr, ss, fs)
+                    Partial n ps2 ->
+                        let newP = Just'
+                                $ Parser pstep (return $ IPartial ps2) pextr
```
harendra-kumar, line 1938: You can move `Just'` to the next line (Just' newP) so that you do not have to break this line.

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -1199,13 +1201,15 @@ takeP i p = D.toParserK $ D.takeP i $ D.fromParserK p
...
 -- This is same as 'Data.Traversable.sequence' but more efficient.
 --
--- /Unimplemented/
+-- /Pre-release/
 --
 {-# INLINE concatSequence #-}
 concatSequence ::
-    -- Foldable t =>
+    (MonadCatch m, Foldable t) =>
     Fold m b c -> t (Parser m a b) -> Parser m a c
```
harendra-kumar, line 1209: Instead of using a Foldable t, you can use `SerialT m (Parser m a b)` here, it would be more general as we can easily convert a Foldable to stream.

harendra-kumar (commented):
You can replicate the `ParserD` benchmark in the `Parser` benchmark module to check how this user facing API performs.

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -1200,7 +1200,9 @@ takeP i p = D.toParserK $ D.takeP i $ D.fromParserK p
 -- container @t@ using the fold @f@. Fails if the input ends or any of the
 -- parsers fail.
 --
--- This is same as 'Data.Traversable.sequence' but more efficient.
+-- This is same as 'Data.Traversable.sequence' but more efficient. An
+-- even more efficient implementation can use ParserD type Parser in
+-- the SerialT stream.
```
harendra-kumar, line 1205: I am not sure if it is more efficient as there is no fusion, I guess it would be the same. But it can use a stream of parsers rather than a traversable container, so is more general.

harendra-kumar:
Can you squash the commits appropriately, so that we have just one or a few commits of logically grouped changes?

core/src/Streamly/Internal/Data/Parser.hs
twitu: Removed the incorrect statement.

adithyaov (approved):
LGTM

adithyaov:
Can you fix the CI

PR #1634, opened 2022-05-22
Use Stream in place of SerialT

rnjtranjan:


rnjtranjan:
a new PR is created: https://github.com/composewell/streamly/pull/1642

Issue #1632, opened 2022-05-20
Implement fromListRevN for array


harendra-kumar:
It is left unimplemented in Streamly.Internal.Data.Array.Foreign.Type module. We need this in the array stream module for better prerformance of parseBreak.

PR #1635, opened 2022-05-22
Implement fromListRevN

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1632

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,24 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
...
+        arr <- newArray (max n 0)
+        return $ FL.Partial (arr, n - 1)
+
+    step (arr, i) x =
+        if i < 0
+        then
+            return $ FL.Done arr
+        else do
+            putIndex i x arr
+            return $ FL.Partial (arr, i - 1)
```
harendra-kumar, line 1868: Can you write this in the same way as `writeN` has been written? Using an array in the state would thread around an additional variable which might cause trouble with optimization. Also the tuple in the state is not strict.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,27 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
 writeN :: forall m a. (MonadIO m, Storable a) => Int -> Fold m a (Array a)
 writeN = writeNWith newArray
 
+-- /Pre-release/
+{-# INLINE writeRevN #-}
+writeRevN :: forall m a. (MonadIO m, Storable a) => Int -> Fold m a (Array a)
+writeRevN n = Fold step initial (return .fromArrayUnsafe . fst)
```
harendra-kumar, line 1855: Add space after "."

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,27 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
...
+    where
+
+    initial = do
+        arr <- newArray (max n 0)
+        return $ FL.Partial (toArrayUnsafe arr, n - 1)
```
harendra-kumar, line 1860: Do we need to store the count as we have start and end of the array? Having redundant/additional data in state causes performance issues. Also, we are not using a strict tuple here.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,27 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
...
+
+    step (arr@(ArrayUnsafe contents start end), i) x =
+        if i < 0
+        then
+            return $ FL.Done (fromArrayUnsafe arr)
```
harendra-kumar, line 1865: Bug: we are losing the element `x` here.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,27 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
...
+        else
+            do
```
harendra-kumar, line 1867: Collapse else and "do" on a single line.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1849,6 +1849,27 @@ writeNWith alloc n = FL.take n (writeNWithUnsafe alloc n)
...
+            liftIO $ poke (PTR_INDEX(start, i, a)) x >> touch contents
```
harendra-kumar, line 1868: Bring "touch" on the next line.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
rnjtranjan: this is the index where we poke the element in reverse order.

rnjtranjan (commented):
looks good.

Issue #1619, opened 2022-05-17
Add a parseContinue API


harendra-kumar:
See `foldOn` (to be renamed to foldContinue), this would be similar for parsing.

```
parseContinue :: Monad m => Parser m a b -> Stream m a -> Parser m a b
```

harendra-kumar:
It may not be possible to implement it correctly with this signature unless we discard the leftover buffer in the end. Because there is no way to return a leftover buffer in the end. Either we need to change the parser type to hold the buffer or we need to return the buffer.

PR #1637, opened 2022-05-23
Implement parseContinue

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1619

core/src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -108,6 +111,7 @@ import Streamly.Internal.Data.SVar.Type (State, adaptState, defState)
 import Streamly.Internal.Data.Unfold.Type (Unfold(..))
 
 import qualified Streamly.Internal.Data.Fold.Type as FL
+import qualified Streamly.Internal.Data.Parser.ParserD.Type as PD
```
harendra-kumar, line 114: importing `as ParserD` would be clearer, we also use `PRD` at other places, if a smaller name is required then that should be preferred to be consistent.

harendra-kumar (commented):
This is totally incorrect implementation. It is just ignoring the backtracking information. Similar to the `parse` operation we need to maintain a backtracking buffer and go back in the buffer whenever a backtrack count is returned.

Also, it looks to me that it may not be possible to implement it correctly with this signature unless we discard the leftover buffer in the end. Because there is no way to return a leftover buffer in the end.

rnjtranjan:
It may not be possible to implement it correctly with this signature unless we discard the leftover buffer in the end. Because there is no way to return a leftover buffer in the end. Either we need to change the parser type to hold the buffer or we need to return the buffer

Issue #1159, opened 2021-07-27
Add a cabal-diff CI to report API diff introduced by each PR


harendra-kumar:

harendra-kumar:
We need to run `cabal-diff streamly 0.8.0 . --no-color` to report the API changes. This would be just like the `cabal-docspec` CI.

PR #1640, opened 2022-05-24
Add a packdiff CI using the default executable

adithyaov:


adithyaov:
Closes #1159

cabal.project.packdiff
```diff
@@ -0,0 +1,23 @@
...
+
+source-repository-package
+  type: git
+  location: https://github.com/composewell/streamly-coreutils.git
+  tag: master
+
+source-repository-package
+  type: git
+  location: https://github.com/composewell/streamly-process.git
+  tag: master
```
harendra-kumar, line 23: This will be a very fragile test. Anytime a commit happens to these repos, streamly CIs will break. We should pin these packages to a commit.

harendra-kumar (changes requested):
Need to remove the test commit before merge.

cabal.project.packdiff
adithyaov: I'll pin them.

PR #1653, opened 2022-06-04
Reuse mutable combinators in immutable arrays

adithyaov:


core/src/Streamly/Internal/Data/Producer/Type.hs
```diff
@@ -112,6 +113,13 @@ translate f g (Producer step inject extract) =
 lmap :: (a -> a) -> Producer m a b -> Producer m a b
 lmap f (Producer step inject extract) = Producer step (inject . f) extract
 
+-- | Map the producer input using the provided translation functions
+--
+-- /Pre-release/
+{-# INLINE_NORMAL lmap_ #-}
+lmap_ :: Functor m => (a -> c) -> (c -> a) -> Producer m c b -> Producer m a b
+lmap_ f g (Producer step inject extract) = Producer step (inject . f) (fmap g . extract)
```
harendra-kumar, line 121: call it `lmap2` instead?

harendra-kumar (commented):
Need to fix benchmark regression first.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1511,51 +1517,63 @@ data ReadUState a = ReadUState
...
 
     inject (Array contents start end _) =
         let p = PTR_PREV(end,a)
          in return $ ReadUState contents start p
 
     {-# INLINE_LATE step #-}
-    step (ReadUState contents start p) | p < start = do
-        liftIO $ touch contents
+    step (ReadUState contents start p) | p < start = liftio $ do
+        touch contents
```
harendra-kumar, line 1564: Make this `!_ <- touch contents`

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1511,51 +1517,63 @@ data ReadUState a = ReadUState
...
         return D.Stop
     step (ReadUState contents start p) = do
-        x <- liftIO $ peek p
+        x <- liftio $ peek p
```
harendra-kumar, line 1567: Make this `!x <- liftio $ peek p`.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Array/Foreign.hs
```diff
@@ -213,6 +195,7 @@ producer = Producer step inject extract
 read :: forall m a. (Monad m, Storable a) => Unfold m (Array a) a
 read = Producer.simplify producer
 
+-- XXX This does not exist in mutable arrays
```
harendra-kumar, line 198: The comments below anyway say that it is not faster, so we can remove the XXX, because its not a todo.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -198,6 +204,10 @@ module Streamly.Internal.Data.Array.Foreign.Mut.Type
     )
 where
 
+-- Notes:
+-- When we use a purely lazy Monad like Identity, we need to force a few actions
+-- for correctness and execution order sanity.
```
harendra-kumar, line 209: "we need to force ordering of some actions for correctness."

core/src/Streamly/Internal/Data/Array/Foreign/Type.hs
```diff
@@ -124,6 +125,9 @@ import qualified Data.Foldable as F
 
 -- XXX Since these are immutable arrays MonadIO constraint can be removed from
 -- most places.
+-- XXX We need to be careful when removing the MonadIO constraint, we should
+-- make sure the order of the computations are maintained. We shouldn't use
+-- unsafeInlineIO anywhere allocations exist.
```
harendra-kumar, line 130: We can remove both these comments now. There is nothing more to do.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #1654, opened 2022-06-04
Deprecate variance and stdDev in folds

adithyaov:


harendra-kumar (approved):

harendra-kumar:
Please fix the CIs before merging:

```
 Streamly/Benchmark/Data/Fold.hs:313:23: error: [-Wdeprecations, -Werror=deprecations]
    In the use of ‘variance’
    (imported from Streamly.Internal.Data.Fold):
    Deprecated: "Use the streamly-statistics package instead"
    |
313 |               (S.fold FL.variance . S.map (fromIntegral :: Int -> Double))
    |                       ^^^^^^^^^^^
Streamly/Benchmark/Data/Fold.hs:317:23: error: [-Wdeprecations, -Werror=deprecations]
    In the use of ‘stdDev’ (imported from Streamly.Internal.Data.Fold):
    Deprecated: "Use the streamly-statistics package instead"
    |
317 |               (S.fold FL.stdDev . S.map (fromIntegral :: Int -> Double))
```

benchmark/Streamly/Benchmark/Data/Fold.hs
```diff
@@ -307,14 +307,6 @@ o_1_space_serial_elimination value =
...
               "mean"
               (S.fold FL.mean . S.map (fromIntegral :: Int -> Double))
-        , benchIOSink
-              value
-              "variance"
-              (S.fold FL.variance . S.map (fromIntegral :: Int -> Double))
-        , benchIOSink
-              value
-              "stdDev"
-              (S.fold FL.stdDev . S.map (fromIntegral :: Int -> Double))
```
harendra-kumar, line 317: We should not remove the benchmarks yet. We can remove those when we remove these functions altogether. Same about tests.

Can we suppress the warnings?

harendra-kumar (changes requested):

benchmark/Streamly/Benchmark/Data/Fold.hs
adithyaov: We can.

harendra-kumar (approved):

PR #1656, opened 2022-06-04
Rename the smart constructors of terminating folds

adithyaov:


core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -33,10 +33,8 @@ module Streamly.Internal.Data.Fold
     , foldl1'
     , foldr
     , foldrM
-    , mkFold
-    , mkFold_
-    , mkFoldM
-    , mkFoldM_
+    , foldt'
+    , foldtM'
```
harendra-kumar, line 37: Put these before the right folds above.

harendra-kumar (approved):

PR #1657, opened 2022-06-05
Add combinators to pin and unpin foreign arrays

adithyaov:


core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -321,13 +327,20 @@ touch (ArrayContents contents) =
...
 fptrToArrayContents :: ForeignPtrContents -> ArrayContents
 arrayToFptrContents :: ArrayContents -> ForeignPtrContents
+arrContentsToMutByteArr :: ArrayContents -> MutableByteArray# RealWorld
+byteArrToArrContents :: MutableByteArray# RealWorld -> ArrayContents
 #ifdef USE_FOREIGN_PTR
 fptrToArrayContents = ArrayContents
 arrayToFptrContents (ArrayContents contents) = contents
+arrContentsToMutByteArr (ArrayContents (PlainPtr mbarr)) = mbarr
+arrContentsToMutByteArr (ArrayContents _) = error "Unsupported foreign ptr"
+byteArrToArrContents mbarr = (ArrayContents (PlainPtr mbarr))
```
harendra-kumar, line 337: In case of foreign ptr we can make pin/unpin as no-op instead of erroring out.

adithyaov: Will that not be a little deciving?

adithyaov:
I will make this change after fixing the array module to work with both pinned and unpinned arrays.

harendra-kumar (approved):

Issue #1529, opened 2022-03-15
Move nubWindowBy to the Fold.Window module


harendra-kumar:

harendra-kumar:
The signature would be:

```
nub :: Fold m (a, Maybe a) (Maybe a)
```

PR #1659, opened 2022-06-05
Move nubwindowby to the Fold.Window module

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1529

harendra-kumar (changes requested):
Can you remove the pollCounts change from this?

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -351,3 +353,23 @@ mean = Fold.teeWith (/) sum length
...
+
+    where
+
+    initial =
+        return $ Partial Set.empty
+
+    step set (new, mOld) =
+        return $ Partial $
+            case mOld of
+                Just a -> Set.delete a set `seq` Set.insert new set
```
harendra-kumar, line 372: BUG 1: the delete is ineffective as you are inserting the new element in the original set in which the element is not deleted.

BUG 2: The element can be deleted from the Set only when the last reference of that element inside the current window is removed. We talked about this earlier, I had suggested to  use a refcount and decrement the refcount when an element is removed from the window, the element can be removed from the set when the refcount reaches 0.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -35,6 +35,7 @@ module Streamly.Internal.Data.Fold.Window
     --
       lmap
     , cumulative
+    , nub
```
harendra-kumar, line 38: This is not the right place to add it. Add it in a separate section at the end.

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -315,3 +318,35 @@ maximum n = fmap (fmap snd) $ range n
...
+-- /Pre-release/
+--
+{-# INLINE nub #-}
+nub :: (Monad m, Ord a) => Fold m (a, Maybe a) (Set.Set a)
+nub = Fold step initial extract
+
+    where
+
+    initial =
+        return $ Partial (Set.empty, Map.empty)
```
harendra-kumar, line 332: Do not use both a Set and a Map. Just the Map should be enough. Make the signature of the function as:

```
nub :: (Monad m, Ord a) => Fold m (a, Maybe a) (Map a Int)
```

We can just return a Map of values and their refcounts. The users can use it in whatever way they want.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -319,34 +318,32 @@ maximum n = fmap (fmap snd) $ range n
...
+        return $ Partial Map.empty
 
-    step (set, map0) (new, mOld) =
+    step map0 (new, mOld) =
         return $ Partial $
             case mOld of
                 Just a ->
                     let m1 = Map.insertWith (+) new (1::Int) map0
-                        (s1, m3) =
+                        m3 =
```
harendra-kumar, line 339: Rename to `m2`.

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -319,34 +318,32 @@ maximum n = fmap (fmap snd) $ range n
...
-                                        let m2 = Map.adjust (\x -> x - 1) a m1
-                                            in (set, m2)
-                                _ -> (set, m1)
-                               in (Set.insert new s1, m3)
-                Nothing -> let m1 = Map.insertWith (+) new (1::Int) map0
-                               in (Set.insert new set, m1)
-
-    extract = return . fst
+                                    then Map.delete a m1
+                                    else Map.adjust (\x -> x - 1) a m1
```
harendra-kumar, line 344: use a single `alterF` operation instead of using two operations (lookup and adjust).

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -319,34 +318,32 @@ maximum n = fmap (fmap snd) $ range n
...
+                                _ -> m1
```
harendra-kumar, line 345: This should never occur, we should use an error statement here to indicate a bug.

core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -319,34 +318,32 @@ maximum n = fmap (fmap snd) $ range n
...
+                               in  m3
+                Nothing -> Map.insertWith (+) new (1::Int) map0
```
harendra-kumar, line 347: This statement is common in Just and Nothing case, so we can factor this out.

harendra-kumar (changes requested):

PR #1680, opened 2022-06-15
Implemet the as Fold

rnjtranjan:


core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -2548,3 +2549,22 @@ top = topBy $ flip compare
...
 bottom = topBy compare
+
+{-# INLINE the #-}
+the :: (Eq a, Monad m) => Fold m a (Maybe a)
+the = Fold step initial extract
+
+    where
+
+    initial = do
+        return $ Partial Nothing
```
harendra-kumar, line 2560: redundant "do", in all functions.

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -2548,3 +2549,22 @@ top = topBy $ flip compare
...
+
+    step Nothing x = do
+        return $ Partial (Just x)
+
+    step (Just x0) x = do
+        if x0 == x
+        then return $ Partial (Just x)
+        else return $ Done Nothing
+
+    extract = return
```
harendra-kumar, line 2570: Just use `return` directly in `Fold` constructor.

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -2548,3 +2549,22 @@ top = topBy $ flip compare
...
+
+    initial = do
+        return $ Partial Nothing
+
+    step Nothing x = do
+        return $ Partial (Just x)
+
+    step (Just x0) x = do
+        if x0 == x
+        then return $ Partial (Just x)
```
harendra-kumar, line 2567: Better use `Just x0` here, or even better to use `step old@(Just x0)` and then use `old` here.

harendra-kumar (changes requested):
Are there any existing benchmark/tests for this which we can reuse?

Issue #1673, opened 2022-06-10
Write init and tail as stream transformations


harendra-kumar:
```
init :: Monad m => Stream m a -> Stream m a
tail :: Monad m => Stream m a -> Stream m a
```

harendra-kumar:
As a stream transformation `tail` is just `drop 1` except that `tail` errors out on empty srteam whereas `drop` does not.

We can generalize `tail` and `init` to `drop` and `dropEnd`. To imitate the `tail` and `init` behavior we may need `dropExact` and `dropEndExact` variants that error out if less than specified elements are dropped.

Also, once we have composable Scan type, we can write these as scans instead of stream specific implementations.

PR #1688, opened 2022-06-25
Implement init and tail as Stream transformations

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1673

harendra-kumar (commented):
The StreamD implementation is good for applying these operations a few times, but it will be have bad performance because of quadratic slowdown if we use these heavily. We will need StreamK implementations for that. 

Though we can keep these as well because they will fuse and be faster in the light use case. We need to check the benchmarks once the benchmarks PR is merged.

PR #1690, opened 2022-06-25
Implement Data.Stream.Top module

rnjtranjan:
fixes:https://github.com/composewell/streamly/issues/1689


core/src/Streamly/Internal/Data/Stream/Top.hs
```diff
@@ -1,3 +1,6 @@
+-- XXX This entire documentation is duplicated. Can you implement IsStream.Top
+-- in terms of this module? Un-duplicate as much code and doc as possible.
```
harendra-kumar, line 2: IsStream would be deprecated, we won't be maintaining it once we start using Data.Stream.

rnjtranjan (commented):
looks good.

PR #1691, opened 2022-06-27
Time related Unfold APIs

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1676

harendra-kumar (commented):
This unfold is exposing its internal state as the input of the unfold. That's bad.

The correct input for this unfold would be `()`. But then we better generate a Stream type.

harendra-kumar:
Closing this in favor of the Stream APIs.

PR #1692, opened 2022-06-27
Add Timed API for stream generations

rnjtranjan:


core/src/Streamly/Internal/Data/Stream/Transform.hs
```diff
@@ -262,6 +263,9 @@ foldrT f z s = D.foldrT f z (toStreamD s)
 mapM :: Monad m => (a -> m b) -> Stream m a -> Stream m b
 mapM f m = fromStreamK $ D.toStreamK $ D.mapM f $ toStreamD m
 
+{-# INLINE mapM_ #-}
+mapM_ :: Monad m => (a -> m b) -> Stream m a -> m ()
+mapM_ f = K.mapM_ f . toStreamK
```
harendra-kumar, line 268: We have decided not to add folds, instead use `Stream.fold Fold.drainBy`.

PR #1700, opened 2022-07-03
Add TreeArray variant to Arrays

twitu:
A TreeArray grows in fixed size chunks allocated as mutable unboxed arrays. It is useful for cases where are large amount of data of unknown length needs to be written to an array. A typical array after becoming full would have to reallocate double it's original size, increasing peak memory usage. A tree array will allocate new memory in fixed sized chunks keeping peak memory usage closer to actual memory usage.

There is also the concern that for very large data use cases, an array may not find a continuous of memory to allocate. In such case again, tree array with it's smaller fixed size allocation will still be able to operate.

A TreeArray has configurable `branchWidth` i.e. the number of children each branch can have and configurable `leafWidth` i.e. the number of elements a single leaf can store.


harendra-kumar:
Moved to another package.

PR #1702, opened 2022-07-07
Add some test cases for unboxed array

adithyaov:


test/Streamly/Test/Data/Array/Foreign.hs
```diff
@@ -229,6 +248,21 @@ testAsPtrUnsafeMA = do
...
+      forAll (chooseInt (bSize - 100, bSize + 100)) $ \newBLen -> do
+          arr <- MA.fromList vec
+          arr1 <- MA.realloc newBLen arr
+          lst <- MA.toList arr
+          lst1 <- MA.toList arr1
+          lst `shouldBe` lst1
+
+    where
+
+    len = 10000
```
harendra-kumar, line 263: Shouldn't this be length of `vec`? If the vector is larger than 10000 then it will get truncated after reallocation. If we want to test truncation as well then we should truncate the original list to the max length of the array before comparing.

test/Streamly/Test/Data/Array/Stream/Foreign.hs
```diff
@@ -55,6 +56,14 @@ parseBreak = do
                 ls2 <- run $ Stream.toList $ ArrayStream.concat str
                 listEquals (==) (ls1 ++ ls2) ls
 
+splitOnSuffix :: Word8 -> [Word8] -> [[Word8]] -> IO ()
+splitOnSuffix sep inp out = do
+    res <-
+        Stream.toList
+            $ ArrayStream.splitOnSuffix sep
+            $ Stream.chunksOf 2 (Array.writeN 3) $ Stream.fromList inp
```
harendra-kumar, line 64: Why not `Array.writeN 2`?

test/Streamly/Test/Data/Array/Foreign.hs
```diff
@@ -229,6 +248,21 @@ testAsPtrUnsafeMA = do
...
 
+reallocMA :: Property
+reallocMA =
+    forAll (vectorOf len (arbitrary :: Gen Char)) $ \vec ->
+      forAll (chooseInt (bSize - 100, bSize + 100)) $ \newBLen -> do
+          arr <- MA.fromList vec
+          arr1 <- MA.realloc newBLen arr
+          lst <- MA.toList arr
+          lst1 <- MA.toList arr1
+          lst `shouldBe` lst1
```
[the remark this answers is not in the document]
adithyaov: We can, I don't want to use the array `==` though. Only testing once specific array function.

PR #1718, opened 2022-07-18
Fixing the "Respect unpinned" - minimizing diff

adithyaov:


.github/workflows/haskell.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     continue-on-error: ${{ matrix.ignore_error }}
     strategy:
-      fail-fast: true
+      fail-fast: false
```
adithyaov, line 53: Revert this

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1048,53 +951,40 @@ snoc = snocWith f
 -- Random reads
 -------------------------------------------------------------------------------
 
+-- XXX Can this be deduplicated with array/foreign
```
adithyaov, line 954: What does this mean?

harendra-kumar: Data.Array.Foreign also has getIndex/getIndexSafe functions with similar code. Maybe we can use common implementation here, passing a liftio argument.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -525,42 +518,38 @@ modifyIndexUnsafe i f (Array{..}) = liftIO $ do
...
-        pokeWith arrContents index x
-        return res
-    else invalidIndex "modifyIndex" i
-
 -- | Modify a given index of an array using a modifier function.
 --
 -- /Pre-release/
 modifyIndex :: forall m a b. (MonadIO m, Storable a) =>
     Int -> (a -> (a, b)) -> Array a -> m b
 -- XXX modifyIndexPtr can be removed now
```
harendra-kumar: Remove this comment and add an INLINE on this function.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -525,42 +518,38 @@ modifyIndexUnsafe i f (Array{..}) = liftIO $ do
...
+modifyIndices f arr= FL.foldlM' step initial
 
     where
 
     initial = return ()
 
     step () i =
         let f1 x = (f i x, ())
-         in liftIO (modifyIndexPtr arrContents i f1 arrStart aEnd)
+         in liftIO (modifyIndex i f1 arr)
```
harendra-kumar, line 551: Remove the liftIO.

core/src/Streamly/Internal/Data/Array/Foreign/Mut/Type.hs
```diff
@@ -1036,11 +1010,20 @@ getIndicesD liftio (D.Stream stepi sti) = Unfold step inject
         r <- stepi defState st
         case r of
             D.Yield i s -> do
-                x <- liftio $ getIndexPtr contents start end i
+                x <- liftio $ getIndex_ contents start end i
```
harendra-kumar, line 1012: We can use `getIndex i (Array contents start end undefined)`

Issue #1723, opened 2022-07-19
Replace the SerialT type in all modules by the "Stream" type


harendra-kumar:
In all internal modules except:

* IsStream/*

In all released modules except:

* Streamly.Prelude

harendra-kumar:
This is not yet fixed. There are still many references to SerialT in several modules that need to be changed to Stream.

rnjtranjan:
Do we need to change test/benchmarks also?

PR #1726, opened 2022-07-20
Replace SerialT with  Stream

rnjtranjan:


rnjtranjan:
fixes:https://github.com/composewell/streamly/issues/1723

PR #1731, opened 2022-07-22
Fix the timer tick in chunksOfTimeout

adithyaov:


src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
```diff
@@ -1013,7 +1013,7 @@ chunksOfTimeout :: (IsStream t, MonadAsync m, Functor (t m))
 chunksOfTimeout n timeout f =
       map snd
     . classifySessionsBy
-        timeout False (const (return False)) timeout (FL.take n f)
+        0.01 False (const (return False)) timeout (FL.take n f)
```
harendra-kumar, line 1016: 10 ms may be too small. 100 ms may be a better default here.

harendra-kumar (changes requested):

PR #1742, opened 2022-07-29
Remove prim arrays and all related code

adithyaov:


core/src/Streamly/Internal/Data/IORef/Prim.hs
```diff
@@ -42,56 +41,58 @@ where
...
+    , peekWith
+    , pokeWith
+    , newUnpinnedArrayContents
+    )
 
 import qualified Streamly.Internal.Data.Stream.StreamD.Type as D
 
 -- | An 'IORef' holds a single 'Prim' value.
-data IORef a = IORef (MutableByteArray# RealWorld)
+type IORef a = ArrayContents a
```
harendra-kumar, line 55: This should be a newtype. This module would become a user facing module, ArrayContents is an internal type which should not be exposed via this.

harendra-kumar (changes requested):
Rename IORef.Prim to IORef.Unboxed.

harendra-kumar (approved):

PR #1751, opened 2022-08-01
Move Prelude.Serial benchmarks to Data.Stream

rnjtranjan:


.github/workflows/regression-check.yml
```diff
@@ -33,7 +33,6 @@ jobs:
         Prelude.Ahead
         Prelude.Async:12
         Prelude.Parallel
-        Prelude.Serial
```
harendra-kumar, line 36: We need to add Data.Stream in this list, otherwise we will not be doing regression checks on it.

.circleci/config.yml
```diff
@@ -387,7 +387,7 @@ jobs:
             environment:
               BUILD: cabal-v2
               HLINT_OPTIONS: lint
-              HLINT_TARGETS: src test benchmark
```
harendra-kumar, line 390: Add back `benchmark`. modify .hlint.yaml instead.

benchmark/Streamly/Benchmark/Data/Stream.hs
```diff
@@ -1,6 +1,6 @@
 -- |
--- Module      : Serial
--- Copyright   : (c) 2018 Composewell Technologies
+-- Module      : Data.Stream
+-- Copyright   : (c) 2022 Composewell Technologies
```
harendra-kumar, line 3: Copyright year means "starting from that year", it should remain 2018 because the original code was written in that year.

benchmark/bench-runner/Main.hs
```diff
@@ -35,22 +35,6 @@ rtsOpts exeName benchName0 = unwords [general, exeSpecific, benchSpecific]
...
-        | "Prelude.Serial/o-n-heap.transformer." `isPrefixOf` benchName =
-            "-K8M -M64M"
-        | "Prelude.Serial/o-n-space.Functor." `isPrefixOf` benchName =
-            "-K4M -M64M"
-        | "Prelude.Serial/o-n-space.Applicative." `isPrefixOf` benchName =
-            "-K8M -M128M"
-        | "Prelude.Serial/o-n-space.Monad." `isPrefixOf` benchName =
-            "-K8M -M64M"
-        | "Prelude.Serial/o-n-space.grouping." `isPrefixOf` benchName = ""
-        | "Prelude.Serial/o-n-space." `isPrefixOf` benchName = "-K4M"
```
harendra-kumar, line 53: Don't we need to add these options for the new Data.Stream benchmarks? Otherwise those will fail to run. Ideally we just need to replace Prelude.Serial with Data.Stream in the code above.

benchmark/streamly-benchmarks.cabal
```diff
@@ -64,13 +64,20 @@ flag bench-core
   manual: True
   default: False
 
+flag use-prelude
+  description: Benchmark using streamly-prelude
```
harendra-kumar, line 68: "Use Streamly.Prelude instead of Streamly.Data.Stream for serial benchmarks"

targets/Targets.hs
```diff
@@ -25,15 +25,7 @@ targets =
...
     -- Streams
-    , ("Prelude.Serial", ["serial_wserial_cmp"])
     , ("Prelude.WSerial", ["serial_wserial_cmp"])
-    , ("Prelude.Serial",
-            [ "prelude_serial_grp"
-            , "infinite_grp"
-            , "serial_wserial_cmp"
-            , "serial_async_cmp"
-            ]
-      )
```
harendra-kumar, line 36: Don't remove these. Just replace Prelude.Serial with Data.Stream.

harendra-kumar (changes requested):

.hlint.yaml
```diff
@@ -23,15 +23,17 @@
 - ignore: {name: "Use <$>"}
 - ignore: {name: "Use uncurry"}
 - ignore: {name: "Redundant $!"}
+- ignore: {name: "Redundant bracket"}
```
harendra-kumar, line 26: We should not add this, it will suppress all redundant brackets. Do we want that?

benchmark/Streamly/Benchmark/Data/Stream/Common.hs
```diff
@@ -0,0 +1,71 @@
...
+    ( sourceUnfoldr
+    , sourceUnfoldrM
+    , sourceUnfoldrAction
+    , benchIOSink
+    , benchIOSrc
+    )
+where
+
+import Streamly.Internal.Data.Stream (Stream, unfold)
+--import qualified Streamly.Internal.Data.Stream as Stream
```
harendra-kumar, line 17: Remove.

benchmark/Streamly/Benchmark/Data/Stream/Common.hs
```diff
@@ -0,0 +1,71 @@
...
+import qualified Streamly.Prelude  as Stream
```
harendra-kumar, line 18: Why are we using Streamly.Prelude here? Why not Data.Stream?

benchmark/Streamly/Benchmark/Data/Stream/Eliminate.hs
```diff
@@ -238,6 +257,7 @@ benchPureSink :: NFData b
     => Int -> String -> (SerialT Identity Int -> b) -> Benchmark
 benchPureSink value name = benchPure name (sourceUnfoldr value)
 
+#ifdef USE_PRELUDE
```
harendra-kumar, line 260: Can we replace the `t` type in thee routines with `Stream` instead? If so we do not need to use the `USE_PRELUDE` conditional here.

benchmark/Streamly/Benchmark/Data/Stream/Eliminate.hs
```diff
@@ -398,40 +426,47 @@ drainWhile = S.drainWhile (const True)
...
-                  , benchIdentitySink value "foldrToStreamLength"
-                        (S.length . runIdentity . foldrToStream)
-                  , benchPureSink value "foldrMToListLength"
+                  [ benchPureSink value "foldrMToListLength"
                         (Prelude.length . runIdentity . foldrMBuild)
+
+                  , benchIdentitySink value "foldrMElem" (foldrMElem value)
+                  , benchIdentitySink value "foldrToStreamLength"
+                        (S.fold Fold.length . runIdentity . foldrToStream)
+
```
harendra-kumar, line 463: Why change the ordering of these benchmarks, that just adds to the diff to be reviewed.

benchmark/Streamly/Benchmark/Data/Stream/Eliminate.hs
```diff
@@ -453,23 +488,25 @@ o_1_space_elimination_folds value =
...
         , benchIOSink value "minimumBy" minimumBy
         , benchIOSink value "minimum" minimum
-
+#ifdef USE_PRELUDE
         , bench "the" $ nfIO $ randomRIO (1,1) >>= the . repeat value
+#endif
         , benchIOSink value "find" (find value)
         , benchIOSink value "findM" (findM value)
-        -- , benchIOSink value "lookupFirst" (lookup 1)
+        , benchIOSink value "lookupFirst" (lookup 1)
```
harendra-kumar, line 496: Uncommenting this will create problems with reporting of benchmarking times, beacuse all times now will come in nanoseconds, as this benchmark is too small.

benchmark/Streamly/Benchmark/Data/Stream/Eliminate.hs
```diff
@@ -453,23 +488,25 @@ o_1_space_elimination_folds value =
...
         , benchIOSink value "lookupNever" (lookup (value + 1))
         , benchIOSink value "(!!)" (value !!)
         , benchIOSink value "findIndex" (findIndex value)
         , benchIOSink value "elemIndex" (elemIndex value)
         -- this is too fast, causes all benchmarks reported in ns
-    -- , benchIOSink value "null" S.null
+        , benchIOSink value "null" S.null
```
harendra-kumar, line 502: Benchmarks is too small, should not uncomment this. See the comment before this which tells why this was commented.

benchmark/Streamly/Benchmark/Data/Stream/Eliminate.hs
```diff
@@ -494,11 +531,12 @@ o_n_heap_elimination_foldl value =
         -- Left folds for building a structure are inherently non-streaming
         -- as the structure cannot be lazily consumed until fully built.
         [ benchIOSink value "foldl'/build/IO" foldl'Build
-        , benchIdentitySink value "foldl'/build/Identity" foldl'Build
         , benchIOSink value "foldlM'/build/IO" foldlM'Build
+        , benchIdentitySink value "foldl'/build/Identity" foldl'Build
```
harendra-kumar, line 535: Why change the order? We had `foldl'` IO/Identity benchmarks together for comparison.

benchmark/Streamly/Benchmark/Data/Stream/Exceptions.hs
```diff
@@ -47,6 +55,28 @@ import qualified Streamly.Internal.Data.Stream.StreamD as D
 -- stream exceptions
 -------------------------------------------------------------------------------
 
+drainG :: SerialT IO a -> IO ()
+#ifdef USE_PRELUDE
+drainG = Stream.drain
+#else
+drainG = Stream.fold Fold.drain
+#endif
```
harendra-kumar, line 63: We can name this just `drain`.

benchmark/Streamly/Benchmark/Data/Stream/Exceptions.hs
```diff
@@ -47,6 +55,28 @@ import qualified Streamly.Internal.Data.Stream.StreamD as D
...
+
+sourceG :: Int -> Int -> SerialT IO Int
+#ifdef USE_PRELUDE
+sourceG length from = S.enumerateFromTo from (from + length)
+#else
+sourceG length from = Stream.unfold Unfold.enumerateFromTo (from, from + length)
+#endif
```
harendra-kumar, line 70: We can name this just enumerateFromTo.

benchmark/Streamly/Benchmark/Data/Stream/Generate.hs
```diff
@@ -151,6 +164,40 @@ o_1_space_generation value =
         , benchIOSrc fromSerial "absTimes" $ absTimes value
         ]
     ]
+#else
```
harendra-kumar, line 167: Why aren't these benchamrks common to both Prelude/Data.Stream cases? Why are these only for Data.Stream case?

harendra-kumar (changes requested):
Did you build it with use-prelude flag? There is no CI so we have to do it manually.

benchmark/Streamly/Benchmark/Data/Stream/Generate.hs
rnjtranjan: bcoz all of them are polymorphic t m a.

harendra-kumar: I am talking about the `else` case. The else case should be common and not conditional.

Issue #1761, opened 2022-08-06
Fix hlint on streamly-core package files


harendra-kumar:
See .hlint.ignore. We need to fix all the files ignored for `core/` directory.

PR #1769, opened 2022-08-10
Fix hlint streamly-core

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1761

rnjtranjan (commented):
looks good

Issue #1737, opened 2022-07-25
Repurpose the streamk build flag


harendra-kumar:
Currently this flag enables USE_STREAMK_ONLY cpp macro. We should remove all usage of that macro.

Use the `streamk` build flag to disable rewrite-rules instead i.e. enable the DISABLE_FUSION macro when using this flag. Make sure that we disable all rewrite rules when this macro is used. This way we can test the code paths that kick in when rewrite rules do not fire.

rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1737

PR #1774, opened 2022-08-11
Repurpose the streamk build flag

rnjtranjan:


streamly.cabal
```diff
@@ -319,7 +319,7 @@ common compile-options
       cpp-options:    -DCABAL_OS_WINDOWS
 
     if flag(streamk)
-      cpp-options:    -DUSE_STREAMK_ONLY
+      cpp-options:    -DDISABLE_FUSION
```
harendra-kumar, line 322: 1. There is already a `no-fusion` flag in streamly-core. So let's use that flag for fusion disabled build instead of using streamk.
2. We do not need to define the DISABLE_FUSION macro in the streamly package, as it is not used anywhere, it is only used in the streamly-core package.
3. Remove the streamk flag from streamly as well as streamly-core packages.
4. Remove streamk build from CI and use the no-fusion flag build instead.

harendra-kumar (changes requested):

rnjtranjan:
Fixes: #1737

src/Streamly/Internal/Data/Stream/IsStream/Common.hs
```diff
@@ -75,7 +75,7 @@ import Streamly.Internal.Data.Array.Unboxed.Type (Array)
 import Streamly.Internal.Data.Fold.Type (Fold (..))
 import Streamly.Internal.Data.Stream.IsStream.Combinators (maxYields)
 import Streamly.Internal.Data.Stream.IsStream.Type
-    (IsStream(..), fromStreamD, toStreamD, fromStreamS, toStreamS)
+    (IsStream(..), fromStreamD, toStreamD, fromStreamD, toStreamD)
```
harendra-kumar, line 78: These are already imported, remove duplication.

src/Streamly/Internal/Data/Stream/IsStream/Eliminate.hs
```diff
@@ -163,7 +163,7 @@ import Streamly.Internal.Data.Stream.IsStream.Common
     ( fold, foldBreak, foldContinue, drop, findIndices, reverse, splitOnSeq
     , take , takeWhile, mkParallel)
 import Streamly.Internal.Data.Stream.IsStream.Type
-    (IsStream, toStreamS, fromStreamD, toStreamD)
+    (IsStream, toStreamD, fromStreamD, toStreamD)
```
harendra-kumar, line 166: Remove duplication of imports.

src/Streamly/Internal/Data/Stream/IsStream/Generate.hs
```diff
@@ -99,7 +99,7 @@ import Streamly.Internal.Data.Stream.IsStream.Common
     ( absTimesWith, concatM, relTimesWith, timesWith, fromPure, fromEffect
     , yield, yieldM, repeatM)
 import Streamly.Internal.Data.Stream.IsStream.Type
-    (IsStream (..), fromSerial, consM, fromStreamD, fromStreamS)
+    (IsStream (..), fromSerial, consM, fromStreamD, fromStreamD)
```
harendra-kumar, line 102: Remove duplication of imports.

src/Streamly/Internal/Data/Stream/IsStream/Lift.hs
```diff
@@ -29,15 +29,10 @@ import Control.Monad.Trans.State.Strict (StateT)
 import Control.Monad.Trans.Class (MonadTrans(..))
 import Data.Functor.Identity (Identity (..))
 import Streamly.Internal.Data.Stream.IsStream.Type
-    (IsStream(..), fromStreamS, toStreamS, fromStreamD, toStreamD)
+    (IsStream(..), fromStreamD, toStreamD, fromStreamD, toStreamD)
```
harendra-kumar, line 32: Remove duplication.

src/Streamly/Internal/Data/Stream/IsStream/Type.hs
```diff
@@ -728,8 +711,8 @@ concatMapWith par f xs = bindWith par xs f
 -- Equivalent to:
 --
 -- @
--- concatMapFoldableWith f g = Prelude.foldr (f . g) S.nil
--- concatMapFoldableWith f g xs = S.concatMapWith f g (S.fromFoldable xs)
+-- concatMapFoldableWith f g = Prelude.foldr (f . g) D.nil
+-- concatMapFoldableWith f g xs = D.concatMapWith f g (D.fromFoldable xs)
```
harendra-kumar, line 715: User examples should not mention D or K. We should always use `Stream` or `S` in the user documentation.

harendra-kumar (changes requested):
We have not removed the `streamk` CI. We should change that to `no-fusion` CI.

streamly.cabal
harendra-kumar: > Remove streamk build from CI and use the no-fusion flag build instead.

This is not take care of yet.

harendra-kumar (approved):
Squash the commits for merge.

PR #1778, opened 2022-08-12
Create a Streamly.Data.Stream module in streamly-core

rnjtranjan:


core/streamly-core.cabal
```diff
@@ -342,6 +342,7 @@ library
                      , Streamly.Internal.Data.Stream.StreamK
 
                      -- streamly-core exposed modules
+                     , Streamly.Data.Stream
```
harendra-kumar, line 345: This should be re-exported via streamly package as well.

harendra-kumar (approved):

PR #1783, opened 2022-08-17
Move Streamly.Internal.Data.SmallArray under dev flag

rnjtranjan:


streamly.cabal
```diff
@@ -709,3 +707,8 @@ library
     -- tests fail
     if flag(dev) && flag(inspection)
       build-depends: inspection-and-dev-flags-cannot-be-used-together
+      
+    if flag(dev)
+      exposed-modules:  Streamly.Internal.Data.SmallArray
+                      , Streamly.Internal.Data.SmallArray.Type
+
```
harendra-kumar, line 714: Would be better if we put this before the build-depends section, next to other modules.

streamly.cabal
```diff
@@ -643,6 +637,10 @@ library
     other-modules:
                        Streamly.Data.Array
                      , Streamly.Data.SmallArray
+    if !flag(dev)
+        other-modules:
+                       Streamly.Internal.Data.SmallArray
+                     , Streamly.Internal.Data.SmallArray.Type
```
harendra-kumar, line 643: Should be `if flag(dev)`  then `exposed-modules:`

harendra-kumar (changes requested):

test/streamly-tests.cabal
```diff
@@ -270,6 +270,10 @@ test-suite Data.SmallArray
   type: exitcode-stdio-1.0
   main-is: Streamly/Test/Data/SmallArray.hs
   ghc-options: -main-is Streamly.Test.Data.SmallArray.main
+  if flag(dev)
+    buildable: True
+  else
+    buildable: False
```
harendra-kumar, line 276: We just need the `else` case, as the default is True.

streamly.cabal
```diff
@@ -691,13 +689,11 @@ library
 
                      -- XXX to be removed
                      , transformers-base >= 0.4   && < 0.5
-                     , primitive         >= 0.5.4 && < 0.8
```
harendra-kumar, line 694: Does it build with the `dev` flag? This dependency is remove unconditionally, isn't it needed for building with SmallArray in `dev` flag case?

streamly.cabal
```diff
@@ -691,13 +689,11 @@ library
...
                      , heaps             >= 0.3     && < 0.5
 
                     -- concurrency
                      , atomic-primops    >= 0.8   && < 0.9
                      , lockfree-queue    >= 0.2.3 && < 0.3
-
```
harendra-kumar, line 700: Do not remove this line, it groups the unicode dependency with concurrency dependencies.

core/src/Streamly/Internal/Data/Array/Unboxed/Type.hs
```diff
@@ -55,7 +55,9 @@ module Streamly.Internal.Data.Array.Unboxed.Type
     , toStream
     , toStreamRev
     , toList
-
+#ifdef DEVBUILD
+    , toStreamD_
+#endif
```
harendra-kumar, line 60: We should not use conditional here unless that is really needed, this is an internal module we can export whatever we need from this.

benchmark/streamly-benchmarks.cabal
```diff
@@ -441,10 +441,10 @@ benchmark Data.Stream.StreamDK
   type: exitcode-stdio-1.0
   hs-source-dirs: Streamly/Benchmark/Data/Stream
   main-is: StreamDK.hs
-  if impl(ghcjs)
-    buildable: False
-  else
+  if flag(dev) && !impl(ghcjs)
     buildable: True
+  else
+    buildable: False
```
harendra-kumar, line 447: We can make it:

```
if !flag(dev || impl(ghcjs)
   buildable: False
```
```

**Review by harendra-kumar — CHANGES_REQUESTED** [review]
We will have to modify the bench-runner/test-runner targets as well.

**Comment by harendra-kumar** [comment]
The regression CI is failing to build. We will have to modify the bench-runner/test-runner targets as well.

File: benchmark/bench-runner/Main.hs [hunk 8]
```diff
@@ -74,7 +75,9 @@ rtsOpts exeName benchName0 = unwords [general, exeSpecific, benchSpecific]
         | "Data.Parser.ParserK/o-n-heap.someAlt" == benchName = "-K4M -M128M"
         | "Data.Parser.ParserK/o-n-heap.sequence" == benchName = "-M64M"
         | "Data.Parser.ParserK/o-n-heap.sequenceA" == benchName = "-M64M"
+#ifdef DEVBUILD
         | "Data.SmallArray.o-1-sp" `isPrefixOf` benchName = "-K128K"
+#endif
```
harendra-kumar, line 80: This code will never be used as this file is never built with DEVBUILD flag. So better remove it than using it as dead code.

PR #1795, opened 2022-08-21
Remove redundant additional state in fromParserK

harendra-kumar:


harendra-kumar:
Benchmarks show no significant difference but the code is simplified.

adithyaov (approved):

PR #1796, opened 2022-08-21
Make streamly-core free from IsStream

rnjtranjan:


core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -416,12 +416,11 @@ import Prelude hiding (concatMap, filter, foldr, map, take)
 -- >>> import Streamly.Data.Fold (Fold)
 -- >>> import Streamly.Internal.Data.Stream.Type (Stream)
 -- >>> import qualified Data.Foldable as Foldable
--- >>> import qualified Streamly.Prelude as Stream
+-- >>> import qualified Streamly.Prelude as PStream
```
harendra-kumar, line 419: The convention is to import it as IsStream, that way we can also find it easily to replace later.

harendra-kumar (changes requested):
I will make minor changes and push it.

harendra-kumar:
Can you also replace the instances of `Streamly.Prelude` imports with `Streamly.Data.Stream`?

rnjtranjan:
Currently, WserialT and ZipSerialT are wrapped on StreamK that need to be on Stream.

core/src/Streamly/Internal/Data/Stream/StreamD/Nesting.hs
```diff
@@ -555,6 +555,19 @@ data ConcatUnfoldInterleaveState o i =
...
+-- goes to the last stream it reverses the traversal to come back to the first
+-- stream yielding elements from each stream on its way back to the first
+-- stream and so on.
+--
+-- >>> input = Stream.fromList [[1,1],[2,2],[3,3],[4,4],[5,5]]
+-- >>> Stream.fold Fold.toList $ Stream.unfoldManyInterleave Unfold.fromList input
+-- [1,2,3,4,5,5,4,3,2,1]
+--
+-- Note that this is order of magnitude more efficient than "concatPairsWith
+-- wSerial"
```
harendra-kumar, line 570: Remove this. This is already present in Streamly.Internal.Data.Stream, we do not want duplication.

harendra-kumar (commented):
Some imports are out of order, some may be redundant e.g. importing Streamly.Data.Stream as well as Streamly.Internal.Data.Stream. This is not mandatory to fix though, but a nice to have.

Issue #1789, opened 2022-08-19
Combine test-core and bench-core flags to a single use-streamly-core flag


harendra-kumar:
We can also add a CI with this flag enabled.

PR #1798, opened 2022-08-21
Combine test-core and bench-core flags to a single use-streamly-core …

rnjtranjan:
…flag


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1789

harendra-kumar (commented):
We already have  a `streamly-core` flag in streamly package, can we use the same here?

PR #1800, opened 2022-08-22
Do not re-export the Internal modules of streamly-core from streamly

rnjtranjan:


harendra-kumar (commented):
Why are some Internal modules still re-exported? Ideally, reexported section should have only exposed modules from streamly-package.

PR #1809, opened 2022-08-25
Add modular scanning functions for transformations

harendra-kumar:
Usable across folds and streams. Implemented common transformations
using scans. Performance of single operation benchmark seems to be
unaffected but in some cases combinations of multiple benchmarks or the
same oepration multiple times is degraded.

The perf impact could perhaps be improved if we have a separate Scan
type where we do not need a pre initialization of state.

### DIFF
--- a/core/src/Streamly/Data/Stream.hs
+++ b/core/src/Streamly/Data/Stream.hs
@@ -227,7 +227,7 @@ module Streamly.Data.Stream
     , uncons
 
     -- ** Folding
-    -- XXX Need to have a general parse operations here which can be used to
+    -- XXX Need to have a general parse operation here which can be used to
     -- express all others.
     , fold -- XXX rename to run? We can have a Stream.run and Fold.run.
     -- XXX fold1 can be achieved using Monoids or Refolds.
@@ -399,7 +399,7 @@ module Streamly.Data.Stream
 
     -- * Buffered Operations
     -- | Operations that require buffering of the stream.
-    -- Reverse is essentially a left fold followed by unfold.
+    -- Reverse is essentially a left fold followed by an unfold.
     , reverse
 
     -- * Multi-Stream folds
@@ -425,8 +425,8 @@ module Streamly.Data.Stream
     , handle
 
     -- * Resource Management
-    -- | 'bracket' is the most general operation, all other operations can be
-    -- expressed using it.
+    -- | 'bracket' is the most general resource management operation, all other
+    -- operations can be expressed using it.
     , before
     , after
     , finally
--- a/core/src/Streamly/Data/Stream.hs
+++ b/core/src/Streamly/Data/Stream.hs
@@ -101,7 +101,8 @@ module Streamly.Data.Stream
     -- , consM2 -- fused version
 
     -- ** Unfolding
-    -- | Generalized way of generating a stream efficiently.
+    -- | 'unfoldrM' is the most general way of generating a stream efficiently.
+    -- All other generation operations can be expressed using it.
     , unfoldr
     , unfoldrM
 
@@ -226,6 +227,8 @@ module Streamly.Data.Stream
     , uncons
 
     -- ** Folding
+    -- XXX Need to have a general parse operations here which can be used to
+    -- express all others.
     , fold -- XXX rename to run? We can have a Stream.run and Fold.run.
     -- XXX fold1 can be achieved using Monoids or Refolds.
     , foldBreak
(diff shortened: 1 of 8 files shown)


core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -12,9 +12,22 @@
...
 --
--- == Folds that never terminate
---
--- An Accumulator is the simplest type of fold, it never fails and never
+-- * Accumulators: Tee/Zip is simple, cannot be appended, good for scanning.
+-- * Terminating folds: Tee/Zip varies based on termination, can be appended,
+--   good for scanning, nesting (many) is easy.
+-- * Non-failing (backtracking only) parsers: cannot be used as scans because
+--   of backtracking, nesting is complicated because of backtracking, appending
+--   is efficient because of no ALternative, alternative does not make sense
```
adithyaov, line 20:
```suggestion
--   is efficient because of no alternative, alternative does not make sense
```

adithyaov: Or `Alternative`

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -1110,26 +1131,113 @@ lmapM f (Fold step begin done) = Fold step' begin done
...
 
+-- | Postscan the input of a 'Fold' to change it in a stateful manner using
+-- another 'Fold'.
+--
+-- @postscan scanner collector@
+--
+-- /Pre-release/
+{-# INLINE postscan #-}
+postscan :: Monad m => Fold m a b -> Fold m b c -> Fold m a c
+postscan (Fold stepL initialL extractL) (Fold stepR initialR extractR) =
```
adithyaov, line 1142: L = Inner
R = Outer

core/src/Streamly/Internal/Data/Fold/Type.hs
```diff
@@ -1110,26 +1131,113 @@ lmapM f (Fold step begin done) = Fold step' begin done
...
+                    $ case rR of
+                        Partial sR1 -> Partial (sL, sR1)
+                        Done bR -> Done bR
+
+    initial = do
+        r <- initialR
+        rL <- initialL
+        case r of
+            Partial sR ->
+                case rL of
```
adithyaov, line 1169: Move `rL <- initialL` here?

adithyaov: Isn't it an extra effect of `r` is `Done`.

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -390,35 +425,59 @@ sequence = rmapM id
...
+--
+-- >>> Stream.fold (Fold.lmapM (Fold.tracing print) Fold.drain) $ (Stream.enumerateFromTo (1 :: Int) 2)
+-- 1
+-- 2
+--
+-- /Pre-release/
+--
+{-# INLINE tracing #-}
+tracing :: Monad m => (a -> m b) -> (a -> m a)
+tracing f x = void (f x) >> return x
```
adithyaov, line 466: Why is this in this module?

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -516,82 +575,110 @@ scan = scanWith False
...
+    -- (Monad m, Eq a) =>
+    (a -> Bool) -> Fold m a (Maybe a)
+prune = error "Not implemented yet!"
+
+-- | Emit only repeated elements, once.
+--
+-- /Unimplemented/
+repeated :: -- (Monad m, Eq a) =>
+    Fold m a (Maybe a)
+repeated = undefined
```
adithyaov, line 681: Any special thing? `error "Not implemented yet!"` vs `undefined`?

core/src/Streamly/Internal/Data/Stream/Bottom.hs
```diff
@@ -313,23 +339,27 @@ smapM step initial stream =
 --
 {-# INLINE take #-}
 take :: Monad m => Int -> Stream m a -> Stream m a
-take n m = fromStreamD $ D.take n $ toStreamD m
+-- take n m = fromStreamD $ D.take n $ toStreamD m
```
adithyaov, line 342: Why is this `scanMaybe`?

core/src/Streamly/Data/Stream.hs
```diff
@@ -414,6 +425,8 @@ module Streamly.Data.Stream
     , handle
 
     -- * Resource Management
+    -- | 'bracket' is the most general operation, all other operations can be
```
adithyaov, line 428: Most general operation? Most general resource management operation?

adithyaov (commented):
Some comments.

core/src/Streamly/Internal/Data/Fold/Type.hs
harendra-kumar: This existing code, just moved in this PR. I won't change it in this PR.

core/src/Streamly/Internal/Data/Fold.hs
harendra-kumar: This was existing code, not changed in this PR.

adithyaov (approved):

PR #1819, opened 2022-09-01
Introduce a streamly-core flag in bench-runner to not build streamly

rnjtranjan:


rnjtranjan:
the issue is not reproducible.

PR #1820, opened 2022-09-02
Move bracket, finally, pollcount and after to streamly

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/pull/1820

core/src/Streamly/Internal/Data/Stream/Exception.hs
```diff
@@ -9,14 +9,7 @@
 module Streamly.Internal.Data.Stream.Exception
     (
       before
-    , after_
```
harendra-kumar, line 12: Do not remove anything from this module. We just have to change these to IO in the next PR. It is much better to just make small changes to these to adapt to Io rather than removing these first and then reintroducing again for IO variants.

core/src/Streamly/Internal/Data/Stream/StreamD/Exception.hs
```diff
@@ -8,16 +8,8 @@
 
 module Streamly.Internal.Data.Stream.StreamD.Exception
     (
-      gbracket_
```
harendra-kumar, line 11: Do not remove anything from this module, these just have to be adapted to IO monad in the next PR.

harendra-kumar (changes requested):

harendra-kumar:
See #2020 and #2021 .

PR #1821, opened 2022-09-02
IO version of bracket operation in streamly-core package

rnjtranjan:


harendra-kumar:
See #2020 and #2021 .

PR #1822, opened 2022-09-02
Remove configure build streamly-core

rnjtranjan:


rnjtranjan:
It seems configured build of streamly-core is required for test and docs packages.

harendra-kumar:
Configure is required for the Time module (`src/Streamly/Internal/Data/Time/Clock/config-clock.h`). Cannot be removed.

Issue #1802, opened 2022-08-22
Use a "fuse" modifier to use fused StreamD operations corresponding to unfused StreamK ones


harendra-kumar:
Basically, rewrite `fuse append` to `append2` (the StreamD version of append) and so on.

PR #1824, opened 2022-09-03
Use a "fuse" modifier to use fused StreamD operations corresponding t…

rnjtranjan:
…o unfused StreamK ones

### DIFF
--- a/core/src/Streamly/Internal/Data/Stream/Expand.hs
+++ b/core/src/Streamly/Internal/Data/Stream/Expand.hs
@@ -188,13 +188,11 @@ infixr 6 `append`
 -- /Pre-release/
 --
 {-# INLINE_NORMAL append #-}
-append :: Monad m => Stream m a -- ^ Monad contraint due to fuse re write RULES
-  -> Stream m a -- ^
-  -> Stream m a
+append :: Monad m => Stream m a -> Stream m a -> Stream m a
 append = (<>)
 
 {-# RULES
-   "fuseAppend" fuse append = append2
+   "fuse append" fuse append = append2
 #-}
 ------------------------------------------------------------------------------
 -- Interleaving
@@ -219,7 +217,7 @@ interleave2 :: Monad m => Stream m a -> Stream m a -> Stream m a
 interleave2 s1 s2 = fromStreamD $ D.interleave (toStreamD s1) (toStreamD s2)
 
 {-# RULES
-   "fuseInterleave" fuse interleave = interleave2
+   "fuse interleave" fuse interleave = interleave2
 #-}
 
 -- | Like `interleave` but stops interleaving as soon as the first stream
@@ -313,7 +311,7 @@ mergeByM2 f m1 m2 =
     fromStreamD $ D.mergeByM f (toStreamD m1) (toStreamD m2)
 
 {-# RULES
-   "fusemergeByM" fuse mergeByM = mergeByM2
+   "fuse mergeByM" fuse mergeByM = mergeByM2
 #-}
 -- | Like 'mergeByM' but stops merging as soon as any of the two streams stops.
 --
--- a/core/src/Streamly/Internal/Data/Stream/Expand.hs
+++ b/core/src/Streamly/Internal/Data/Stream/Expand.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE GADTs #-}
 -- |
 -- Module      : Streamly.Internal.Data.Stream.Expand
 -- Copyright   : (c) 2017 Composewell Technologies
@@ -37,6 +36,9 @@ module Streamly.Internal.Data.Stream.Expand
 
     -- ** Append
     , append2
+
+    -- ** Modifier for Stream fusion
+    , fuse
 
     -- ** Interleave
     , interleave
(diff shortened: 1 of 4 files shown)


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1802

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE GADTs #-}
```
harendra-kumar, line 1: Why do we need this?

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -186,10 +187,17 @@ infixr 6 `append`
 --
 -- /Pre-release/
 --
-{-# INLINE append #-}
-append :: Stream m a -> Stream m a -> Stream m a
+{-# INLINE[1] append #-}
```
harendra-kumar, line 190: Use `INLINE_NORMAL` instead.

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -186,10 +187,17 @@ infixr 6 `append`
...
+append :: Monad m => Stream m a -> Stream m a -> Stream m a
 append = (<>)
 
+{-# INLINE[1] fuse #-}
+fuse :: a -> a
+fuse  = id
```
harendra-kumar, line 196: How would someone use this, if this is not exported?

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -37,6 +37,9 @@ module Streamly.Internal.Data.Stream.Expand
     -- ** Append
     , append2
 
+    -- ** Modifier for Stream fusion
+    , fuse
```
harendra-kumar, line 41: This is a common operation, should not be in the Expand module. Can go in the Stream/Type module.

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -298,6 +312,9 @@ mergeByM2
 mergeByM2 f m1 m2 =
     fromStreamD $ D.mergeByM f (toStreamD m1) (toStreamD m2)
 
+{-# RULES
+   "fusemergeByM" fuse mergeByM = mergeByM2
```
harendra-kumar, line 316: "fuse mergeByM".

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -204,10 +210,18 @@ append = (<>)
...
+{-# INLINE_NORMAL interleave #-}
+interleave :: Monad m => Stream m a -> Stream m a -> Stream m a
 interleave s1 s2 = fromStreamK $ K.interleave (toStreamK s1) (toStreamK s2)
 
+{-# INLINE interleave2 #-}
+interleave2 :: Monad m => Stream m a -> Stream m a -> Stream m a
+interleave2 s1 s2 = fromStreamD $ D.interleave (toStreamD s1) (toStreamD s2)
+
+{-# RULES
+   "fuseInterleave" fuse interleave = interleave2
```
harendra-kumar, line 222: "fuse interleave"

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -186,10 +187,15 @@ infixr 6 `append`
...
-{-# INLINE append #-}
-append :: Stream m a -> Stream m a -> Stream m a
+{-# INLINE_NORMAL append #-}
+append :: Monad m => Stream m a -- ^ Monad contraint due to fuse re write RULES
+  -> Stream m a -- ^
+  -> Stream m a
 append = (<>)
 
+{-# RULES
+   "fuseAppend" fuse append = append2
```
harendra-kumar, line 197: "fuse append"

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -186,10 +187,15 @@ infixr 6 `append`
 --
 -- /Pre-release/
 --
-{-# INLINE append #-}
-append :: Stream m a -> Stream m a -> Stream m a
+{-# INLINE_NORMAL append #-}
+append :: Monad m => Stream m a -- ^ Monad contraint due to fuse re write RULES
```
harendra-kumar, line 191: This comment applies to "Stream m a", haddock will show it for that which would be wrong.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Stream/Type.hs
```diff
@@ -538,3 +541,7 @@ concatMapWith
     -> Stream m a
     -> Stream m b
 concatMapWith par f xs = bindWith par xs f
+
+{-# INLINE_NORMAL fuse #-}
+fuse :: a -> a
+fuse = id
```
harendra-kumar, line 547: We can use `error "Operation with "fuse" modifier could not be rewritten". So that we get to know any problems because of the operation not fusing or using the modifier on an operation which does not have a rewrite rule to fuse.

harendra-kumar (changes requested):

harendra-kumar:
We have now separated StreamD and StreamK, this is no longer needed.

Issue #1746, opened 2022-07-31
Create a Streamly.Data.Stream module in streamly-core


harendra-kumar:
Export only those functions from Streamly.Internal.Data.Stream that are also exposed earlier in Streamly.Prelude.

Also mark all these functions as `since 0.9.0`.

PR #1827, opened 2022-09-07
Since Annotations for Streamly.Data.Stream functions

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1746

PR #1838, opened 2022-09-16
Changes for 0.9.0 release

adithyaov:


benchmark/Streamly/Benchmark/Data/Array.hs
```diff
@@ -61,8 +63,8 @@ o_1_space_generation value =
     [ bgroup
         "generation"
         [ benchIOSrc "write . intFromTo" (sourceIntFromToFromStream value)
-        , let testStr = mkListString value
-           in testStr `deepseq` bench "read" (nf readInstance testStr)
+        -- , let testStr = mkListString value
+        --    in testStr `deepseq` bench "read" (nf readInstance testStr)
```
harendra-kumar, line 67: Uncomment this.

core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -306,30 +290,82 @@ strip p arr =
...
+            | otherwise =
+                let v1 = getIndexUnsafe a1 i
+                    v2 = getIndexUnsafe a2 i
+                 in case compare v1 v2 of
+                        LT -> LT
+                        GT -> GT
+                        EQ -> loop (i + 1)
+
+instance Show a => Show (Array a) where
+    show arr = show $ toList arr
```
harendra-kumar, line 368: Should use something like "fromList" in the show output. Should be consistent with the unboxed array show/read instances.

core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -175,50 +171,38 @@ fromList xs = unsafePerformIO $ fromStreamD $ D.fromList xs
...
-    {-# INLINE_LATE step #-}
-    step _ i
-        | i < 0 = return D.Stop
-    step _ (I# i) =
-        return $
-        case Exts.indexArray# (array# arr) i of
-            (# x #) -> D.Yield x (I# i - 1)
+toStreamDRev arr@Array{..} =
+    D.map (getIndexUnsafe arr)
+        $ D.enumerateFromThenToIntegral (arrLen - 1) (arrLen - 2) 0
```
harendra-kumar, line 202: We could implement it like `MArray.toStreamDRev . unsafeThaw`

harendra-kumar (approved):

core/src/Streamly/Internal/Data/Array.hs
adithyaov: Yes, We can move a lot of functions to the `Mut`. But I'm holding this off in this change.

PR #1852, opened 2022-09-21
Add concurrent interleave operations to Async module

harendra-kumar:


src/Streamly/Internal/Data/Stream/Async/Channel.hs
```diff
@@ -35,19 +34,25 @@ module Streamly.Internal.Data.Stream.Async.Channel
     -- * Stream operations
     , evalWithK
     , appendWithK
-    , concatMapWithK
+    , interleaveWithK
+    , concatMapWithK -- appendMap
+    , concatMapInterleaveWithK -- interleaveMap
```
adithyaov, line 39: Append-ish
- `withChannelK` (Rename this to withAppendWithK. Basically indicate append-ish)
- `appendWithK`
- `evalWithK` (Rename this to indicate append-ish)
- `concatMapWithK` -- appendMap

Interleave-ish
- `withInterleaveChannelK`
- `interleaveWithK`
- `concatMapInterleaveWithK` -- interleaveMap

adithyaov: Or is the default append-ish.

adithyaov, line 39: If you're exporting `withChannelK`, Also export `withInterleaveChannelK`? Keeping the corresponding interleave function will help make the distinction clearer and in a sense describe if it's a general API or async strategy specific..

adithyaov: Also, `evalWithInterleaveK`.

harendra-kumar: Yes the default is append, to keep the default name shorter. I considered naming explicitly for both, but was not sure.

harendra-kumar: append or interleave does not apply to eval as it is a single stream. In fact, eval should only exist in the Stream.Parallel module.

Issue #1834, opened 2022-09-13
Move Streamly.Unicode.* to streamly-core


harendra-kumar:
Only those modules are to be moved which do not depend on and will never depend on unicode-data.

PR #1853, opened 2022-09-22
Move Streamly.Unicode to streamly-core

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1834

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -1076,8 +1073,10 @@ stripHead = S.dropWhile isSpace
 --
 -- /Pre-release/
 {-# INLINE lines #-}
-lines :: (Monad m, IsStream t) => Fold m Char b -> t m Char -> t m b
-lines = S.splitOnSuffix (== '\n')
+lines :: (Monad m) => Fold m Char b -> Stream.Stream m Char -> Stream.Stream m b
+lines f m =
+    Stream.fromStreamD $
+    D.foldMany (Fold.takeEndBy_ (== '\n') f) (Stream.toStreamD m)
```
harendra-kumar, line 1079: Indent this line.

Instead of using D.foldMany we can use Stream.foldMany, that will avoid using StreamD.
Need to check the performance, if there is degradation we can use D.splitOnSuffix here.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -118,12 +116,11 @@ import qualified Streamly.Data.Fold as Fold
...
-import qualified Streamly.Internal.Data.Stream as Stream (fromByteStr#)
-import qualified Streamly.Internal.Data.Stream.Serial as Serial
-import qualified Streamly.Data.Array.Unboxed as Array
-import qualified Streamly.Internal.Data.Array.Unboxed.Type as A (Array (..))
-import qualified Streamly.Internal.Data.Stream.IsStream as S
+import qualified Streamly.Internal.Data.Stream as Stream
+import qualified Streamly.Internal.Data.Stream as Serial
+import qualified Streamly.Internal.Data.Array.Unboxed.Type as Array
+import qualified Streamly.Internal.Data.Array.Unboxed.Type as A
+import qualified Streamly.Internal.Data.Stream as S
```
harendra-kumar, line 123: Data.Stream is imported 3 times with different names. We can use just one name.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -118,12 +116,11 @@ import qualified Streamly.Data.Fold as Fold
...
-    (Parser(..), Step(..), Initial(..), toParserK, toFold)
-import qualified Streamly.Internal.Data.Stream as Stream (fromByteStr#)
-import qualified Streamly.Internal.Data.Stream.Serial as Serial
-import qualified Streamly.Data.Array.Unboxed as Array
-import qualified Streamly.Internal.Data.Array.Unboxed.Type as A (Array (..))
-import qualified Streamly.Internal.Data.Stream.IsStream as S
+import qualified Streamly.Internal.Data.Stream as Stream
+import qualified Streamly.Internal.Data.Stream as Serial
+import qualified Streamly.Internal.Data.Array.Unboxed.Type as Array
+import qualified Streamly.Internal.Data.Array.Unboxed.Type as A
```
harendra-kumar, line 122: We have imported it twice. Can we use just `Array`.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Array/Char.hs
```diff
@@ -18,9 +18,8 @@ module Streamly.Internal.Unicode.Array.Char
 where
 
 import Control.Monad.IO.Class (MonadIO)
-import Streamly.Data.Array.Unboxed (Array)
-import Streamly.Internal.Data.Stream.IsStream (IsStream)
-import Streamly.Prelude (MonadAsync)
+import Streamly.Data.Stream (Stream)
+import Streamly.Internal.Data.Array.Unboxed (Array)
```
adithyaov, line 22: Use the non-`Internal` module here.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -1103,12 +1097,12 @@ isSpace c
...
 --
--- > words = S.wordsBy isSpace
+-- > words = Stream.wordsBy isSpace
 --
 -- /Pre-release/
 {-# INLINE words #-}
-words :: (Monad m, IsStream t) => Fold m Char b -> t m Char -> t m b
-words = S.wordsBy isSpace
+words :: (Monad m) => Fold m Char b -> Stream m Char -> Stream m b
+words f m =  Stream.fromStreamD $ D.wordsBy isSpace f (Stream.toStreamD m)
```
adithyaov, line 1105: You should not use `StreamD` here. You can use `Stream.foldMany (Fold.takeEndBy_ isSpace)` instead.

harendra-kumar: Maybe a good idea to keep it as it is so that we do not have to worry about perf comparison and write the `foldMany` snippet in a comment. We can replace it later if performance is found to be equivalent.

core/src/Streamly/Internal/Unicode/String.hs
```diff
@@ -43,8 +43,7 @@ import Language.Haskell.TH.Quote
 
 import qualified Streamly.Data.Fold as Fold
 import qualified Streamly.Internal.Data.Parser as Parser
-    (some, many, takeWhile1)
```
adithyaov, line 46: Keep these imports.

core/streamly-core.cabal
```diff
@@ -326,6 +326,11 @@ library
                      , Streamly.Internal.Data.Array.Stream.Fold.Foreign
 
                      , Streamly.Internal.Data.Array.Stream.Foreign
+                    -- streamly-unicode
```
adithyaov, line 329:
```suggestion

                    -- streamly-unicode
```

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -749,15 +743,15 @@ decodeUtf8ArraysWithD cfm (D.Stream step state) =
         r <- step (adaptState gst) st
         return $
             case r of
-                Yield A.Array {..} s ->
+                Yield Array.Array {..} s ->
```
harendra-kumar, line 746: it is easier to read if we import the type/constructors unqualified.

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: Ideally StreamD should not lead into other modules. Stream is the aggregator. If required we can add `wordsBy` in Data.Stream and use this implementation there.

adithyaov: As a quick solution, we can just leave this as it is and make a note here I guess.

PR #1859, opened 2022-09-26
Add a simple test for the Unbox type class

adithyaov:


adithyaov:
See 
https://github.com/composewell/streamly/pull/1893

PR #1870, opened 2022-09-28
Explicit concurrency for Parallel stream type based on channels

harendra-kumar:


harendra-kumar:
Closing this in favor of the eagerEval option PR #1911 .

PR #1878, opened 2022-09-30
Expose Unicode.Char.Parser

rnjtranjan:


core/src/Streamly/Unicode/Char/Parser.hs
```diff
@@ -0,0 +1,55 @@
...
+    , lower
+    , upper
+    , mark
+    , print
+    , punctuation
+    , separator
+    , space
+    , symbol
+
+    -- digits
```
adithyaov, line 42: Make this a heading?

adithyaov:
@harendra-kumar 
1. Do we want to expose all the APIs?
2. We need to add more docs if we are exposing it. Maybe some examples.

harendra-kumar (changes requested):
As discussed, rename the module to Streamly.Unicode.Parser.

harendra-kumar (changes requested):
Change "print" to "printable".

core/src/Streamly/Unicode/Parser.hs
```diff
@@ -0,0 +1,55 @@
...
+    , separator
+    , space
+    , symbol
+
+    -- * Digits
+    , digit
+    , octDigit
+    , hexDigit
+
+    -- * Numeric
```
adithyaov, line 47: This should be `Numbers`

core/src/Streamly/Unicode/Parser.hs
```diff
@@ -0,0 +1,55 @@
...
+    , signed
+    , numeric
```
adithyaov, line 49: This should be under `Digit`

adithyaov (commented):
Please address the comments.

adithyaov:
Fixed the comments.

core/src/Streamly/Internal/Unicode/Char/Parser.hs
```diff
@@ -77,7 +76,7 @@ import qualified Streamly.Internal.Data.Parser as Parser
 
 -- XXX It may be possible to implement faster predicates for ASCII byte stream.
 -- We can measure if there is a signficant difference and if so we can add such
--- predicates to Streamly.Unicode.Char.Parser.Latin1.
+-- predicates to Streamly.Unicode.Parser.Latin1.
```
adithyaov, line 79: What is `Unicode.Parser.Latin1`? Where should `Latin1` stuff go in? @harendra-kumar

PR #1879, opened 2022-09-30
Expose Unicode.String

rnjtranjan:


core/src/Streamly/Unicode/String.hs
```diff
@@ -0,0 +1,16 @@
+-- |
+-- Module      : Streamly.Unicode.String
+-- Copyright   : (c) 2022 Composewell Technologies
+-- License     : Apache-2.0
```
adithyaov, line 4: BSD-3-Clause

adithyaov:
```suggestion
-- License     : BSD-3-Clause```

File: core/src/Streamly/Unicode/String.hs [hunk 2]
```diff
@@ -0,0 +1,16 @@
...
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
```
adithyaov, line 6:
```suggestion
-- Stability   : released
```

adithyaov (changes requested):
Few changes

adithyaov (approved):

Issue #1874, opened 2022-09-29
Expose Unicode.Char.Parser and Unicode.String modules


harendra-kumar:

rnjtranjan:
https://github.com/composewell/streamly/pull/1878
https://github.com/composewell/streamly/pull/1879

PR #1881, opened 2022-09-30
Move Streamly.Data.Array module to streamly-core

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1874

core/streamly-core.cabal
```diff
@@ -349,6 +349,7 @@ library
                      , Streamly.Data.Fold.Tee
                      , Streamly.Data.Parser
                      , Streamly.Data.Unfold
+                     , Streamly.Data.Array
```
adithyaov, line 352: This should be in `other-modules`. This should not be released.

PR #1887, opened 2022-10-03
Rename head to one

rnjtranjan:


core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -241,9 +241,6 @@ module Streamly.Internal.Data.Parser
     , retryMaxTotal
     , retryMaxSuccessive
     , retry
-
-     -- * Deprecated
-    , next
```
adithyaov, line 246: We can't remove this. This should still be there. We have deprecated it. 
We can remove this after 2 releases have passed.
When was this introduced?

adithyaov (commented):
Please address the comments.

Issue #1882, opened 2022-09-30
Parser/Fold combinator names


adithyaov:
WRT parsers and folds, you can think of,
1. Performing an action (`dropWhile`, `takeWhile`)
2. Getting a resulting object

In all cases, both 1 and 2 happen but the hirerchy depends on the context.

WRT, getting a resulting object, when naming the parsers, try to be as specific as possible. The object should be primary player.

Consider the action of getting the first element:

Few possible candidates discussed for this kind of action is `head`, `one`, and `next`.

`head` talks about the position of the element in the stream more than the element itself.

`next`, again has some positional connotation to it, and we are not talking about the element itself.

`one` on the other hand has more of an element-ish connotation to it, more than the position.

"Whatever this may be, get me one element"

Similarly, consider getting the element at the end of the stream.

Possible candidates are, `end`, and `last`.

`last`, similar to `head`, talks about the position of the element in the stream more than the element itself.

`end`, in a bit of gray area, although has some positional connotation to it, also has an element-ish connotation.

`end`, in this case, is lesser of the two evils.

Final words:
- Don't be too pedantic.
- Keep a balance between intuitiveness and convention.

PR #1889, opened 2022-10-03
Rename last to end  for Fold and Parser

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1882

rnjtranjan:
As discussed "last" is as good as "end".

PR #1890, opened 2022-10-03
Rename and move unboxed array APIs

adithyaov:
Added Unboxed.newUnpinnedBytes and Mut.Type.newUnpinnedArray

Array.Unboxed.Mut.Type.nilArrayContents -> Unboxed.nil Array.Unboxed.Mut.Type.newAlignedArrayContents -> Unboxed.newAlignedPinnedBytes

Mut.Type:
newArray -> newPinnedArray
newArrayAligned -> newAlignedPinnedArray

Unboxed:
newUnpinnedArrayContents -> newUnpinnedBytes

### DIFF
--- a/core/src/Streamly/Data/Array/Unboxed/Mut.hs
+++ b/core/src/Streamly/Data/Array/Unboxed/Mut.hs
@@ -19,7 +19,8 @@ module Streamly.Data.Array.Unboxed.Mut
     -- ** Construction
 
     -- Uninitialized Arrays
-    , newArray
+    , new
+    , newPinned
 
     -- From containers
     , fromListN
--- a/core/src/Streamly/Internal/Data/Array/Unboxed/Mut/Type.hs
+++ b/core/src/Streamly/Internal/Data/Array/Unboxed/Mut/Type.hs
@@ -18,7 +18,6 @@ module Streamly.Internal.Data.Array.Unboxed.Mut.Type
     -- $arrayNotes
       Array (..)
     , MutableByteArray
-    , nilArrayContents
     , touch
     , pin
     , unpin
@@ -28,11 +27,11 @@ module Streamly.Internal.Data.Array.Unboxed.Mut.Type
     , nil
 
     -- *** Uninitialized Arrays
-    , newArray
-    , newPinnedArrayBytes
-    , newArrayAligned
+    , newPinned
+    , newPinnedBytes
+    , newAlignedPinned
+    , new
     , newArrayWith
-    , newAlignedArrayContents
 
     -- *** Initialized Arrays
     , withNewArrayUnsafe
@@ -234,9 +233,11 @@ import Streamly.Internal.Data.Unboxed
     , touch
     )
 import GHC.Base
-    ( IO(..), byteArrayContents#
-    , Int(..), newAlignedPinnedByteArray#
-    , copyMutableByteArray#, compareByteArrays#
+    ( IO(..)
+    , Int(..)
+    , byteArrayContents#
+    , compareByteArrays#
+    , copyMutableByteArray#
     )
 import GHC.Base (noinline)
 import GHC.Exts (unsafeCoerce#)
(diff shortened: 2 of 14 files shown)


core/src/Streamly/Internal/Data/Array/Unboxed/Mut/Type.hs
```diff
@@ -385,6 +387,8 @@ unpin arr@Array{..} = do
 -- Alignment must be greater than or equal to machine word size and a power of
 -- 2.
 --
+-- Alignment is ignored if the allocator is allocation unpinned memory.
```
harendra-kumar, line 390: typo.

adithyaov:
`newArrayWith` can be renamed but it is pretty okay

core/src/Streamly/Data/Array/Unboxed/Mut.hs
```diff
@@ -19,7 +19,8 @@ module Streamly.Data.Array.Unboxed.Mut
     -- ** Construction
 
     -- Uninitialized Arrays
-    , newArray
+    , newPinned
+    , newUnpinned
```
harendra-kumar, line 23: This should probably be the default. So that without thinking about pinned/unpinned you always create an unpinned array unless you specifically require pinned. Usually pinning will be handled transparently by IO operations and not required for normal use cases.

core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -14,18 +14,21 @@ module Streamly.Internal.Data.Unboxed
     , getMutableByteArray#
     , pin
     , unpin
-    , newUnpinnedArrayContents
+    , newUnpinnedBytes
+    , newAlignedPinnedBytes
+    , nil
```
harendra-kumar, line 19: this module is looking more like MutableByteArray module.

core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -53,16 +56,30 @@ touch (MutableByteArray contents) =
...
 --------------------------------------------------------------------------------
 
-{-# INLINE newUnpinnedArrayContents #-}
-newUnpinnedArrayContents :: Int -> IO (MutableByteArray a)
-newUnpinnedArrayContents nbytes | nbytes < 0 =
-  errorWithoutStackTrace "newUnpinnedArrayContents: size must be >= 0"
-newUnpinnedArrayContents (I# nbytes) = IO $ \s ->
+{-# NOINLINE nil #-}
+nil :: MutableByteArray a
+nil = unsafePerformIO $ newAlignedPinnedBytes 0 0
```
harendra-kumar, line 61: Why is nil pinned? It does not have to be pinned.

core/src/Streamly/Internal/Data/Array/Unboxed/Mut/Type.hs
```diff
@@ -399,68 +401,63 @@ newArrayWith alloc alignSize count = do
...
+{-# INLINE newPinned #-}
+newPinned :: forall m a. (MonadIO m, Unboxed a) => Int -> m (Array a)
+newPinned = newAlignedPinned (alignment (undefined :: a))
+
+-- | Allocates an empty unpinned array that can hold 'count' items.  The memory
+-- of the array is uninitialized.
+--
+-- /Pre-release/
+{-# INLINE newUnpinned #-}
+newUnpinned :: forall m a. (MonadIO m, Unboxed a) => Int -> m (Array a)
```
harendra-kumar, line 455: Don't need the `forall` in this.

core/src/Streamly/Internal/Data/Unboxed.hs
adithyaov: I will rename it in another PR.

adithyaov: See https://github.com/composewell/streamly/issues/1910

PR #1884, opened 2022-10-03
 Remove CPP macros related to __GLASGOW_HASKLL

rnjtranjan:

PR #1892, opened 2022-10-04
Add a CI for Inspection test for __GLASGOW_HASKELL__ > 810

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1891

.github/workflows/haskell.yml
```diff
@@ -94,6 +94,15 @@ jobs:
             cabal_version: 3.6.2.0
             cabal_project: cabal.project
             ignore_error: false
+          - name: 9.4.2-fusion-inspection-Werror
```
harendra-kumar, line 97: Can you move this to line 81, next to the other 9.4.2 CI. Also, we can consider converting the existing 9.4.2 CI to fusion-plugin CI to avoid too many CIs.

.github/workflows/haskell.yml
```diff
@@ -94,6 +94,15 @@ jobs:
...
+            ghc_version: 9.4.2
+            runner: ubuntu-latest
+            build: cabal
+            cabal_version: 3.8.1.0
+            disable_sdist_build: "y"
+            cabal_project: cabal.project
```
harendra-kumar, line 103: The test title says Werror but it is not using the Werror cabal-project file. Also, we will have to fix the warnings for 9.4.2 to make the Werror CI work.

harendra-kumar: We do not need to fix the warnings in the IsStream/Prelude modules as those are going to be deprecated.

.github/workflows/haskell.yml
```diff
@@ -79,6 +79,15 @@ jobs:
...
             ignore_error: false
+          - name: 9.4.2-fusion-inspection
+            ghc_version: 9.4.2
+            runner: ubuntu-latest
+            build: cabal
+            cabal_version: 3.8.1.0
+            disable_sdist_build: "y"
+            cabal_project: cabal.project
+            cabal_build_options: "--flag fusion-plugin --flag inspection"
+            ignore_error: false
```
harendra-kumar, line 90: Let's remove the other 9.4.2 test, there are 12 CIs right now, should try to reduce.

harendra-kumar:
The CIs are failing because of 8.10.7 inspection test failing. Since we have added inspection for 9.4.2 we can remove the same for 8.10.7.

rnjtranjan:
https://github.com/composewell/streamly/pull/1884 takes care of this.

PR #1893, opened 2022-10-04
Add various instances for Unbox

adithyaov:


core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -219,6 +266,57 @@ DERIVE_UNBOXED( Double
...
+#if MIN_VERSION_base(4,14,0)
+DERIVE_WRAPPED_UNBOX(Unbox a =>,(Down a),Down)
+#endif
+DERIVE_WRAPPED_UNBOX(Unbox a =>,(Const a b),Const)
+DERIVE_BINARY_UNBOX(forall a. Unboxed a =>,(Complex a),(:+),a)
+DERIVE_BINARY_UNBOX(forall a. (Integral a, Unboxed a) =>,(Ratio a),(:%),a)
+DERIVE_BINARY_UNBOX(,Fingerprint,Fingerprint,Word64)
+
+-- XXX There was some problem with the Storable instance of (). Something
+-- related to the size being 0. Some division error?
```
harendra-kumar, line 301: We can remove this comment unless there is something specific to note. The problem in the past was about the size of () I guess. And we used a SIZE_OF macro everywhere to avoid that.

harendra-kumar (approved):

PR #1902, opened 2022-10-06
Some refactoring and doc changes to concurrent modules

harendra-kumar:


src/Streamly/Internal/Data/Stream/Async/Channel.hs
```diff
@@ -72,6 +72,8 @@ withChannelK evaluator modifier input = K.concatEffect action
         toChannelK chan (evaluator chan input)
         return $ fromChannelK chan
 
+-- XXX Dedup with withChannelK
```
adithyaov, line 75: Dedup == Deduplicate?

harendra-kumar: yes.

PR #1903, opened 2022-10-06
Add two queues (outer and inner) in Async streams

harendra-kumar:
To avoid the possibility of out of order execution of the outer loop blowing up the queue size to unbounded size.


src/Streamly/Internal/Data/Stream/Async/Channel/Type.hs
```diff
@@ -84,7 +84,7 @@ data Channel m a = Channel
     , remainingWork  :: Maybe (IORef Count)
     , yieldRateInfo  :: Maybe YieldRateInfo
 
-    , enqueue        :: (RunInIO m, Stream m a) -> IO ()
+    , enqueue        :: Bool -> (RunInIO m, Stream m a) -> IO ()
```
adithyaov, line 87: We are leaking the inner details of the queue here.

adithyaov (approved):

src/Streamly/Internal/Data/Stream/Async/Channel/Type.hs
harendra-kumar: That can be made part of the design of the queue having two components, inner and outer. We can think about this later.

PR #1911, opened 2022-10-06
Add an eagerEval option to Async stream

harendra-kumar:


src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,102 @@ pushWorker yieldMax sv = do
...
+    modThread = modifyThread (workerThreads sv) (outputDoorBell sv)
     exception = handleChildException (outputQueue sv) (outputDoorBell sv)
 
--- Returns:
--- True: can dispatch more
--- False: cannot dispatch any more
-dispatchWorker :: (MonadIO m, MonadBaseControl IO m) => Count -> Channel m a -> m Bool
-dispatchWorker yieldCount sv = do
+{-# INLINE getEffectiveWorkerLimit #-}
+getEffectiveWorkerLimit :: MonadIO m => Channel m a -> m Limit
```
adithyaov, line 83: The maximum number of workers required?
Can you explain a little about what this does?

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,102 @@ pushWorker yieldMax sv = do
...
+                            Limited lim -> Limited $ min lim (fromIntegral n)
+
+{-# INLINE checkMaxThreads #-}
+checkMaxThreads :: MonadIO m => Int -> Channel m a -> m Bool
+checkMaxThreads active sv = do
+    -- Note that we may deadlock if the previous workers (tasks in the
+    -- stream) wait/depend on the future workers (tasks in the stream)
+    -- executing. In that case we should either configure the maxWorker
+    -- count to higher or use parallel style instead of ahead or async
+    -- style.
```
adithyaov, line 105: Can you write about what `Config` options to use, instead of the word `parallel`.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,102 @@ pushWorker yieldMax sv = do
...
+        Limited lim -> do
+            (_, n) <- liftIO $ readIORef (outputQueue sv)
+            return $ fromIntegral lim > n + active
+
+dispatchWorker :: (MonadIO m, MonadBaseControl IO m) =>
+    Count -> Channel m a -> m Bool
+dispatchWorker yieldCount sv = do
     -- XXX in case of Ahead streams we should not send more than one worker
     -- when the work queue is done but heap is not done.
+    -- XXX Should we have a single abstraction for checking q and
```
adithyaov, line 132: q is work right?

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,102 @@ pushWorker yieldMax sv = do
...
+    case limit of
+        Unlimited -> return True
+        Limited lim -> do
+            (_, n) <- liftIO $ readIORef (outputQueue sv)
+            return $ fromIntegral lim > n + active
+
+dispatchWorker :: (MonadIO m, MonadBaseControl IO m) =>
+    Count -> Channel m a -> m Bool
+dispatchWorker yieldCount sv = do
     -- XXX in case of Ahead streams we should not send more than one worker
```
adithyaov, line 130: We need to update the internal dev docs.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -243,85 +280,93 @@ dispatchWorkerPaced sv = do
...
-        else do
-            liftIO
-                $ withDiagMVar
-                    (svarInspectMode sv)
-                    (dumpSVar sv)
-                    "sendWorkerWait: nothing to do"
-                $ takeMVar (outputDoorBell sv)
-            (_, len) <- liftIO $ readIORef (outputQueue sv)
-            when (len <= 0) $ sendWorkerWait delay dispatch sv
+sendWorkerWait eager delay dispatch sv = go
```
adithyaov, line 288: Had the indentation been the same, I could rely on the diff for review. The aesthetic change could have been a separate commit I guess.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -243,85 +280,93 @@ dispatchWorkerPaced sv = do
...
+
+    where
+
+    go = do
```
adithyaov, line 292: What was the point of this change? 
You've essentially pushed the function into a go-loop.
Any specific reason?

src/Streamly/Internal/Data/Stream/Async/Channel/Interleave.hs
```diff
@@ -198,6 +198,7 @@ getFifoSVar mrun cfg = do
             , workerThreads    = running
             , workLoop         = wloop q sv
             , enqueue          = \_ -> enqueueFIFO sv q
+            , eagerDispatch    = return ()
```
adithyaov, line 201: Eager dispatch is not applicable in interleave?

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
harendra-kumar: This refactor is a separate commit. Also, the diff tool can visually tell if there are any non-whitespace changes or not.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
harendra-kumar: this is static argument transformation, the commit message states the reason of the change. It is performance.

harendra-kumar: sendWorkerWait was a recursive function threading around all its arguments in the recursive calls. With this change no arguments are passed in the recursive call.

src/Streamly/Internal/Data/Stream/Async/Channel/Interleave.hs
harendra-kumar: Not implemented as of now.

src/Streamly/Internal/Data/Stream/Async/Channel.hs
```diff
@@ -190,9 +191,15 @@ mkEnqueue :: MonadAsync m =>
...
     return
-        $ let q stream =
+        $ let q stream = do
                 -- Enqueue the outer loop
                 liftIO $ enqueue chan False (runInIO, runner q stream)
+                -- XXX In case of eager dispatch we can just directly dispatch
+                -- a worker with the tail stream here rather than first queuing
+                -- and then dispatching a worker which dequeues the work. The
+                -- older implementation did a direct dispatch here and its perf
+                -- characterstics looked much better.
```
adithyaov, line 201: Can you elaborate a little more? Doesn't `dispatchWorker` always dequeue from the `Channel`? For directly dispatching a worker with a `Stream` we would need a new function right? That function would do what a `workLoop` does bust instead of dequing it would take the stream as a parameter?

adithyaov (commented):
I have not looked at the benchmarks thoroughly. I am able to understand the most part. Need some help gluing some implementation details.

src/Streamly/Internal/Data/Stream/Async/Channel.hs
harendra-kumar: Yes. Currently we dispatch a worker i.e. fork a new thread, the thread runs and tries to dequeue from the channel work queue, if it finds the queue empty then the thread exits.

Here we already have the work item that we are queuing, so instead of queuing it we can pass it to the worker we are forking. So we will need a variant of dispatchWorker that would fork the workloop with the stream argument.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,102 @@ pushWorker yieldMax sv = do
...
+            Unlimited -> True
+            -- Note that the use of remainingWork and workerCount is not
+            -- atomic and the counts may even have changed between reading
+            -- and using them here, so this is just approximate logic and
+            -- we cannot rely on it for correctness. We may actually
+            -- dispatch more workers than required.
+            Limited lim -> fromIntegral lim > active
+
+{-# INLINE checkMaxBuffer #-}
+checkMaxBuffer :: MonadIO m => Int -> Channel m a -> m Bool
```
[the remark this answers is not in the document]
harendra-kumar: This is an old comment, still kind of valid, because we do have parallel, ahead, async. We can change it in future.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
harendra-kumar: This is an old comment, still kind of valid, because we do have parallel, ahead, async. We can change it in future.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
harendra-kumar: In case of ahead we have a queue and a heap, both together are work. The queue may be empty but the heap may have some pending work to do.

src/Streamly/Internal/Data/Stream/Async/Channel/Dispatcher.hs
```diff
@@ -62,65 +62,108 @@ pushWorker yieldMax sv = do
...
+            Unlimited -> True
+            -- Note that the use of remainingWork and workerCount is not
+            -- atomic and the counts may even have changed between reading
+            -- and using them here, so this is just approximate logic and
+            -- we cannot rely on it for correctness. We may actually
+            -- dispatch more workers than required.
+            Limited lim -> fromIntegral lim > active
+
+-- | Determine whether we would exceed max buffer if we dispatch more workers
+-- based on the current outputQueue size and active workers..
```
adithyaov, line 122:
```suggestion
-- based on the current outputQueue size and active workers.
```

PR #1923, opened 2022-10-09
Implement an "ordered" option in concurrent streams

harendra-kumar:
This is the equivalent of the Ahead style streams.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/Async/Channel.hs
+++ b/src/Streamly/Internal/Data/Stream/Async/Channel.hs
@@ -25,6 +25,7 @@ module Streamly.Internal.Data.Stream.Async.Channel
     , inspectMode
     , StopWhen (..)
     , stopWhen
+    , ordered
 
     -- * Operations
     , fromChannel
--- a/src/Streamly/Internal/Data/Stream/Async/Channel.hs
+++ b/src/Streamly/Internal/Data/Stream/Async/Channel.hs
@@ -132,14 +132,14 @@ evalWithD modifier m = D.Stream step Nothing
 -- appending two streams
 -------------------------------------------------------------------------------
 
-{-# INLINE appendGeneric #-}
-appendGeneric :: MonadAsync m =>
+{-# INLINE _appendGeneric #-}
+_appendGeneric :: MonadAsync m =>
        ((Config -> Config) -> m (Channel m a))
     -> (Config -> Config)
     -> K.Stream m a
     -> K.Stream m a
     -> K.Stream m a
-appendGeneric newChan modifier stream1 stream2 = K.concatEffect action
+_appendGeneric newChan modifier stream1 stream2 = K.concatEffect action
 
     where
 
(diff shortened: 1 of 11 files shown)


src/Streamly/Internal/Data/Stream/Async/Channel/Operations.hs
```diff
@@ -281,6 +278,9 @@ _fromChannelD svar = D.Stream step FromSVarInit
     step _ (FromSVarLoop sv (ev : es)) = do
         case ev of
             ChildYield a -> return $ D.Yield a (FromSVarLoop sv es)
+            ChildStopChannel -> do
```
adithyaov, line 281: Does this mean that this function was initially incorrect? It did not account for the exception `ChildStop`?

src/Streamly/Internal/Data/Stream/Async/Channel.hs
```diff
@@ -163,7 +164,10 @@ appendGeneric newChan modifier stream1 stream2 = K.concatEffect action
 {-# INLINE appendWithK #-}
 appendWithK :: MonadAsync m =>
     (Config -> Config) -> K.Stream m a -> K.Stream m a -> K.Stream m a
-appendWithK = appendGeneric Append.newChannel
+appendWithK modifier stream1 stream2 =
+    if getOrdered (modifier defaultConfig)
+    then concatMapWithK modifier id (stream1 `K.cons` K.fromPure stream2)
+    else appendGeneric Append.newChannel modifier stream1 stream2
```
adithyaov, line 170: When the `ordered` is not set in the modifier are both the branches here same?

adithyaov: If so, is `appendGeneric` more performant? Hench the `if` statement exists?

src/Streamly/Internal/Data/Stream/Async/Channel/Append.hs
```diff
@@ -207,6 +213,727 @@ workLoopLIFOLimited qref sv winfo = run
...
+-- efficiently.
+--
+-- XXX Maybe we can start the ahead threads at a lower cpu and IO priority so
+-- that they do not hog the resources and hinder the progress of the threads in
+-- front of them.
+
+-- Left associated ahead expressions are expensive. We start a new SVar for
+-- each left associative expression. The queue is used only for right
+-- associated expression, we queue the right expression and execute the left.
+-- Thererefore the queue never has more than on item in it.
```
adithyaov, line 467: All this is also written above. This documentation is duplicated.

src/Streamly/Internal/Data/Stream/Async/Channel/Append.hs
```diff
@@ -274,17 +1011,32 @@ getLifoSVar mrun cfg = do
...
+                    else enqueueLIFO sv q inner
+            , eagerDispatch = when eager $ void $ dispatchWorker 0 sv
+            , isWorkDone =
+                if inOrder
+                then isWorkDoneAhead sv aheadQ outH
+                else workDone sv
+            , isQueueDone =
+                if inOrder
+                then isQueueDoneAhead sv aheadQ
+                else workDone sv
```
adithyaov, line 1032: I've only reviewed this part properly. I assume most of the code is copied from the other module. If not, let know.

adithyaov (commented):
Some comments.

src/Streamly/Internal/Data/Stream/Async/Channel/Operations.hs
harendra-kumar: Yes, should have handled the ChannelStop exception in the previous change.

src/Streamly/Internal/Data/Stream/Async/Channel.hs
harendra-kumar: Checked performance, it is the same. We can use concat itself for both cases. But in future if we want to use StreamD for fusion, the special append impl may be useful.

src/Streamly/Internal/Data/Stream/Async/Channel/Append.hs
harendra-kumar: Yes, will remove.

Issue #1924, opened 2022-10-09
Rename Data.Stream.Async to Data.Stream.Concurrent


harendra-kumar:

harendra-kumar:
And the corresponding tests and benchmarks as well.

PR #1926, opened 2022-10-11
Rename Data.Stream.Async to Data.Stream.Concurrent

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1924

Issue #1938, opened 2022-10-12
Implement timestamp* and timeIndex* operations in Data.Stream.Transform


harendra-kumar:
We just need to copy all these operations including any associated comments/docs from src/Streamly/Internal/Data/Stream/IsStream/Transform.hs to core/src/Streamly/Internal/Data/Stream/Transform.hs .

Remove the `MonadAsync` constraint after copying.

These are only internal, not to be exposed yet.

PR #1944, opened 2022-10-13
Implement timestamp* and timeIndex* operations

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1938

PR #1945, opened 2022-10-13
Expose Streamly.Data.Stream.Zip from streamly-core package

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1941

core/src/Streamly/Data/Stream/Zip.hs
```diff
@@ -0,0 +1,33 @@
...
+-- License     : BSD3
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+module Streamly.Data.Stream.Zip
+    (
+      ZipStream (..)
+    , ZipSerialM
+    , ZipSerial
```
harendra-kumar, line 14: Remove ZipSerial and ZipSerialM, these are for Streamly.Prelude.

core/src/Streamly/Data/Stream/Zip.hs
```diff
@@ -0,0 +1,33 @@
...
+    , zipWithM
+    , zipWith
```
harendra-kumar, line 16: Remove these two, these are already exported from Data.Stream. This module is only for the newtype wrapper for Applicative instance.

harendra-kumar (changes requested):

Issue #1942, opened 2022-10-12
Add zipWith to Data.Stream.Concurrent


harendra-kumar:
And possibly a Concurrent Zip Applicative type in Data.Stream.Zip.Concurrent module.

PR #1954, opened 2022-10-14
Add a Concurrent Zip Applicative  type  in Data.Stream.Zip.Concurrent module

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/1942

src/Streamly/Internal/Data/Stream/Zip/ConcurrentZip.hs
```diff
@@ -0,0 +1,70 @@
...
+type ConcurrentZip = ConcurrentZipM IO
+
+consMConcurrentZip :: Monad m =>
+    m a -> ConcurrentZipM m a -> ConcurrentZipM m a
+consMConcurrentZip m (ConcurrentZipM r) = ConcurrentZipM $ Stream.consM m r
+
+instance Monad m => Functor (ConcurrentZipM m) where
+    {-# INLINE fmap #-}
+    fmap f (ConcurrentZipM m) =
+        ConcurrentZipM $ fmap f m
```
harendra-kumar, line 64: We can derive the Functor instance.

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/Zip/ConcurrentZip.hs
```diff
@@ -0,0 +1,70 @@
+
+{-# LANGUAGE UndecidableInstances #-}
```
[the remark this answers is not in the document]
rnjtranjan: The constraint ‘Control.Monad.Trans.Control.MonadBaseControl
                    IO m’
    is no smaller than
    the instance head ‘Applicative (ZipConcurrent m)’
  (Use UndecidableInstances to permit this)
• In the instance declaration for ‘Applicative (ZipConcurrent m)’

src/Streamly/Internal/Data/Stream/Zip/Concurrent.hs
```diff
@@ -0,0 +1,47 @@
...
+import qualified Streamly.Internal.Data.Stream as Stream (repeat)
+import Prelude hiding (map, repeat, zipWith)
+
+-- $setup
+-- >>> import qualified Streamly.Internal.Data.Stream as Stream
+-- >>> import qualified Streamly.Data.Fold as Fold
+--
+
+newtype ZipConcurrent m a = ZipConcurrent {getStream :: Stream m a}
+        deriving (Semigroup, Monoid, Functor)
```
harendra-kumar, line 32: Rather than using the Semigroup/Monoid of the Serial stream we may want to use concurrent versions of those. Remove these two instances.

src/Streamly/Internal/Data/Stream/Zip/Concurrent.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE UndecidableInstances #-}
```
harendra-kumar, line 1: Is this required?

harendra-kumar (changes requested):

adithyaov:
@harendra-kumar Can we deprecate `Streamly.Internal.Data.Stream.ZipAsync` in favor of this?
Also `Stream.Zip.Concurrent` vs `Stream.Concurrent.Zip`

adithyaov:
We shouldn't be adding this module.
We should be moving `Streamly.Internal.Data.Stream.ZipAsync` to `Streamly.InternalData.Stream.Zip.Concurrent` and change stuff accordingly instead.
That was we track what is being changed.

src/Streamly/Internal/Data/Stream/Zip/Concurrent.hs
```diff
@@ -0,0 +1,47 @@
...
+
+import qualified Streamly.Internal.Data.Stream as Stream (repeat)
+import Prelude hiding (map, repeat, zipWith)
+
+-- $setup
+-- >>> import qualified Streamly.Internal.Data.Stream as Stream
+-- >>> import qualified Streamly.Data.Fold as Fold
+--
+
+newtype ZipConcurrent m a = ZipConcurrent {getStream :: Stream m a}
```
harendra-kumar, line 31: Rename `getStream` to `getZipConcurrent`.

harendra-kumar (changes requested):

PR #1957, opened 2022-10-14
Add repeatM to Streamly.Data.Stream module

rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/1919

### DIFF
--- a/core/src/Streamly/Data/Stream.hs
+++ b/core/src/Streamly/Data/Stream.hs
@@ -119,7 +119,6 @@ module Streamly.Data.Stream
     , fromPure
     , fromEffect
     , repeat
-    , repeatM
     , replicate
 
     -- Note: Using enumeration functions e.g. 'Prelude.enumFromThen' turns out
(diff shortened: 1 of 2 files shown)


core/src/Streamly/Data/Stream.hs
```diff
@@ -119,6 +119,7 @@ module Streamly.Data.Stream
     , fromPure
     , fromEffect
     , repeat
+    , repeatM
```
adithyaov, line 122: We only want to add it to an `Internal` module. We don't want to expose it.

core/src/Streamly/Internal/Data/Stream/Generate.hs
```diff
@@ -178,6 +180,10 @@ unfoldrM step = fromStreamD . D.unfoldrM step
 repeat :: Monad m => a -> Stream m a
 repeat = fromStreamD . D.repeat
 
+{-# INLINE_NORMAL repeatM #-}
+repeatM :: Monad m => m a -> Stream m a
+repeatM = Stream.sequence . repeat
```
harendra-kumar, line 185: Add the documentation from Streamly.Prelude.

harendra-kumar (changes requested):

PR #1959, opened 2022-10-14
Implement interleaved config option for concurrent streams

harendra-kumar:


src/Streamly/Internal/Data/Stream/Concurrent.hs
```diff
@@ -42,35 +43,46 @@ module Streamly.Internal.Data.Stream.Concurrent
...
     , mapM
     , mapMWith
     , sequence
     , sequenceWith
+    -- Add repeatM/replicateM?
 
-    -- XXX experimental binary ops, move to parallel?
+    -- ** Combine two
+    -- | Use a channel for each pair.
+    -- combine/concur/conjoin
```
adithyaov, line 64: combine

src/Streamly/Internal/Data/Stream/Concurrent.hs
```diff
@@ -127,23 +139,14 @@ eval = evalWith id
 -------------------------------------------------------------------------------
 
 -- | Like 'append' but with a Config modifier.
-{-# INLINE appendWith #-}
-appendWith :: MonadAsync m =>
+{-# INLINE combineWith #-}
+combineWith :: MonadAsync m =>
```
adithyaov, line 143: You changed this but not `append`?

adithyaov: Oh ok. Nvm. `combineWith` is the primitive.

src/Streamly/Internal/Data/Stream/Concurrent.hs
adithyaov: On second thought, leave it as it is. Remove this comment.
`combineWith` is the primitive.

adithyaov (approved):

PR #1965, opened 2022-10-16
Use list based concat functions in place of binary ops

harendra-kumar:


src/Streamly/Internal/Data/Stream/Concurrent.hs
```diff
@@ -46,13 +46,13 @@ module Streamly.Internal.Data.Stream.Concurrent
     -- | Stream combinators using a concurrent channel
 
     -- ** Evaluate
-    -- | Evaluate a stream concurrently using a channel.
+    -- | Evaluates a stream concurrently using a channel.
```
adithyaov, line 49: I wonder if we can reduce export list doc duplication.

src/Streamly/Internal/Data/Stream/Concurrent.hs
```diff
@@ -62,22 +62,32 @@ module Streamly.Internal.Data.Stream.Concurrent
     -- ** Combine two
     -- | Use a channel for each pair.
     -- combine/concur/conjoin
+    , append2
+    , interleave2
+    , ahead2
+    , parallel2
+    , parallelFst2
+    , parallelMin2
+    , combineWith
```
adithyaov, line 71: We should encourage the use of combining many streams at once or explicitly use channels to combine streams. Even if they combine 2 streams they can always keep the channel around to send another stream to it later. 

So, first use `Config -> Config` to create an opaque channel.

```
newchannelWith :: (Config -> Config) -> Channel
```

And `combineWith` would then be etc.

```
combineWith :: Channel -> Stream m a -> Stream m a -> Stream m a
```

Or use `toChannel` and `fromChannel` accordingly.

adithyaov: Or should we just have `Channel` instead of `Config -> Config` everywhere?

adithyaov: Binary ops should always take in a `Channel`.

PR #1966, opened 2022-10-16
 Separate interleaved benchmarks in another module 

harendra-kumar:


adithyaov (approved):

benchmark/streamly-benchmarks.cabal
```diff
@@ -244,7 +244,7 @@ benchmark Data.Stream
     else
       ghc-options: +RTS -M2500M -RTS
 
-benchmark Data.Stream.Async
+benchmark Data.Stream.Concurrent
```
adithyaov, line 247: This is done by @rnjtranjan along with some more file renaming as well.
See: https://github.com/composewell/streamly/pull/1968

PR #1998, opened 2022-10-19
Change Handle, Stdio, File unfold APIs

harendra-kumar:
Use the new convention, deprecate the old ones.


adithyaov (commented):
Check for `readChunk` and `writeChunk` and rename it to `getChunk` and `putChunk` accordingly.

PR #2004, opened 2022-10-19
Change the unfold/toStream naming for arrays

harendra-kumar:


core/src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -84,9 +84,10 @@ module Streamly.Internal.Data.Array.Mut.Type
 
     -- * Eliminating and Reading
 
-    -- ** To streams
-    , read
-    -- , readRev
+    -- ** Unfolds
+    , reader
+    -- , readerRev
+    , producer -- experimental
```
adithyaov, line 90: There is a `toStreamD` and `toStreamK` below.
Any reason you did not rename those?

core/src/Streamly/Internal/Data/Array/Unboxed.hs
```diff
@@ -103,6 +108,10 @@ module Streamly.Internal.Data.Array.Unboxed
     -- ** Folding
     , streamFold
     , fold
+
+    -- * Deprecated
+    , A.toStream
+    , A.toStreamRev
```
adithyaov, line 114: Instead, would you like to move these to the released module? We can remove the deprecated header here and we will track deprecations more nicely.

core/src/Streamly/Internal/Data/Array/Unboxed/Mut/Type.hs
```diff
@@ -96,9 +96,9 @@ module Streamly.Internal.Data.Array.Unboxed.Mut.Type
     -- * Eliminating and Reading
 
     -- ** To streams
-    , read
-    , readRevWith
-    , readRev
+    , reader
+    , readerRevWith
+    , readerRev
```
adithyaov, line 101: There are multiple `toStream*` combinators below.

core/src/Streamly/Internal/Data/Array/Unboxed/Type.hs
```diff
@@ -367,19 +369,33 @@ toStreamKRev arr =
 -- | Convert an 'Array' into a stream.
 --
 -- /Pre-release/
+{-# INLINE_EARLY read #-}
+read :: (Monad m, Unboxed a) => Array a -> Stream m a
+read = Stream.fromStreamD . toStreamD
+
+-- | Same as 'read'
+--
+{-# DEPRECATED toStream "Please use 'read' instead." #-}
```
adithyaov, line 378: Remove deprecations from this module?
Move them to the released modules?

Or move these to the deprecated section in the export list?

adithyaov (commented):
Few comments.

adithyaov:
Please fix the CIs.

PR #2013, opened 2022-10-20
Rename unfold APIs in the FileSystem.Dir module

harendra-kumar:
Also add deprecated versions of renamed APIs in the FileSystem.File module.

### DIFF
--- a/src/Streamly/Internal/FileSystem/File.hs
+++ b/src/Streamly/Internal/FileSystem/File.hs
@@ -81,6 +81,16 @@ module Streamly.Internal.FileSystem.File
     -- , appendShared
     , appendArray
     , appendChunks
+
+    -- * Deprecated
+    , readWithBufferOf
+    , readChunksWithBufferOf
+    , readChunksFromToWith
+    , toBytes
+    , toChunks
+    , toChunksWithBufferOf
+    , writeWithBufferOf
+    , fromBytesWithBufferOf
     )
 where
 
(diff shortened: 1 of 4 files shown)


rnjtranjan (approved):

src/Streamly/Internal/FileSystem/File.hs
```diff
@@ -81,6 +81,16 @@ module Streamly.Internal.FileSystem.File
...
+
+    -- * Deprecated
+    , readWithBufferOf
+    , readChunksWithBufferOf
+    , readChunksFromToWith
+    , toBytes
+    , toChunks
+    , toChunksWithBufferOf
+    , writeWithBufferOf
+    , fromBytesWithBufferOf
```
adithyaov, line 93: These were not exposed previously?
I don't see any lines highlighted in red?

src/Streamly/Internal/FileSystem/File.hs
```diff
@@ -222,6 +232,12 @@ readChunksWith :: (MonadCatch m, MonadAsync m)
 readChunksWith size file =
     withFile file ReadMode (FH.readChunksWith size)
 
+{-# DEPRECATED toChunksWithBufferOf "Please use 'readChunksWith' instead" #-}
```
adithyaov, line 235: You've added and deprecated these?

adithyaov: Were these removed previously?

src/Streamly/Internal/FileSystem/File.hs
harendra-kumar: These were renamed earlier without deprecating. Though this is an internal module, this is used by several users so we should deprecate rather than directly removing. You can diff with 0.8.3 to check.

src/Streamly/Internal/FileSystem/File.hs
harendra-kumar: Yes. were renamed without deprecating the older ones.

adithyaov (approved):

PR #2020, opened 2022-10-25
Use IO based bracket in "streamly-core" and lifted in "streamly"

harendra-kumar:


core/src/Streamly/Internal/Data/Stream/Exception.hs
```diff
@@ -108,32 +104,30 @@ onException action xs = fromStreamD $ D.onException action $ toStreamD xs
...
 --
 -- /Inhibits stream fusion/
 --
--- @since 0.9.0
-{-# INLINE finally #-}
-finally :: (MonadIO m, MonadBaseControl IO m, MonadCatch m) =>
-    m b -> Stream m a -> Stream m a
-finally action xs = fromStreamD $ D.finally action $ toStreamD xs
+{-# INLINE finallyIO #-}
+finallyIO :: (MonadIO m, MonadCatch m) =>
```
adithyaov, line 121: This is a weird signature.

PR #2021, opened 2022-10-25
Move pollCounts to streamly, fix use-unliftio flag, move fs modules to streamly-core

harendra-kumar:
There are several changes bundled together, commit-by-commit review should work.


adithyaov (approved):
I went through it already. Skimmed through it now. Have not tested the build with the flag though. We can add a CI but we have too many CIs already.

PR #2022, opened 2022-10-25
Doc comments, module expose/hide, base bound changes

harendra-kumar:


core/src/Streamly/Internal/Unicode/Array/Char.hs
```diff
@@ -6,6 +6,11 @@
 -- Maintainer  : streamly@composewell.com
 -- Stability   : experimental
 -- Portability : GHC
+
+-- XXX Remove this or move this to Unicode.Utf32 making "Array Char" as a
```
adithyaov, line 10: `Array Char` can be an alias for `Utf32` perhaps.

src/Streamly/Internal/Unicode/Utf8.hs
```diff
@@ -5,7 +5,12 @@
 -- Maintainer  : streamly@composewell.com
 -- Stability   : experimental
 -- Portability : GHC
---
+
+-- XXX We can move this to stream-core/streamly-unicode-core, and provide an
+-- additional module in streamly-unicode for case conversions (because it
+-- depends on unicode-data). Or just keep all of it in streamly-unicode
+-- which will have a dependency on unicode-data.
```
adithyaov, line 12: We can just do `streamly-unicode` first and further split if required.

core/src/Streamly/Internal/Data/Either/Strict.hs
```diff
@@ -19,6 +19,8 @@
 --
 module Streamly.Internal.Data.Either.Strict
     ( Either' (..)
+    -- XXX Remove these and add lazyEither/strictEither to convert to and from
```
adithyaov, line 22: Why remove these? We need to check for perf. Converting to lazy `Either` and checking adds another level of if/else which might be less performant.

adithyaov: We can remove these accordingly.

core/streamly-core.cabal
```diff
@@ -355,6 +355,7 @@ library
                      , Streamly.Data.Fold.Tee
                      , Streamly.Data.Parser
                      , Streamly.Data.Unfold
+                     , Streamly.Data.Array
```
adithyaov, line 382: We are exposing this. We might need to review the API.

core/streamly-core.cabal
```diff
@@ -364,6 +363,8 @@ library
                      , Streamly.FileSystem.Handle
                      , Streamly.Console.Stdio
 
+    other-modules: Streamly.Data.Stream.Zip
```
adithyaov, line 391: We can also hide the `Fold.Tee` type, eventually.

core/src/Streamly/Internal/Data/Either/Strict.hs
harendra-kumar: > Why remove these? We need to check for perf. Converting to lazy `Either` and checking adds another level of if/else which might be less performant.

It should get simplified by the compiler. We can check that.

core/streamly-core.cabal
harendra-kumar: Yes, we will review. I had taken a look and it looked ok.

PR #2024, opened 2022-10-25
Rename operations in Data.Stream.Concurrent

harendra-kumar:


src/Streamly/Data/Stream/Concurrent.hs
```diff
@@ -9,8 +9,8 @@
 -- Suggested idioms:
 --
 -- >>> joinAsync = joinUsing id
--- >>> joinOrdered = joinUsing (ordered True)
--- >>> joinEager = joinUsing (eager True)
+-- >>> joinOrdered = joinUsing ordered
+-- >>> joinEager = joinUsing eager
```
adithyaov, line 13: This will need to be reverted.

src/Streamly/Data/Stream/Concurrent.hs
```diff
@@ -84,4 +84,40 @@ module Streamly.Data.Stream.Concurrent
...
+
+{-# INLINE concatUsing #-}
+concatUsing :: MonadAsync m =>
+    (Config -> Config) -> Stream m (Stream m a) -> Stream m a
+concatUsing = concatWith
+
+{-# INLINE concatMapUsing #-}
+concatMapUsing :: MonadAsync m =>
+    (Config -> Config) -> (a -> Stream m b) -> Stream m a -> Stream m b
+concatMapUsing = concatMapWith
```
adithyaov, line 123: You should change this in the `Internal` module itself. Don't define them here.

adithyaov (commented):
The CIs need to be fixed.
I don't see a major advantage with the name change. I like `Concur.concatMapWith` and `Stream.concatMapWith`.
I like the correspondence.

`concatMapUsing` brings up unnecessary questions like,
* How is it different from `concatMapWith`?
* If it's named differently, there must be some internal difference.
* Why is there no `Concur.concatMapWith`?

harendra-kumar:
We have moved to using the `par` prefix now.

PR #2026, opened 2022-10-25
Fixing minor release issues

adithyaov:
- Make streamly-core docspec independent of streamly
- Fix the example in Streamly.Network.Socket module

### DIFF
--- a/core/src/Streamly/Internal/Data/Stream/Expand.hs
+++ b/core/src/Streamly/Internal/Data/Stream/Expand.hs
@@ -627,9 +627,9 @@ concatPairsWith par f m =
 -- It can be used to traverse a tree structure.  For example, to list a
 -- directory tree:
 --
--- >> input = Stream.fromPure (Left ".")
--- >> f = either Dir.toEither (const Stream.nil)
--- >> ls = Stream.iterateMapWith Stream.append f input
+-- >>> input = Stream.fromPure (Left ".")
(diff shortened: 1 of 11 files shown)


core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -627,6 +626,7 @@ concatPairsWith par f m =
 -- It can be used to traverse a tree structure.  For example, to list a
 -- directory tree:
 --
+-- >> import qualified Streamly.Internal.FileSystem.Dir as Dir
```
harendra-kumar, line 629: Need to use `>>>` for docspec to work.

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -721,6 +721,7 @@ concatMapEitherWith = undefined
 --
 -- To traverse a directory tree:
 --
+-- >> import qualified Streamly.Internal.FileSystem.Dir as Dir
```
harendra-kumar, line 724: Use `>>>`.

core/src/Streamly/Internal/Data/Stream/Expand.hs
adithyaov: I know. But it won't as `Dir` is not in `streamly-core`. This change should be reverted once your PRs are merged.

PR #2029, opened 2022-10-25
Replace Unboxed with Unbox and remove the Unboxed type alias

adithyaov:


harendra-kumar (commented):
1. Should we suggest enabling QuantifiedConstraints extension when using the `Unbox` type so that another constraint is not required?
2. CIs need to be fixed.

PR #2037, opened 2022-10-26
Allow "Continue 0" case in parser extract

harendra-kumar:
It can happen if the first parser in an alternative does not receive sufficient input but the next parser can work even without input.


adithyaov (approved):
Looks like the logic is the same everywhere. No obvious errors.

PR #2066, opened 2022-11-01
Add streamly-core sdist CI

adithyaov:


.github/workflows/haskell.yml
```diff
@@ -235,5 +245,14 @@ jobs:
           export TOOLS_DIR=/opt
           # /usr/local/opt/curl/bin for macOS
           export PATH=$HOME/.local/bin:$HOME/.ghcup/bin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/opt/curl/bin
+          if test -n "$SUBDIR"
```
adithyaov, line 248: This looks a but hacky but works for the time being.

.github/workflows/haskell.yml
```diff
@@ -235,5 +245,14 @@ jobs:
...
+          then
+            mv "$PACKCHECK_LOCAL_PATH" "$SUBDIR/packcheck.sh"
+            # This is required as dist-newstyle will be created where
+            # cabal.project exists in the tree. It is considered the working
+            # directory.
+            echo "packages: streamly-core.cabal" > "$SUBDIR/cabal.project"
```
harendra-kumar, line 254: Why do we need a cabal.project for sdist build. ideally we should not use a cabal.project. We can remove the cabal.project in the parent dir instead if packacheck does not have a way to ignore it.

adithyaov: Ah, this is what I've done initially.

.github/workflows/haskell.yml
adithyaov: See https://github.com/composewell/streamly/issues/2317

.github/workflows/haskell.yml
```diff
@@ -103,6 +106,13 @@ jobs:
             cabal_version: 3.6.2.0
             cabal_project: cabal.project.streamly
             ignore_error: true
+          - name: 9.0.2-streamly-core-sdist
+            ghc_version: 9.0.2
+            runner: ubuntu-latest
+            build: cabal
+            cabal_version: 3.6.2.0
+            ignore_error: true
```
harendra-kumar, line 114: We should not ignore error on this build.

harendra-kumar (approved):

PR #2067, opened 2022-11-02
Update Data.Fold API and related changes

harendra-kumar:
Fold/Parser/Stream related:

* Update Fold docs, move containers ops to streamly pkg
* Move the container dependent code from streamly-core to streamly
* Rename some functions.
* Release some functions.

And some more changes:

* Update network module docs
* Add some notes about the stream/fold types


core/src/Streamly/Internal/Data/Stream/Bottom.hs
```diff
@@ -170,6 +170,7 @@ relTimesWith = fmap snd . timesWith
 -- Elimination - Running a Fold
 ------------------------------------------------------------------------------
 
+-- XXX Rename this to buildLazy instead? buildl sounds like left, not lazy
```
harendra-kumar, line 173: It is internal. Let it be whatever it is for now, we may never release it.

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -451,7 +452,7 @@ oneEq :: (Monad m, Eq a) => a -> Parser m a a
 oneEq x = satisfy (== x)
 
 -- Rename to "exclude".
-
+-- XXX No. oneNotEq is good.
```
harendra-kumar, line 455: other options are "onlyThis" and "notThis". We can leave this for now.

core/src/Streamly/Internal/Data/Stream/Bottom.hs
```diff
@@ -182,40 +170,66 @@ relTimesWith = fmap snd . timesWith
...
+--  sum1 <- Stream.build Fold.sum (Stream.enumerateFromTo 1 10)
+--  sum2 <- Stream.build sum1 (Stream.enumerateFromTo 11 20)
+--  Stream.fold sum2 (Stream.enumerateFromTo 21 30)
+-- :}
+-- 465
+--
+build :: Monad m => Fold m a b -> Stream m a -> m (Fold m a b)
+build f = fold (Fold.duplicate f)
+
+-- XXX The doc removed here might be useful.
```
harendra-kumar, line 207: It was a bit out of context for the 'fold' operation.

PR #2073, opened 2022-11-03
Modify examples in the markdown files

adithyaov:


adithyaov:
This turned out to be really ugly.
See https://github.com/composewell/streamly/pull/2085 for cleaner changes.

PR #2082, opened 2022-11-06
Add a chunked parser implementation

harendra-kumar:


core/src/Streamly/Internal/Data/Parser/Chunked/Type.hs
```diff
@@ -0,0 +1,356 @@
...
+           Int
+           -- (alt nesting level, alt used elem count). Nesting level is
+           -- increased whenever we enter an Alternative composition and
+           -- decreased when it is done. The used element count is a count of
+           -- elements consumed by the Alternative. If the Alternative fails we
+           -- need to backtrack by this amount.
+           --
+           -- The nesting level is used in parseDToK to optimize the case when
+           -- we are not in an alternative, in that case we do not need to
+           -- maintain the element count for backtracking.
```
rnjtranjan, line 126: documentation is not aligned with the argument. (seems old one).

PR #2138, opened 2022-11-24
Return remaining stream for ParseError

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2136

PR #2141, opened 2022-11-24
Rename getZipStream

rnjtranjan:


core/src/Streamly/Internal/Data/Stream/Zip.hs
```diff
@@ -55,10 +55,10 @@ import qualified Streamly.Internal.Data.Stream as Stream
 -- >>> s2 = Stream.ZipStream $ Stream.fromFoldable [3, 4]
 -- >>> s3 = Stream.ZipStream $ Stream.fromFoldable [5, 6]
 -- >>> s = (,,) <$> s1 <*> s2 <*> s3
--- >>> Stream.fold Fold.toList (Stream.getZipStream s)
+-- >>> Stream.fold Fold.toList (Stream.unZipStream s)
 -- [(1,3,5),(2,4,6)]
 --
-newtype ZipStream m a = ZipStream {getZipStream :: Stream m a}
+newtype ZipStream m a = ZipStream {unZipStream :: Stream m a}
```
harendra-kumar, line 61: We can add a `getZipStream = unZipStream` and export it from Streamly.Prelude for backward compatibility.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Stream/Zip.hs
```diff
@@ -89,3 +90,6 @@ instance Monad m => Applicative (ZipStream m) where
 
     {-# INLINE (<*>) #-}
     ZipStream m1 <*> ZipStream m2 = ZipStream $ Stream.zipWith id m1 m2
+
+getZipStream :: ZipStream m a -> Stream m a
+getZipStream = unZipStream
```
harendra-kumar, line 95: It was `getZipSerialM` - https://streamly.composewell.com/haddocks/streamly-0.8.2/Streamly-Internal-Data-Stream-Zip.html#t:ZipSerialM . But looks like that constructor was not exported because the internal stream was StreamK, so there is not point adding getZipStream. Let's remove it. You can revert this commit.

harendra-kumar (changes requested):

PR #2156, opened 2022-12-07
Rename toFold in Tee to unTee

rnjtranjan:


rnjtranjan:
Fix:Rename toFold in Tee to unTee

core/src/Streamly/Internal/Data/Fold/Tee.hs
```diff
@@ -24,7 +24,7 @@ import qualified Streamly.Internal.Data.Fold.Type as Fold
 -- instances.
 --
 newtype Tee m a b =
-    Tee { toFold :: Fold m a b }
+    Tee { unTee :: Fold m a b }
```
harendra-kumar, line 27: We can add a `toFold` function for backward compatibility and deprecate it.

harendra-kumar (approved):

Issue #2083, opened 2022-11-07
Review and release Data.Array.Generic.Mut


harendra-kumar:
Mutable arrays.

PR #2161, opened 2022-12-07
Release Data.Array.Generic.Mut

rnjtranjan:


core/src/Streamly/Data/Array/Generic/Mut.hs
```diff
@@ -0,0 +1,63 @@
...
+-- License     : BSD3-3-Clause
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+module Streamly.Data.Array.Generic.Mut
+(
+    -- * Type
+    -- $arrayNotes
+      Array (..)
```
harendra-kumar, line 13: Do not expose the constructor.

core/src/Streamly/Data/Array/Generic/Mut.hs
```diff
@@ -0,0 +1,63 @@
...
+
+    -- * Constructing and Writing
+    -- ** Construction
+    -- *** Uninitialized Arrays
+    , newArray
```
harendra-kumar, line 18: The names in this module should be consistent with names in Streamly.Data.Array/Streamly.Data.Array.Mut .

core/src/Streamly/Data/Array/Generic/Mut.hs
```diff
@@ -0,0 +1,63 @@
...
+
+    -- *** From streams
+    , writeNUnsafe
```
harendra-kumar, line 21: Do not expose everything from the internal module. Please see what is exposed from other array modules and expose only carefully reviewed APIs that are consistent across all array modules.

harendra-kumar (changes requested):

core/src/Streamly/Data/Array/Generic/Mut.hs
```diff
@@ -0,0 +1,43 @@
...
+    -- bounds in the array representation.
+
+    -- ** Appending elements
+    , snoc
+
+    -- * Eliminating
+    , toList
+
+    -- ** Random reads
+    , getIndex
```
harendra-kumar, line 38: Not consistent with the API in Data.Array.Mut.

harendra-kumar (changes requested):
Can we keep the haddock headings consistent with Data.Array.Mut?
Export reader.

rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2083

Issue #2160, opened 2022-12-07
Expose filesystem file/dir modules


harendra-kumar:
With a stability warning that the type FilePath may change to OsPath or some typed Path in future.

rnjtranjan:
https://github.com/composewell/streamly/issues/2160

PR #2162, opened 2022-12-07
Expose filesystem file module

rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2160

### DIFF
--- a/core/src/Streamly/FileSystem/File.hs
+++ b/core/src/Streamly/FileSystem/File.hs
@@ -1,0 +1,77 @@
+-- |
+-- Module      : Streamly.FileSystem.File
+-- Copyright   : (c) 2019 Composewell Technologies
+--
+-- License     : BSD3
+-- Maintainer  : streamly@composewell.com
+-- Stability   : pre-release
+-- Portability : GHC
+--
+-- Read and write streams and arrays to and from files specified by their paths
+-- in the file system. Unlike the handle based APIs which can have a read/write
+-- session consisting of multiple reads and writes to the handle, these APIs
+-- are one shot read or write APIs. These APIs open the file handle, perform
+-- the requested operation and close the handle. Thease are safer compared to
+-- the handle based APIs as there is no possibility of a file descriptor
+-- leakage.
+--
+-- > import qualified Streamly.FileSystem.File as File
+--
+module Streamly.FileSystem.File
+    (
+    -- * Streaming IO
+    -- | Stream data to or from a file or device sequentially.  When reading,
+    -- the stream is lazy and generated on-demand as the consumer consumes it.
+    -- Read IO requests to the IO device are performed in chunks limited to a
+    -- maximum size of 32KiB, this is referred to as @defaultChunkSize@ in the
+    -- documentation. One IO request may or may not read the full
+    -- chunk. If the whole stream is not consumed, it is possible that we may
+    -- read slightly more from the IO device than what the consumer needed.
+    -- Unless specified otherwise in the API, writes are collected into chunks
+    -- of @defaultChunkSize@ before they are written to the IO device.
+
+    -- Streaming APIs work for all kind of devices, seekable or non-seekable;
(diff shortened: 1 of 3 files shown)


core/src/Streamly/FileSystem/File.hs
```diff
@@ -0,0 +1,77 @@
...
+
+    -- ** Folds
+    , write
+    , writeWith
+    , writeChunks
+
+    -- ** Writing Streams
+    , fromBytes -- putBytes?
+    , fromBytesWith
+    , fromChunks
```
harendra-kumar, line 66: Do not expose these.

core/src/Streamly/FileSystem/File.hs
```diff
@@ -0,0 +1,77 @@
...
+    , read
+    , readChunksWith
+    , readChunks
+
+    -- ** Unfolds
+    , readerWith
+    , reader
+    , chunkReaderWith
+    , chunkReaderFromToWith
+    , chunkReader
```
harendra-kumar, line 53: Let's not expose the unfolds yet.

core/src/Streamly/FileSystem/File.hs
```diff
@@ -0,0 +1,77 @@
...
+    -- ** Writing Streams
+    , fromBytes -- putBytes?
+    , fromBytesWith
+    , fromChunks
+
+    -- ** Append To File
+    , append
+    , appendWith
+    , appendArray
+    , appendChunks
```
harendra-kumar, line 72: The naming should be `writeAppend` , same as in mutable arrays. Do not expose appendArray yet.

core/src/Streamly/FileSystem/File.hs
```diff
@@ -0,0 +1,77 @@
...
+
+    -- ** Unfolds
+    , readerWith
+    , reader
+    , chunkReaderWith
+    , chunkReaderFromToWith
+    , chunkReader
+
+    -- ** Write To File
+    , putChunk -- writeChunk?
```
harendra-kumar, line 56: Do not expose this yet.

harendra-kumar (changes requested):
Can we also use `FilePath` in the signatures instead of String?

Need to ensure consistency between the naming/signatures of FileSystem.Handle, Arrays and this module.

core/src/Streamly/FileSystem/File.hs
```diff
@@ -0,0 +1,61 @@
...
+
+    -- ** Folds
+    , write
+    , writeWith
+    , writeChunks
+
+    -- ** Append To File
+    , writeAppend
+    , writeAppendWith
+    , writeAppendChunks
```
harendra-kumar, line 56: These are not folds, do not expose these. These should be folds consistent with the `write` APIs.

harendra-kumar (changes requested):
* Remove the since annotations from the APIs.
* Make all constraints in the order `(MonadIO m, MonadCatch m)`.

core/src/Streamly/Internal/FileSystem/File.hs
```diff
@@ -476,34 +482,37 @@ write = writeWith defaultChunkSize
...
--- @since 0.7.0
-{-# INLINE appendChunks #-}
-appendChunks :: (MonadIO m, MonadCatch m)
+-- /Pre-release/
+--
+{-# INLINE writeAppendChunks #-}
+writeAppendChunks :: (MonadIO m, MonadCatch m)
     => FilePath -> Stream m (Array a) -> m ()
-appendChunks = fromChunksMode AppendMode
+writeAppendChunks = fromChunksMode AppendMode
```
harendra-kumar, line 490: We can undo these renamings, in case anyone is using these they won't get affected. We anyway have to introduce writeAppend etc as Fold type.

harendra-kumar (changes requested):

harendra-kumar (approved):

Issue #2159, opened 2022-12-07
Rename unicode internal modules to make them consistent with released modules


harendra-kumar:
```
Streamly.Internal.Unicode.Array.Char =>  Streamly.Internal.Unicode.Array
Streamly.Internal.Unicode.Char.Parser =>  Streamly.Internal.Unicode.Parser
```

PR #2164, opened 2022-12-07
Rename unicode internal modules

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2159

Issue #2154, opened 2022-12-06
Export the Tee type via Data.Fold


harendra-kumar:
It would help in removing one module import.

harendra-kumar:
We can deprecate the Fold.Tee module and keep exported only from the streamly package.

PR #2166, opened 2022-12-09
Export the Tee type via Data.Fold

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2154

core/src/Streamly/Data/Fold/Tee.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -Wno-deprecations #-}
```
harendra-kumar, line 1: This should not be needed.

core/src/Streamly/Data/Fold/Tee.hs
```diff
@@ -6,34 +7,8 @@
...
--- input to both the folds and combine the outputs using Monoid or Semigroup
--- instances of the output types:
---
--- >>> import Data.Monoid (Sum(..))
--- >>> t = Tee Fold.one <> Tee Fold.latest
--- >>> Stream.fold (toFold t) (fmap Sum $ Stream.enumerateFromTo 1.0 100.0)
--- Just (Sum {getSum = 101.0})
---
--- The 'Num', 'Floating', and 'Fractional' instances work in the same way.
---
```
harendra-kumar, line 35: We should note remove this.

core/src/Streamly/Internal/Data/Fold/Tee.hs
```diff
@@ -10,6 +10,7 @@
 -- 'Semigroup', 'Monoid', 'Num', 'Floating' and 'Fractional' instances.
 --
 module Streamly.Internal.Data.Fold.Tee
+{-# DEPRECATED "Please use \"Streamly.Internal.Data.Fold\" instead." #-}
```
harendra-kumar, line 13: Do not deprecate this. Just import this module in Streamly.Data.Fold.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold/Tee.hs
```diff
@@ -9,6 +9,31 @@
...
+-- input to both the folds and combine the outputs using Monoid or Semigroup
+-- instances of the output types:
+--
+-- >>> import Data.Monoid (Sum(..))
+-- >>> t = Tee Fold.one <> Tee Fold.latest
+-- >>> Stream.fold (unTee t) (fmap Sum $ Stream.enumerateFromTo 1.0 100.0)
+-- Just (Sum {getSum = 101.0})
+--
+-- The 'Num', 'Floating', and 'Fractional' instances work in the same way.
+--
```
harendra-kumar, line 36: Move the documentation to the Tee type so that it is visible to users.

harendra-kumar (changes requested):

harendra-kumar (approved):

PR #2167, opened 2022-12-09
Get streamly-0.7.3.1 compiling with ghc-9.2.5.

philderbeast:
Stackage has come out with an LTS version for `ghc-9.2.5`. There's not many changes needed to get `streamly-0.7.3` compiling with this version.


harendra-kumar:
Do you need to have 0.7.3.1 with newer GHC support uploaded to Hackage? Would it be possible to upgrade to version 0.8.x instead?

philderbeast:
> Do you need to have 0.7.3.1 with newer GHC support uploaded to Hackage? Would it be possible to upgrade to version 0.8.x instead?

Yes this would help me and might help others upgrading from ghc-8.10 to ghc-9.2 so that I can separate the upgrade of the compiler (and compiler bundled packages) from the upgrade of streamly.

ghc-9.2.5.stack.yaml
```diff
@@ -0,0 +1,19 @@
+resolver: lts-20.3
```
harendra-kumar, line 1: Please remove the stack.yaml and stack.yaml.lock files. These are anyway not shipped with the distribution.

philderbeast: Done. I was able to test `ghc-9.2.5` with `tack test --resolver=lts-20.3`.

adithyaov:
See https://hackage.haskell.org/package/streamly-0.7.3.1
The changes are squashed to v0.7.3.1 branch along with some metadata changes

Apologies for the delay.

PR #2170, opened 2022-12-11
Consistent siganture with Array.Mut

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2083

harendra-kumar (approved):

Issue #2177, opened 2022-12-14
Replace "String" with FilePath in Streamly.FileSystem.Dir


harendra-kumar:
For paths we should use FilePath.

PR #2178, opened 2022-12-15
 Replace "String" with FilePath in Streamly.FileSystem.Dir

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2177

PR #2181, opened 2022-12-17
A host of cleanup changes and releasing some internal APIs

harendra-kumar:


core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -172,8 +172,8 @@ module Streamly.Internal.Data.Fold
     , with
 
     -- ** Transforming the Monad
-    , hoist
-    , generally
+    , morphInner
```
adithyaov, line 177: Isn't hoist already a common name that developers are familiar with?

core/src/Streamly/Internal/Data/Parser/Chunked.hs
```diff
@@ -7,6 +7,18 @@
 -- Maintainer  : streamly@composewell.com
 -- Stability   : pre-release
 -- Portability : GHC
+--
+-- Differences from an item parser:
```
adithyaov, line 9: `element parser`?
`item parser` sounds weird.

core/src/Streamly/Unicode/Parser.hs
```diff
@@ -12,17 +12,13 @@
 
 module Streamly.Unicode.Parser
     (
-     -- * Generic
+     -- * Single Chars
+
+     -- Any char
```
adithyaov, line 17: Make these subheadings?

core/src/Streamly/Internal/Data/Stream/Bottom.hs
```diff
@@ -488,7 +488,7 @@ drop n m = fromStreamD $ D.drop n $ toStreamD m
 -- | Find all the indices where the element in the stream satisfies the given
 -- predicate.
 --
--- >>> findIndices p = Stream.fold (Fold.findIndices p)
+-- >>> findIndices p = Stream.scanMaybe (Fold.findIndices p)
```
adithyaov, line 487: Maybe you can also add an example doc spec here so we don't make these kinds of mistakes again.

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -584,6 +584,10 @@ concatSmapMWith combine f initial =
 -- >>> Stream.fold Fold.toList $ Stream.concatPairsWith combine generate s
 -- [1,2,5,7,9]
 --
+-- Note that if the stream length is not a power of 2, the binary tree composed
+-- by concatPairsWith would not be balanced, which may or may not be important
+-- depending on what you are trying to achieve.
+--
```
adithyaov, line 609: Is the `caution` below correct?
`the stream of streams must be finite`
In the example above, the stream of streams is not infinite.

core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -581,24 +581,24 @@ concatSmapMWith combine f initial =
...
 -- depending on what you are trying to achieve.
 --
 -- /Caution: the stream of streams must be finite/
 --
 -- /Pre-release/
 --
-{-# INLINE concatPairsWith #-}
-concatPairsWith ::
+{-# INLINE mergeMapWith #-}
+mergeMapWith ::
```
adithyaov, line 617: I'm unable to relate `mergeMapWith` with the function signature.
`concatPairsWith` is relatively more intuitive.
`mergeMapWith` does not make sense to me at all.

core/src/Streamly/Internal/Data/Unfold.hs
```diff
@@ -237,7 +237,7 @@ module Streamly.Internal.Data.Unfold
     , crossWithM
     , crossWith
     , cross
-    , apply
+    , crossApply
```
adithyaov, line 240: `cross` vs `nest`

core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -170,17 +170,64 @@ unpin arr@(MutableByteArray marr#) =
...
+-- instance Unbox Object where
+--     peekByteIndex arr i = do
+--         let p = castContents arr
+--         x0 <- peekByteIndex p i
+--         x1 <- peekByteIndex p (i + 8)
+--         return $ Object x0 x1
+--     pokeByteIndex arr i (Object x0 x1) = do
+--         let p = castContents arr
+--         pokeByteIndex p i x0
+--         pokeByteIndex p (i + 8)x1
```
adithyaov, line 225: space, `(i + 8) x1`

adithyaov: We are going to rename `castContents` to `castMutableByteArray` right?

core/src/Streamly/Data/Stream.hs
```diff
@@ -234,6 +234,7 @@ module Streamly.Data.Stream
     -- express all others.
     , fold -- XXX rename to run? We can have a Stream.run and Fold.run.
     -- XXX fold1 can be achieved using Monoids or Refolds.
+    -- XXX We can call this just "break" and parseBreak as "munch"
```
adithyaov, line 246: What? `foldBreak` and `parseBreak` sound great. The names should be similar. Folds and parsers are sisters.

core/src/Streamly/Data/Stream.hs
```diff
@@ -79,6 +79,15 @@
...
 -- repository for many more real world examples of stream programming.
 --
+-- == Performance Notes
+--
+-- Operations annotated as /CPS/ force the use of continuation passing style
+-- streams. Therefore, such operations are not subjected to stream fusion.
+-- However, for some of these operations you can find fusible alternatives in
+-- the internal modules, which are perfectly fine to use but you need to
+-- understand the implications especially the O(n^2) nature of those
+-- operations..
```
adithyaov, line 89: These are implementation details.
Is this a step towards explicit usage of cps ops?

core/src/Streamly/Internal/Data/Fold.hs
```diff
@@ -155,7 +155,7 @@ module Streamly.Internal.Data.Fold
 
     -- * Running A Fold
     , drive
-    , driveBreak
+    , breakStream
```
adithyaov, line 158: fold & foldBreak
vs
drive & breakStream

foldBreak vs breakStream

adithyaov: This seems inconsistent. Does `driveContinue` make sense?

core/src/Streamly/Internal/Data/Stream/Bottom.hs
```diff
@@ -174,33 +180,25 @@ relTimesWith = fmap snd . timesWith
...
--- :}
--- 465
+-- >>> foldAdd f = Stream.foldAddLazy f >=> Fold.reduce
+
+-- |
+-- >>> foldAdd = flip Fold.addStream
 --
-build :: Monad m => Fold m a b -> Stream m a -> m (Fold m a b)
-build f = fold (Fold.duplicate f)
+foldAdd :: Monad m => Fold m a b -> Stream m a -> m (Fold m a b)
```
adithyaov, line 196: Is this `flip addStream`?
Can we discuss these names? 
Names of all the builders and family?

core/src/Streamly/Internal/Data/Fold.hs
harendra-kumar: `hoist` is not intuitive. the word hoist is a synonym for lift, however it does not lift, it transforms the monad. The name does not give any idea about that. I do not know the origin of this usage but it sounds wrong. We have `liftInner` for lifting the inner monad and similarly morphInner for transforming the inner monad makes better sense, the name gives an idea about transformation nature of the operation.

core/src/Streamly/Internal/Data/Parser/Chunked.hs
harendra-kumar: This is internal, element/item sound the same to me, can make it element.

core/src/Streamly/Unicode/Parser.hs
harendra-kumar: Less nesting is better.

core/src/Streamly/Internal/Data/Stream/Expand.hs
harendra-kumar: Not sure I understand. it is consistent, the example is finite and caution is also finite.

core/src/Streamly/Internal/Data/Stream/Expand.hs
harendra-kumar: `concatMap` can also be called `appendMap` because it maps and then combines the streams using an `append` style operation, it basically combines the new stream with the previously aggregated result. Or we can also call it a leftAssociativeLinearFlatteningMap. The canonical example of concatMap is appending (or concatenation) of generated streams. it is a map followed by concatenation.

The flaw with the `concatPairs` name is that it does not give any idea about the `map` operation. We will have to make the name longer to add map to it. If we look at the operation, it is essentially an associativeTreeFlatteningMap. The canonical application of this operation is merging of streams (e.g. merge sort). it is a map followed by a tree merge. So I called it `mergeMap`. 

I searched if anyone else is using a similar name, and found that rxjs seems to be having the same name for a similar operation. Other more intuitive suggestions are welcome.

core/src/Streamly/Internal/Data/Unfold.hs
harendra-kumar: nest is a very general name, cross implies cross product which is exactly what it is.

core/src/Streamly/Internal/Data/Unboxed.hs
harendra-kumar: there is an issue about it.

core/src/Streamly/Data/Stream.hs
harendra-kumar: Implementation details that are not really transparent, they impact the users, so should be known to them.

core/src/Streamly/Internal/Data/Fold.hs
harendra-kumar: We want to specify what is being used to drive the fold e.g. stream or unfold. Depending on that the operation would be breakStream or breakUnfold, just like in stream module we have foldBreak or parseBreak. Also, driveBreak sounded a bit odd. Also, this is internal as of now so we need not worry too much about it.

core/src/Streamly/Internal/Data/Stream/Bottom.hs
harendra-kumar: Yes, and yes. `foldAdd` is not exposed. The main operation is `addStream`.

core/src/Streamly/Internal/Data/Stream/Expand.hs
adithyaov: Ah, my bad. I read finite as `infinite`.

PR #2182, opened 2022-12-17
Add a binary encoding module and a type class

harendra-kumar:


rnjtranjan (approved):

core/src/Streamly/Internal/Data/Binary.hs
```diff
@@ -0,0 +1,27 @@
...
+import Streamly.Internal.Data.Stream (Stream)
+import Streamly.Internal.Data.Parser (Parser)
+
+class ToBytes a where
+    -- | Convert a Haskell type to a byte stream.
+    toBytes :: a -> Stream m Word8
+
+class FromBytes a where
+    -- | Decode a byte stream to a Haskell type.
+    fromBytes :: Parser Word8 m a
```
rnjtranjan, line 27: It would be more obvious if fromBytes  ::  Stream m Word8 -> a signature.

rnjtranjan: We can have a single class named. class SerDe a where {
toBytes :: a -> Stream m Word8
fromBytes :: Stream m Word8 -> a
}
SerDe: SerializerDeserializer.

harendra-kumar: Most serialization libraries use separate type classes e.g. ToJSON/FromJSON, not sure what the reason is though, can there be an assymetric case where we need one but not the other?

The signature is decided by what is more composable from performance perspective. This is only tentative, we have to do some perf experiments.

PR #2188, opened 2022-12-25
Add genericPeekByteIndex and genericPokeByteIndex

harendra-kumar:
For building instances for the Unbox type class


core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -169,6 +212,18 @@ unpin arr@(MutableByteArray marr#) =
 -- size and alignment. We can possibly write tests for checking compatibility
 -- with Storable.
 
+-- The main goal of the Unbox type class is to be used in arrays. Invariants
+-- for the sizeOf value required for use in arrays::
```
adithyaov, line 216:
```suggestion
-- for the sizeOf value required for use in arrays:
```

adithyaov (approved):
Looks good, overall.
Did not go into specifics.

PR #2190, opened 2022-12-26
Refactor Stream.Top, move transformers, containers ops

harendra-kumar:
Keep only those operation that depend only on "base" package, move out operations that depend on transformers, containers to Stream.Extra.

Add crossWith for cross product operation

Rename, refactor, document set-like, join ops


core/src/Streamly/Internal/Data/Stream/Container.hs
```diff
@@ -0,0 +1,302 @@
+-- |
+-- Module      : Streamly.Internal.Data.Stream.Extra
```
rnjtranjan, line 2: module name should be Streamly.Internal.Data.Stream.Container.

rnjtranjan (approved):

PR #2196, opened 2022-12-28
Remove Storable constraint from Unbox typeclass

adithyaov:


core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -297,8 +261,12 @@ class Storable a => Unbox a where
     -- IMPORTANT: The implementation of this interface may not check the bounds
     -- of the array, the caller must not assume that.
     pokeByteIndex :: MutableByteArray a -> Int -> a -> IO ()
+    -- | Get the size.
+    sizeOf :: a -> Int
+    -- | Get the alignment.
+    alignment :: a -> Int
```
harendra-kumar, line 267: we can remove alignment for now, if required we can add it later.

adithyaov: Removing this would make the code rather unclean. We use alignment in unboxed pinned arrays.
We can probably ignore alignment as a whole I guess later if it does not serve any benefit.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Array/ArrayMacros.h
```diff
@@ -9,6 +9,7 @@
 -- XXX Check the core to see if max can be statically eliminated. llvm can
 -- eliminate the comparison, but not sure if GHC NCG can.
 #define SIZE_OF(a) max 1 (sizeOf (undefined :: a))
+#define ALIGNMENT(a) max 1 (alignment (undefined :: a))
```
harendra-kumar, line 12: We can remove this?

core/src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -443,7 +440,10 @@ newAlignedPinned =
 --
 {-# INLINE newPinned #-}
 newPinned :: forall m a. (MonadIO m, Unbox a) => Int -> m (Array a)
-newPinned = newAlignedPinned (alignment (undefined :: a))
+newPinned =
+    newArrayWith
```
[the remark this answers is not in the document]
adithyaov: 1. We already have a `newPinnedAligned`
2. I'm not making any changes to `newArrayWith` as of now as that would require some internal followup changes. Creating an issue instead.

core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -454,6 +457,9 @@ instance Unbox Bool where
         then pokeByteIndex (castContents arr) i (1 :: Int32)
         else pokeByteIndex (castContents arr) i (0 :: Int32)
 
+    {-# INLINE sizeOf #-}
+    sizeOf _ = 4
```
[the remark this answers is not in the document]
adithyaov: We need to check the generic deriving.

harendra-kumar (approved):

PR #2198, opened 2022-12-31
Add iterateUnfoldMany - a fused version of iterateConcatMap

harendra-kumar:


core/src/Streamly/Internal/Data/Stream/StreamD/Type.hs
```diff
@@ -1006,6 +1010,126 @@ iterateConcatMapDFS f stream = Stream step (stream, [])
...
+       Unfold m a a
+    -> Stream m a
+    -> Stream m a
+iterateUnfoldManyDFS (Unfold istep inject) (Stream ostep ost) =
+    Stream step (IterateUnfoldOuter ost)
+
+    where
+
+    {-# INLINE_LATE step #-}
+    step gst (IterateUnfoldOuter o) = do
```
adithyaov, line 1029: I wonder if you can match on only 1 constructor here.
You can give the same importance to the outer and inner states.
`IterateState currentState otherStates`

adithyaov: Why have 2 different type variables? The `inject` is from the same unfold right?

adithyaov (approved):

PR #2199, opened 2023-01-01
Add concurrent version of iterateConcatMap, rename

harendra-kumar:


core/src/Streamly/Internal/Data/Stream/Expand.hs
```diff
@@ -120,32 +120,31 @@ module Streamly.Internal.Data.Stream.Expand
...
 
-    , iterateConcatMapWith
-    , iterateMergeMapWith
+    , concatIterateWith
+    , mergeIterateWith
 
-    , iterateConcatMapDFS
-    , iterateConcatMapBFS
+    , concatIterateDfs
+    , concatIterateBfs
```
adithyaov, line 131: Why?

adithyaov: small case `s`.

src/Streamly/Internal/Data/Stream/Concurrent/Channel/Append.hs
```diff
@@ -228,26 +230,19 @@ workLoopLIFOLimited qref sv winfo = run
...
-    Channel m a -> IORef ([K.Stream m a], Int) -> K.Stream m a -> IO ()
-reEnqueueAhead sv q m = do
-    atomicModifyIORefCAS_ q $ \ case
-        ([], n) -> ([m], n)  -- DO NOT increment sequence
-        _ -> error "reEnqueueAhead: queue is not empty"
+    -- XXX The queue is LIFO. When parConcatIterate queues more than one items
+    -- to the queue it will perform a DFS style traversal. For BFS we will have
+    -- to use a FIFO data structure here. That would require another Config
+    -- option.
+    atomicModifyIORefCAS_ q $ \(xs, n) -> (snd m:xs, n)
```
adithyaov, line 245: I do not understand this. Can you explain this change?

adithyaov: This is `reEnqueueAhead` now?

src/Streamly/Internal/Data/Stream/Concurrent/Channel/Append.hs
```diff
@@ -276,8 +271,19 @@ dequeueAhead :: MonadIO m
 dequeueAhead q = liftIO $
     atomicModifyIORefCAS q $ \case
             ([], n) -> (([], n), Nothing)
-            (x : [], n) -> (([], n), Just (x, n))
-            _ -> error "more than one item on queue"
+            (x : xs, n) -> ((xs, n + 1), Just (x, n + 1))
+
+-- Dequeue only if the seq no matches the expected seq no.
```
adithyaov, line 276:
```suggestion
-- Dequeue only if the seq number matches the expected seq number.
```

src/Streamly/Internal/Data/Stream/Concurrent/Channel/Append.hs
```diff
@@ -785,47 +785,42 @@ processWithToken q heap sv winfo action sno = do
...
+                updateHeapSeq heap nextSeqNo
+                incrementYieldLimit (remainingWork sv)
+        yieldLimitOk <- liftIO $ decrementYieldLimit (remainingWork sv)
+        -- To avoid a race when another thread puts something
+        -- on the heap and goes away, the consumer will not get
+        -- a doorBell and we will not clear the heap before
+        -- executing the next action. If the consumer depends
+        -- on the output that is stuck in the heap then this
+        -- will result in a deadlock. So we always clear the
+        -- heap before executing the next action.
```
adithyaov, line 798: I might need to re-read about how `Ahead` is implemented.

adithyaov: Or you can explain it to me and reason with your changes.

src/Streamly/Internal/Data/Stream/Concurrent/Channel/Append.hs
harendra-kumar: 1. Now we can have more then one item in the queue, because iterate appends to the queue.
2. Because of that we cannot increment the sequence number on enqueue because sequence number was the number to be used for the enqueued item, now we have more than one. Instead we increment the sequence number on dequeue now.

core/src/Streamly/Internal/Data/Stream/Expand.hs
harendra-kumar: Mixing all caps abbreviations with camel case is problematic e.g. unfoldIterateBFSRev. See https://github.com/composewell/haskell-dev/blob/master/coding-style.rst#id8 .

adithyaov (approved):

PR #2224, opened 2023-01-12
Upgrade package versions in cabal.project.packdiff

rnjtranjan:


.github/workflows/regression-check.yml
```diff
@@ -19,7 +19,6 @@ jobs:
     env:
       CI_BENCHMARKS_WITH_CUTOFF: >-
         Data.Array
-        Data.Array.Unboxed
```
harendra-kumar, line 22: We need to make it `Data.Array.Generic`.

harendra-kumar (changes requested):

PR #2230, opened 2023-01-13
Fix benchmark error in Data.Array.Stream

rnjtranjan:
**parseBreak (recursive, small arrays):** hangs for ever.

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/Array/Stream.hs
+++ b/benchmark/Streamly/Benchmark/Data/Array/Stream.hs
@@ -284,18 +284,20 @@ main = do
 
     where
 
-    alloc value = do
-        let val = max value 1
-        small <- Stream.toList $ Stream.arraysOf 100 $ sourceUnfoldrM val 0
-        big <- Stream.toList $ Stream.arraysOf val $ sourceUnfoldrM val 0
-        return (small, big)
+    alloc value =
+        if value <= 0
+        then return  (undefined, undefined)
+        else
+            do
+            small <- Stream.toList $ Stream.arraysOf 100 $ sourceUnfoldrM value 0
+            big <- Stream.toList $ Stream.arraysOf value $ sourceUnfoldrM value 0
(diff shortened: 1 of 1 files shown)


benchmark/Streamly/Benchmark/Data/Array/Stream.hs
```diff
@@ -285,15 +285,17 @@ main = do
     where
 
     alloc value = do
-        small <- Stream.toList $ Stream.arraysOf 100 $ sourceUnfoldrM value 0
-        big <- Stream.toList $ Stream.arraysOf value $ sourceUnfoldrM value 0
+        let val = max value 1
+        small <- Stream.toList $ Stream.arraysOf 100 $ sourceUnfoldrM val 0
+        big <- Stream.toList $ Stream.arraysOf val $ sourceUnfoldrM val 0
```
harendra-kumar, line 290: Only `runWithCLIOptsEnv` calls this with value `0`. And it perhaps does not use the return value of alloc. So we can just try returning (undefined, undefined) if the value passed is 0.

benchmark/Streamly/Benchmark/Data/Array/Stream.hs
```diff
@@ -285,15 +285,17 @@ main = do
...
         return (small, big)
 
     allBenchmarks env arrays value =
         let (arraysSmall, arraysBig) = arrays
+            val = max value 1
         in [ bgroup (o_1_space_prefix moduleName) $ Prelude.concat
           [ o_1_space_read_chunked env
-          , o_1_space_serial_array value arraysSmall arraysBig
+          , o_1_space_serial_array val arraysSmall arraysBig
```
harendra-kumar, line 298: Changes in this function should not be required.

harendra-kumar (changes requested):

rnjtranjan:
benchmark: **parseBreak (recursive, small arrays)** is hanging

Issue #2105, opened 2022-11-09
Improve the module deprecation messages


adithyaov:
- Streamly.Internal.Data.Stream.IsStream.*
- Streamly.Prelude
- Streamly.Internal.Data.Stream.Serial
- Streamly.Internal.Data.Stream.Async
- Streamly.Internal.Data.Stream.Parallel

harendra-kumar:
""Please use \"Streamly.Data.Stream\", \"Streamly.Data.Stream.Concurrent\", & \"Streamly.Data.Stream.Time\" instead."

Suggest Corresponding Internal modules, these modules are not exposed. Also add the Exception.Lifted module.

harendra-kumar:
Also say from which package (streamly or streamly-core) these modules are to be used.

PR #2235, opened 2023-01-16
Improve Deprecation message

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2105

src/Streamly/Prelude.hs
```diff
@@ -318,7 +318,7 @@
 -- performance reasons.
 
 module Streamly.Prelude
-{-# DEPRECATED "Please use \"Streamly.Data.Stream.Prelude\" instead." #-}
+{-# DEPRECATED "Please use \"Streamly.Internal.Data.Stream\", \"Streamly.Internal.Data.Stream.Concurrent\", \"Streamly.Internal.Data.Stream.Time\" & \"Streamly.Internal.Data.Stream.Exception\" from streamly-core package instead." #-}
```
harendra-kumar, line 321: This is a released module so here we should just suggest released modules `Streamly.Data.Stream.Prelude` and `Streamly.Data.Fold` as we have moved the folds to the Fold module.

harendra-kumar (changes requested):

harendra-kumar:
The issue mentions these modules as well:

Streamly.Internal.Data.Stream.Async
Streamly.Internal.Data.Stream.Parallel

PR #2239, opened 2023-01-18
Update packdiff and process version in cabal.project.*

rnjtranjan:


harendra-kumar:
Some benchmark is exceeding the heap limit. You can check which one by running manually. And change the limit in 

```
Data.Stream.ConcurrentEager: Heap exhausted;
Data.Stream.ConcurrentEager: Current maximum heap size is 1073741824 bytes (1024 MB).
Data.Stream.ConcurrentEager: Use `+RTS -M<size>' to increase it.
Error: Benchmark execution failed.
```

This likely to be Data.Stream.ConcurrentEager/o-1-space.monad-outer-product.toNullAp benchmark because that is the one with 1024MB memory limit, we can change it to 2048 in benchmark/bench-runner/Main.hs .

PR #2241, opened 2023-01-18
Add Haskell lazy evaluation doc module

harendra-kumar:


docs/User/Tutorials/haskell-lazy-evaluation.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+Hard pressed, Haskell now has no choice but to start evaluating the
+expression but we know it can never finish because the list is infinite.
+
+All expressions in Haskell are unevaluated by default.  Haskell
+evaluates the expressions only when it is absolutely required e.g. when
+we have to print something, or write it to a file or to network.
+
+## On-demand Evaluation
```
adithyaov, line 53: The title does not seem to justify the content. We are talking about memory.

docs/User/Tutorials/haskell-lazy-evaluation.md
```diff
@@ -0,0 +1,171 @@
...
+    let largeList = [1..1000000::Int]
+        x1 = last largeList
+        x2 = head largeList
+     in print (x1 `seq` x2)
+```
+
+Similar to the example in previous section, this also retains the entire list
+in memory because `x1` is forced to be evaluated before `x2`.
+
+## Reusing without retaining
```
adithyaov, line 132: The contents of this seem a little out of place for some reason.

docs/User/Tutorials/haskell-lazy-evaluation.md
harendra-kumar: I changed the content a little bit.

docs/User/Tutorials/haskell-lazy-evaluation.md
harendra-kumar: This tightly related to the previous section, just showing how the problem discussed there can be avoided.

rnjtranjan (approved):

PR #2265, opened 2023-02-15
DirStream implementation

rnjtranjan:


core/src/Streamly/Internal/FileSystem/Dir.hs
```diff
@@ -243,6 +246,27 @@ reader =
       UF.filter (\x -> x /= "." && x /= "..")
     $ UF.lmapM (liftIO . Dir.getDirectoryContents) UF.fromList
 
+{-# INLINABLE childOf #-}
+childOf :: FilePath -> IO [FilePath]
```
harendra-kumar, line 250: This will accumulate the entire dir into a list and then that list is unfolded by streamReader. 

We need to write this as an Unfold rather than making a list first.

PR #2268, opened 2023-02-16
Fix genericSizeOf

rnjtranjan:


core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -687,7 +687,9 @@ instance (MaxArity256 (SumArity (f :+: g)), SizeOfRepSum f, SizeOfRepSum g) =>
 
 {-# INLINE genericSizeOf #-}
 genericSizeOf :: forall a. (SizeOfRep (Rep a)) => Proxy a -> Int
-genericSizeOf _ = sizeOfRep (undefined :: Rep a x)
+genericSizeOf _ =
+    let s = sizeOfRep (undefined :: Rep a x)
+      in if s == 0 then 1 else s
```
harendra-kumar, line 692: Need to fix:

```
instance SizeOfRep U1 where
    {-# INLINE sizeOfRep #-}
sizeOfRep _ = 1
```

to return 0, for a complete fix.

PR #2278, opened 2023-02-25
Implement double Parser

rnjtranjan:


rnjtranjan:
IS.parse double (IS.fromList "3e")    == 30 is not working
ghci> S.parse P.double (S.fromList "3100e")
*** Exception: Assertion failed
CallStack (from HasCallStack):
  assert, called at src/Streamly/Internal/Data/Parser/ParserD/Type.hs:922:17 in streamly-core-0.1.0-inplace:Streamly.Internal.Data.Parser.ParserD.Type

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -294,5 +295,29 @@ signed p = (negate <$> (char '-' *> p)) <|> (char '+' *> p) <|> p
 -- \"Infinity\".
 --
 -- /Unimplemented/
-double :: Parser Char m Double
-double = undefined
+-- A strict coeff and exp pair
+data CE = CE !Integer {-# UNPACK #-}!Int
+
+double :: (Monad m) => Parser Char m Double
+double = do
```
harendra-kumar, line 302: Can you write a benchmark for it? I suspect the performance will be poor because it is using Monad. We should write it directly by defining a step function.

harendra-kumar: Also, enable the doctests in the documentation to check if they pass.

rnjtranjan:
Unicode.Parser/o-n-heap
    double:  OK (1.44s)
      479  ms ± 7.4 ms
    double2: OK (1.35s)
      449  ms ± 8.0 ms

It seems direct implementation and monadic have similar performance.

harendra-kumar:
> It seems direct implementation and monadic have similar performance.

Try adding an INLINE on double2.

hie.yaml
```diff
@@ -88,6 +88,8 @@ cradle:
               component: "lib:streamly-benchmarks"
             - path: "./benchmark/Streamly/Benchmark/Data/Fold/Window.hs"
               component: "bench:Data.Fold.Window"
+            - path: ".benchmark/Streamly/Benchmark/Unicode/Parser.hs"
```
harendra-kumar, line 91: Fix the path.

harendra-kumar (changes requested):

rnjtranjan:
after INLINE:
All
  Unicode.Parser/o-n-heap
    double:  OK (1.38s)
      457  ms ±  20 ms
    double2: OK (1.34s)
      446  ms ± 2.8 ms

harendra-kumar:
The benchmark was parsing a single number of 100,000 digits, which is not a realistic use case. Used 100,000 numbers of few digits instead.

```
Unicode.Parser(cpuTime)
Benchmark                           default(ms)
----------------------------------- -----------
All.Unicode.Parser/o-n-heap.double       151.71
All.Unicode.Parser/o-n-heap.double2       92.91

Unicode.Parser(Allocated)
Benchmark                           default(MiB)
----------------------------------- ------------
All.Unicode.Parser/o-n-heap.double        513.56
All.Unicode.Parser/o-n-heap.double2       129.98
```

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -294,5 +297,83 @@ signed p = (negate <$> (char '-' *> p)) <|> (char '+' *> p) <|> p
...
+            then fromIntegral (sc * 10 ^ ex)
+            else fromRational (sc % 10 ^ (-ex))
+    return $ h signedCoeff expo
+
+data TE =
+    TE  {-# UNPACK #-}!Int
+        {-# UNPACK #-}!Int
+        {-# UNPACK #-}!Int
+        !Integer
+        {-# UNPACK #-}!Int
```
harendra-kumar, line 333: Can you write comments about each field of the state, what they mean?

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -294,5 +297,83 @@ signed p = (negate <$> (char '-' *> p)) <|> (char '+' *> p) <|> p
...
+
+    initial = return $ IPartial (TE 0 0 0 0 0)
+
+    step (TE c p s m e) a = do
+        let stp = c + 1
+        return $
+            case a of
+                '-' -> Partial 0 (TE stp 0 (-1) m e)
+                '+' -> Partial 0 (TE stp 0 0 m e)
+                '.' -> Partial 0 (TE stp 1 s m e)
```
harendra-kumar, line 349: Can these occur anywhere in the number? Can you write some common test cases to make sure this works as expected for common cases?

harendra-kumar (changes requested):
we can remove the Monad based implementation and keep the other implementation.

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -263,36 +265,127 @@ hexadecimal = Parser.takeWhile1 isHexDigit (Fold.foldl' step 0)
...
+--
+-- >>> doubleParser "-."
+-- Right 0.0
+--
+-- >>> doubleParser "4."
+-- Right 4.0
+--
+-- >>> doubleParser "a."
+-- Left (ParseError "Invalid double fromat")
+--
```
harendra-kumar, line 346: Many of the above tests do not look right. Can we make sure that the results are compliant with the double parser in attoparsec https://hackage.haskell.org/package/attoparsec-0.14.4/docs/Data-Attoparsec-ByteString-Char8.html#g:14 ?

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -266,42 +266,58 @@ signed :: (Num a, Monad m) => Parser Char m a -> Parser Char m a
...
 data TE =
-    TE  {-# UNPACK #-}!Int      -- ^ Total number of digits
-        {-# UNPACK #-}!Int      -- ^ Decimal point exsistence
+    TE  {-# UNPACK #-}!Bool     -- ^ Found digits
+        {-# UNPACK #-}!Bool     -- ^ Decimal point exsistence
         {-# UNPACK #-}!Int      -- ^ The sign positive or negative
         !Integer                -- ^ The coefficient of a scientific number.
         {-# UNPACK #-}!Int      -- ^ The base-10 exponent of a scientific number.
 
 -- | Parse a 'Double'.
```
harendra-kumar, line 275: Parse a `decimal` 'Double'.

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -266,42 +266,58 @@ signed :: (Num a, Monad m) => Parser Char m a -> Parser Char m a
...
 -- This parser accepts an optional sign (+ or -) followed by
--- at least one decimal digit.  The syntax is similar to that accepted by
+-- at most one decimal digit.  The syntax is similar to that accepted by
 -- the 'read' function, with the exception that a trailing @\'.\'@ is
 -- consumed.
```
harendra-kumar, line 279: Let's not consume the trailing .  unless we know the reason why.

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -266,42 +266,58 @@ signed :: (Num a, Monad m) => Parser Char m a -> Parser Char m a
...
--- >>> doubleParser = S.parse double . S.fromList
+-- >>> doubleParser = Stream.parse Unicode.double . Stream.fromList
 --
 -- >>> doubleParser "3"
 -- Right 3.0
 --
--- >>> doubleParser ".3"
--- Right 0.3
+-- >>> doubleParser "4."
+-- Right 4.0
```
harendra-kumar, line 292: We should return the `.` to the input.

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -323,69 +339,65 @@ data TE =
...
 -- >>> doubleParser "-."
--- Right 0.0
---
--- >>> doubleParser "4."
--- Right 4.0
+-- Left (ParseError "Not enough input")
 --
 -- >>> doubleParser "a."
--- Left (ParseError "Invalid double fromat")
+-- Left (ParseError "Not enough input")
```
harendra-kumar, line 370: Remove the duplicate examples. 

Keep the documentation shorter. Move some less important examples and the not-enough input examples to a test module.

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -266,42 +266,58 @@ signed :: (Num a, Monad m) => Parser Char m a -> Parser Char m a
 signed p = (negate <$> (char '-' *> p)) <|> (char '+' *> p) <|> p
 
 data TE =
```
harendra-kumar, line 268: Rename `TE` to `DoubleState` or so.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -79,8 +81,8 @@ import qualified Streamly.Internal.Data.Parser as Parser
 --
 #define CHAR_PARSER_SIG(NAME)         NAME :: Monad m => Parser Char m Char
 -- XXX Need to use the predicates from Unicode.Char module/unicode-data package
-#define CHAR_PARSER(NAME, PREDICATE)  NAME = Parser.satisfy Char.PREDICATE
-#define CHAR_PARSER_DOC(PREDICATE) -- | Match any character that satisfies 'Char.PREDICATE'
+#define CHAR_PARSER(NAME, PREDICADoubleState)  NAME = Parser.satisfy Char.PREDICADoubleState
+#define CHAR_PARSER_DOC(PREDICADoubleState) -- | Match any character that satisfies 'Char.PREDICADoubleState'
```
harendra-kumar, line 85: Why this change?

rnjtranjan: due to copy replace.

PR #2283, opened 2023-03-01
Fix and enhance deintercalate and sepBy parsers

harendra-kumar:


core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -2538,7 +2688,163 @@ deintercalate :: Monad m =>
...
+      Deintercalate1InitL !Int !fs !sp
+    | Deintercalate1InitR !fs
+    | Deintercalate1R !Int !fs !ss
+    | Deintercalate1RL !Int !b !fs !sp
+
+-- | Apply two parsers alternately to an input stream. The input stream is
+-- considered an interleaving of two patterns. The two parsers represent the
+-- two patterns. Parsing starts at the first parser and stops at the first
+-- parser. It can be used to parse a infix style pattern e.g. p1 p2 p1 . Empty
+-- input or single parse of the first parser is accepted.
```
adithyaov, line 2825: From the examples below, empty input isn't accepted. Also as the name suggests.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -2551,93 +2857,150 @@ deintercalate
...
-                            resL <- initialL
-                            case resL of
-                                IPartial sL ->
-                                    return $ Partial n $ DeintercalateL fs1 sL
-                                IDone _ -> errMsg "left" "succeed"
-                                IError _ -> errMsg "left" "fail"
-                        FL.Done c -> return $ Done n c
-                else error "deintercalate: infinite loop"
-            Error err -> return $ Error err
+            Partial n s -> return $ Continue n (Deintercalate1R (cnt1 - n) fs s)
```
adithyaov, line 2898: Why `Continue` here on `Partial`?

core/src/Streamly/Internal/Data/Parser/ParserD.hs
adithyaov: This should be the same as `deintercalate` except for the `extract` function right?

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -2551,93 +2857,150 @@ deintercalate
...
 --
 {-# INLINE sepBy #-}
 sepBy :: Monad m =>
     Parser a m b -> Parser a m x -> Fold m b c -> Parser a m c
+-- This has similar performance as the custom impl below.
+-- sepBy p1 p2 f = deintercalate p1 p2 (FL.catLefts f)
 sepBy
-    (Parser pstep pinitial pextract)
-    (Parser sstep sinitial _)
+    (Parser stepL initialL extractL)
```
adithyaov, line 3002: This should be exactly the same as `deintercalate` except for the folding end resullts.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -2721,6 +3115,139 @@ sepBy1 p sep sink = do
...
+-- Left (ParseError "takeWhile1: end of input")
+-- >>> Stream.parse p $ Stream.fromList "1"
+-- Right ["1"]
+-- >>> Stream.parse p $ Stream.fromList "1+"
+-- Right ["1"]
+-- >>> Stream.parse p $ Stream.fromList "1+2+3"
+-- Right ["1","2","3"]
+--
+{-# INLINE sepBy1 #-}
+sepBy1 :: Monad m =>
```
adithyaov, line 3141: I have just skimmed through this. But this should be the same as `deintercalate1` except for the folding function.

adithyaov (commented):
Some comments.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
harendra-kumar: No, the states and some logic is different to first collect a single input and then collect more.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
harendra-kumar: States are simpler.  As the docs say:

-- >>> sepBy p1 p2 f = Parser.deintercalate p1 p2 (Fold.catLefts f)

PR #2285, opened 2023-03-01
Fix escape processing in wordWithQuotes

harendra-kumar:
To escape characters in the same way as shell and to be able to specify escape transliteration, and transliterate based on the type of quote.

### DIFF
--- a/core/src/Streamly/Internal/Data/Parser/ParserD.hs
+++ b/core/src/Streamly/Internal/Data/Parser/ParserD.hs
@@ -146,7 +146,8 @@ module Streamly.Internal.Data.Parser.ParserD
     -- spaces or quotes. No nesting.
     , wordFramedBy -- XXX Remove this? Covered by wordWithQuotes?
     , wordWithQuotes
-    , wordStripQuotes
+    , wordKeepQuotes
+    , wordProcessQuotes
 
     -- Framed by separate start and end characters, potentially nested.
     -- blockWithQuotes allows quotes inside a block. However,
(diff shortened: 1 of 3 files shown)


core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1774,12 +1842,19 @@ wordWithQuotesWith isInvalid keepQuotes isEsc toRight isSep
...
+                     else processQuoted s a n ql qr
     step (WordUnquotedEsc s) a = processUnquoted s a
-    step (WordQuotedEsc s n ql qr) a = process s a n ql qr
+    step (WordQuotedEsc s n ql qr) a =
+        case tr ql a of
+            Nothing -> do
+                res <- fstep s escChar
+                case res of
+                    FL.Partial s1 -> processQuoted s1 a n ql qr
+                    FL.Done b -> return $ Done 0 b
```
adithyaov, line 1856: If there is no translated char you keep the escape char and the char after it.
Maybe you should error out here?
Shouldn't escape char always be processed?
You should never escape something that is not escapable.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
```diff
@@ -1800,86 +1875,42 @@ wordWithQuotesWith isInvalid keepQuotes isEsc toRight isSep
...
+-- and processing the quoting and escaping at the same time.
 --
--- Similarly, when printing, double quotes are escaped by Haskell.
+-- >>> wordProcessQuotes = Parser.wordWithQuotes False (\_ _ -> Nothing)
 --
-{-# INLINE wordStripQuotes #-}
-wordStripQuotes :: (Monad m, Eq a) =>
-       (a -> Bool)     -- ^ Matches an escape elem?
+{-# INLINE wordProcessQuotes #-}
+wordProcessQuotes :: (Monad m, Eq a) =>
```
adithyaov, line 1905: I like `wordStripQuotes` over `wordProcessQuotes`.
By default, I can't think of process as utilizing the quotes.
I guess you can use `wordUtilizeQuotes`.
I guess process is also fine.

adithyaov (commented):
Some comments.

core/src/Streamly/Internal/Data/Parser/ParserD.hs
adithyaov: This is actually pretty general.
You can escape the quotes or escape the escape character.

adithyaov: For example,
`"abc\"\\d"` is valid.
`"abc\"\d"` should be invalid.
The function above can be used to represent both I guess.

adithyaov: NVM, actually looks good.

adithyaov (approved):

PR #2293, opened 2023-03-03
Rename chunksOf to groupsOf

adithyaov:
* Deprecate chunksOf in Data.Fold as it is exposed.
* No deprecation is required in Data.Stream as it is a new module.

### DIFF
--- a/src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
+++ b/src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
@@ -193,7 +193,7 @@ import qualified Streamly.Internal.Data.Parser.ParserD as PRD (Parser(..))
 import qualified Streamly.Internal.Data.Stream.IsStream.Type as IsStream
 import qualified Streamly.Internal.Data.Stream.StreamD as D
     ( foldMany
-    , chunksOf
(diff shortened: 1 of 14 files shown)


src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
```diff
@@ -983,7 +983,7 @@ splitOnSuffixSeqAny _subseq _f _m = undefined
 chunksOf
     :: (IsStream t, Monad m)
     => Int -> Fold m a b -> t m a -> t m b
-chunksOf n f = fromStreamD . D.chunksOf n f . toStreamD
+chunksOf n f = fromStreamD . D.groupsOf n f . toStreamD
```
harendra-kumar, line 986: No need to change this. This is a deprecated module. Why make a breaking change in this?

harendra-kumar (changes requested):

src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
adithyaov: There is no breaking change in this.

harendra-kumar (approved):

PR #2314, opened 2023-03-07
Fix Windows hanging test cases

rnjtranjan:


rnjtranjan:
On Windows defaultChunkSize > 1024 hangs.

PR #2318, opened 2023-03-07
Fix Targets.hs

rnjtranjan:


hie.yaml
```diff
@@ -110,8 +110,6 @@ cradle:
               component: "test:Data.Stream.Concurrent"
             - path: "./test/Streamly/Test/Data/Stream/Common.hs"
               component: "test:Data.Stream.Concurrent"
-            - path: "./test/Streamly/Test/Data/Parser/Chunked.hs"
-              component: "test:Data.Parser.Chunked"
             - path: "./test/Streamly/Test/Data/Parser/ParserD.hs"
               component: "test:Data.Parser.ParserD"
```
harendra-kumar, line 114: This can also be removed.

harendra-kumar (changes requested):
"Data.Parser.ParserD" target can also be removed.

PR #2344, opened 2023-03-13
Post release tasks

adithyaov:


adithyaov:
Bumping the version here means we need to bump the versions of `streamly` and `streamly-core` in all the tools that we use the latest `streamly` on. We can make the tools independent of the running `streamly` version and use the one that is on hackage.

harendra-kumar (approved):
streamly-process can come from hackage

PR #2370, opened 2023-03-20
Werror CI for ghc-9.4.4

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2366

harendra-kumar (changes requested):
We do not need 8.10.7-werror anymore, we can remove it:

- name: 8.10.7-Werror

Issue #2353, opened 2023-03-15
Expose unsafe mutable array APIs


harendra-kumar:
Required by the `WordCountParallelUTF8` example:

```
import qualified Streamly.Internal.Data.Array.Mut.Type as MArray
       (getIndexUnsafe, putIndexUnsafe, modifyIndexUnsafe)
```

PR #2375, opened 2023-03-20
Expose unsafe mutable array APIs

rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2353

### DIFF
--- a/core/src/Streamly/Data/MutArray.hs
+++ b/core/src/Streamly/Data/MutArray.hs
@@ -47,7 +47,6 @@ module Streamly.Data.MutArray
 
     -- * Appending streams
     , writeAppendN
-    , writeAppendNUnsafe
     , writeAppend
 
     -- * Inplace mutation
@@ -72,7 +71,7 @@ module Streamly.Data.MutArray
 
     -- * Size
     , length
-    
+
     -- * Unbox Type Class
     , Unbox (..)
     )
--- a/core/src/Streamly/Data/MutArray.hs
+++ b/core/src/Streamly/Data/MutArray.hs
@@ -47,13 +47,17 @@ module Streamly.Data.MutArray
 
     -- * Appending streams
     , writeAppendN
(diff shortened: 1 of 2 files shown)


core/src/Streamly/Data/MutArray.hs
```diff
@@ -69,6 +69,12 @@ module Streamly.Data.MutArray
     -- * Size
     , length
 
+    -- * Unsafe functions
+    , getIndexUnsafe
+    , putIndexUnsafe
+    , modifyIndexUnsafe
+    , writeAppendNUnsafe
```
harendra-kumar, line 76: Is writeAppendNUnsafe being used by some code example or some user?

core/src/Streamly/Data/MutArray.hs
```diff
@@ -69,6 +69,12 @@ module Streamly.Data.MutArray
     -- * Size
     , length
 
+    -- * Unsafe functions
+    , getIndexUnsafe
+    , putIndexUnsafe
+    , modifyIndexUnsafe
```
harendra-kumar, line 75: We should not group unsafe functions separately, it will be harder to find these. Should be put next to the similar safe functions.

Also can you check if the documentation of these is accurate.

harendra-kumar (changes requested):

core/src/Streamly/Data/MutArray.hs
rnjtranjan: No.

core/src/Streamly/Data/MutArray.hs
```diff
@@ -47,13 +47,17 @@ module Streamly.Data.MutArray
 
     -- * Appending streams
     , writeAppendN
+    , writeAppendNUnsafe
```
harendra-kumar, line 50: Remove this.

harendra-kumar (changes requested):

Issue #2352, opened 2023-03-15
Expose splitOn and chunksSplitOn


harendra-kumar:
Required by the CSV parser example.

PR #2376, opened 2023-03-20
Expose splitOn and chunksSplitOn

rnjtranjan:
https://github.com/composewell/streamly/issues/2352

### DIFF
--- a/core/src/Streamly/Data/Stream.hs
+++ b/core/src/Streamly/Data/Stream.hs
@@ -436,6 +436,10 @@ module Streamly.Data.Stream
     , parseMany
     , Array.chunksOf
 
+    -- * Splitting
+    , splitOn
+    , chunksSplitOn
+
     -- * Buffered Operations
     -- | Operations that require buffering of the stream.
     -- Reverse is essentially a left fold followed by an unfold.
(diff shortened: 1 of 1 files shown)


core/src/Streamly/Data/Array.hs
```diff
@@ -166,3 +170,7 @@ read = reader
 {-# INLINE_NORMAL readRev #-}
 readRev :: (Monad m, Unbox a) => Unfold m (Array a) a
 readRev = readerRev
+
+chunksSplitOn :: (Monad m, Unbox a) =>
+    (a -> Bool) -> Array a -> Stream m (Array a)
+chunksSplitOn = A.splitOn
```
harendra-kumar, line 176: This is not what I meant by chunksSplitOn. You can remove this from this PR. We can discuss what is the right thing and add it in another PR.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Stream/StreamD/Transform.hs
```diff
@@ -1895,7 +1895,8 @@ catEithers = fmap (either id id)
 -- separator elements determined by the supplied predicate, separator is
 -- considered as infixed between two segments:
 --
--- >>> splitOn' p xs = Stream.fold Fold.toList $ Stream.splitOn p Fold.toList (Stream.fromList xs)
+-- >>> import Streamly.Internal.Data.Stream.StreamD.Transform (splitOn)
```
harendra-kumar, line 1898: Cannot use internal imports in documentation of exposed APIs.

core/src/Streamly/Internal/Data/Stream/StreamD/Transform.hs
```diff
@@ -1895,7 +1895,8 @@ catEithers = fmap (either id id)
...
+-- >>> splitOn' p xs = Stream.fold Fold.toList $ splitOn p Fold.toList (Stream.fromList xs)
```
harendra-kumar, line 1899: Should use Stream.splitOn . The example should work if someone uses the imports as shown in the haddock docs in GHCi.

harendra-kumar (changes requested):

Issue #2351, opened 2023-03-15
Expose the Enumeration APIs from the Unfold module


harendra-kumar:
These are used in an example in streamly-examples.

harendra-kumar:
We should export the same APIs as we are exporting from the Stream module.

PR #2377, opened 2023-03-20
Expose the Enumeration APIs from the Unfold module

rnjtranjan:
https://github.com/composewell/streamly/issues/2351

### DIFF
--- a/core/src/Streamly/Data/Unfold.hs
+++ b/core/src/Streamly/Data/Unfold.hs
@@ -231,7 +231,7 @@ import Streamly.Internal.Data.Unfold
 -- writing an unfold you can convert it to stream for free using
 -- 'Stream.unfold'.
 
--- | Unfolds @from@ generating a stream starting with the element
+-- | Enumerates @from@ generating a stream starting with the element
   -- @from@, enumerating up to 'maxBound' when the type is 'Bounded' or
   -- generating an infinite stream when the type is not 'Bounded'.
   --
@@ -257,7 +257,7 @@ import Streamly.Internal.Data.Unfold
 enumerate :: (Monad m, Enumerable a) => Unfold m a a
 enumerate = enumerateFrom
 
--- | Unfolds @(from, to)@ generating a finite stream starting with the element
+-- | Enumerates @(from, to)@ generating a finite stream starting with the element
   -- @from@, enumerating the type up to the value @to@. If @to@ is smaller than
   -- @from@ then an empty stream is returned.
   --
--- a/core/src/Streamly/Data/Unfold.hs
+++ b/core/src/Streamly/Data/Unfold.hs
@@ -53,6 +53,9 @@ module Streamly.Data.Unfold
     , replicateM
     , iterateM
 
+    -- ** Enumeration
+    , Enumerable (..)
+
     -- ** From Containers
     , fromList
     , fromListM
@@ -85,10 +88,6 @@ module Streamly.Data.Unfold
     -- ** Nesting
     , many
 
-    -- ** Enumerating 'Num' Types
-    , Enumerable (..)
-    , enumerate
-    , enumerateTo
     )
 where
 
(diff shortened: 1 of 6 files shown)


core/src/Streamly/Data/Unfold.hs
```diff
@@ -85,6 +85,37 @@ module Streamly.Data.Unfold
...
+    , enumerateFromThenSmallBounded
+    , enumerateFromToSmall
+    , enumerateFromThenToSmall
+
+    -- ** Enumerating 'Fractional' Types
+    -- | Enumeration of 'Num' specialized to 'Fractional' types.
+    , enumerateFromFractional
+    , enumerateFromThenFractional
+    , enumerateFromToFractional
+    , enumerateFromThenToFractional
```
harendra-kumar, line 117: Why are we exposing all these internal APIs? Please read the issue. Expose APIs similar to what is exposed in the Stream module.

harendra-kumar (changes requested):

core/src/Streamly/Data/Unfold.hs
```diff
@@ -85,6 +85,9 @@ module Streamly.Data.Unfold
     -- ** Nesting
     , many
 
+    -- ** Enumerating 'Num' Types
+    , enumerate
+    , enumerateTo
```
harendra-kumar, line 90: We have to expose the Enumerable type class and its members as well, that was the main thing to do in this PR.

harendra-kumar (changes requested):

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,58 @@ import Streamly.Internal.Data.Unfold
 -- streams are written, it is easy to adapt a stream to an unfold. If you are
 -- writing an unfold you can convert it to stream for free using
 -- 'Stream.unfold'.
+
+-- | Unfolds @from@ generating a stream starting with the element
```
adithyaov, line 234: The API is called `enumerate`
The documentation needs to be changed!
Why is it named `enumerate` and not `enumerateFrom`?
Why not change the name to `enumerate` in the internal module itself?

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,58 @@ import Streamly.Internal.Data.Unfold
...
+  -- [1.1,2.1,3.1,4.1]
+  --
+  -- >>> Stream.fold Fold.toList $ Stream.unfold Unfold.enumerateFromTo (1.1, 4.6)
+  -- [1.1,2.1,3.1,4.1,5.1]
+  --
+  -- @
+  --
+{-# INLINE enumerateTo #-}
+enumerateTo :: (Monad m, Enumerable a) => Unfold m (a, a) a
+enumerateTo = enumerateFromTo
```
adithyaov, line 287: The same comments as earlier.

core/src/Streamly/Data/Unfold.hs
rnjtranjan: The name "enumerate" API is used in Stream, so to be the matching name.

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,58 @@ import Streamly.Internal.Data.Unfold
 -- streams are written, it is easy to adapt a stream to an unfold. If you are
 -- writing an unfold you can convert it to stream for free using
 -- 'Stream.unfold'.
+
+-- | Enumerates @from@ generating a stream starting with the element
+  -- @from@, enumerating up to 'maxBound' when the type is 'Bounded' or
+  -- generating an infinite stream when the type is not 'Bounded'.
+  --
+  -- >>> import qualified Streamly.Data.Stream as Stream
+  -- >>> import qualified Streamly.Data.Unfold as Unfold
```
harendra-kumar, line 239: Any imports should be part of #include "DocTestDataUnfold.hs" .

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,58 @@ import Streamly.Internal.Data.Unfold
...
+  --
+  -- @
+  -- >>> Stream.fold Fold.toList $ Stream.take 4 $ Stream.unfold Unfold.enumerateFrom (0 :: Int)
+  -- [0,1,2,3]
+  --
+  -- @
```
harendra-kumar, line 245: doctests should not be inside `@` blocks.

harendra-kumar (changes requested):

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,46 @@ import Streamly.Internal.Data.Unfold
...
+  -- For 'Fractional' types, enumeration is numerically stable. However, no
+  -- overflow or underflow checks are performed.
+  --
+  -- @
+  -- >>> Stream.fold Fold.toList $ Stream.take 4 $ Stream.unfold Unfold.enumerateFrom 1.1
+  -- [1.1,2.1,3.1,4.1]
+  --
+{-# INLINE enumerate #-}
+enumerate :: (Monad m, Enumerable a) => Unfold m a a
+enumerate = enumerateFrom
```
harendra-kumar, line 252: There is no point of this function if it is the same as enumerateFrom.

core/src/Streamly/Data/Unfold.hs
```diff
@@ -226,3 +230,46 @@ import Streamly.Internal.Data.Unfold
...
+  -- @
+  -- >>> Stream.fold Fold.toList $ Stream.unfold Unfold.enumerateFromTo (1.1, 4)
+  -- [1.1,2.1,3.1,4.1]
+  --
+  -- >>> Stream.fold Fold.toList $ Stream.unfold Unfold.enumerateFromTo (1.1, 4.6)
+  -- [1.1,2.1,3.1,4.1,5.1]
+  --
+{-# INLINE enumerateTo #-}
+enumerateTo :: (Monad m, Enumerable a) => Unfold m (a, a) a
+enumerateTo = enumerateFromTo
```
harendra-kumar, line 275: There is no point of this function if this is same as enumerateFromTo.

harendra-kumar (commented):
I am going to rework and commit this.

Issue #2347, opened 2023-03-14
Expose wordsBy from Data.Stream module


harendra-kumar:
The wordBy parser used in the word frequency counter example did not perform as well as the wordsBy in the stream module. We can investigate that but in the mean time we can expose the stream wordsBy instead.

PR #2379, opened 2023-03-20
Expose wordsBy from Data.Stream module

rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2347

### DIFF
--- a/core/src/Streamly/Data/Stream.hs
+++ b/core/src/Streamly/Data/Stream.hs
@@ -200,6 +200,7 @@ module Streamly.Data.Stream
 
     -- ** Parsing
     , parse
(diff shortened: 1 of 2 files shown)


core/src/Streamly/Data/Stream.hs
```diff
@@ -200,6 +200,7 @@ module Streamly.Data.Stream
 
     -- ** Parsing
     , parse
+    , wordsBy
```
harendra-kumar, line 203: Can we copy the documentation from Streamly.Prelude?

harendra-kumar (changes requested):

Issue #2346, opened 2023-03-14
Expose toHashMapIO from the Data.Fold.Prelude module in "streamly" package


harendra-kumar:
Just like the toMapIO function in Data.Fold. This is required to implement the word frequency counter example without using internal modules.

PR #2380, opened 2023-03-20
Expose toHashMapIO from the Data.Fold.Prelude module in "streamly"

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2346

Issue #2334, opened 2023-03-12
Verify the Array APIs for consistency


harendra-kumar:
All array modules should have similar immutable APIs.

All mutable array modules should have similar mutation APIs.

Must be consistent with file/Handle API.

harendra-kumar:
Done.

PR #2391, opened 2023-04-13
Unify Array APIs

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2334

core/src/Streamly/Internal/Data/Array/Generic/Mut/Type.hs
```diff
@@ -27,6 +27,11 @@ module Streamly.Internal.Data.Array.Generic.Mut.Type
     , writeN
     , writeWith
     , write
+    , fromStreamN
+    , fromStream
+
+    , fromListN
+    , fromList
```
harendra-kumar, line 34: What is the reason for exposing these? I do need see these used or exported from any exposed module. Are these supposed to be exposed from the released module?

core/src/Streamly/Internal/Data/Array/Generic/Mut/Type.hs
```diff
@@ -529,12 +535,12 @@ toList arr@MutArray{..} = mapM (`getIndexUnsafe` arr) [0 .. (arrLen - 1)]
...
 --
--- @toStreamD = D.unfold read@
+-- @read = D.unfold read@
 --
 -- We can try this if the unfold has any performance issues.
-{-# INLINE_NORMAL toStreamD #-}
-toStreamD :: MonadIO m => MutArray a -> D.Stream m a
-toStreamD arr@MutArray{..} =
+{-# INLINE_NORMAL read #-}
+read :: MonadIO m => MutArray a -> D.Stream m a
```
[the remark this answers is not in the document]
rnjtranjan: it uses "getIndexUnsafe" which requires a MonadIO constraint.

core/src/Streamly/Data/MutArray/Generic.hs
```diff
@@ -35,15 +35,22 @@ module Streamly.Data.MutArray.Generic
...
     , putIndex
     , modifyIndex
+
+    -- * Size
+    , length
     )
 where
 
 import Streamly.Internal.Data.Array.Generic.Mut.Type
+import  Prelude hiding(read, length)
```
adithyaov, line 54:
```suggestion
import  Prelude hiding (read, length)
```

core/src/Streamly/Internal/Data/Array/Generic/Mut/Type.hs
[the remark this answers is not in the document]
adithyaov: This is a mutable array. This needs to have a MonadIO constraint.

core/src/Streamly/Internal/Data/Array/Generic/Mut/Type.hs
```diff
@@ -638,6 +638,26 @@ writeWith elemCount = FL.rmapM extract $ FL.foldlM' step initial
...
+-- array is allocated to size @n@, if the stream terminates before @n@
+-- elements then the array may hold less than @n@ elements.
+--
+{-# INLINE fromStreamN #-}
+fromStreamN :: MonadIO m => Int -> Stream m a -> m (MutArray a)
+fromStreamN n = D.fold (writeN n)
+
+{-# INLINE fromStream #-}
+fromStream :: MonadIO m => Stream m a -> m (MutArray a)
+fromStream = D.fold write
```
adithyaov, line 651: Do we need these? INLINE these definations?

PR #2393, opened 2023-04-18
Fix hlint

rnjtranjan:


adithyaov:
We can use any of these https://phoenixnap.com/kb/flatpak-vs-snap-vs-appimage to package applications, including hlint for CIs.

adithyaov:
Use `AppImage`

harendra-kumar:
We can use https://docs.appimage.org/ to build a distribution independent image of hlint.

PR #2400, opened 2023-05-01
Add bound config param & expose for concurrent folds

shlok:
(My use case: Concurrently running multiple folds (put together with `demux`) on bound threads.)

I have tried to make this change unintrusive: The new `bound` flag is only exposed in `*Fold.Concurrent*` and all existing calls to `doFork` (apart from the one in `*Fold.Concurrent*`) look exactly the same as before.

Adding `bound` support everywhere would apparently require more sweeping changes in the codebase (adding an extra `Bool` parameter to all sorts of functions, etc.), so I hesitated to do this for now. If this is something you really would prefer at this time, or if you have any other suggestions for improvements, feel free to let me know.

### DIFF
--- a/src/Streamly/Internal/Control/ForkLifted.hs
+++ b/src/Streamly/Internal/Control/ForkLifted.hs
@@ -9,12 +9,13 @@
 module Streamly.Internal.Control.ForkLifted
     (
       doFork
+    , doFork'
     , fork
     , forkManaged
     )
 where
 
-import Control.Concurrent (ThreadId, forkIO)
+import Control.Concurrent (ThreadId, forkIO, forkOS)
 import Control.Exception (SomeException(..), catch, mask)
 import Data.Functor (void)
 import Streamly.Internal.Control.Concurrent (MonadRunInIO, RunInIO(..), withRunInIO, withRunInIONoRestore)
@@ -32,11 +33,23 @@ doFork :: MonadRunInIO m
     -> RunInIO m
     -> (SomeException -> IO ())
     -> m ThreadId
-doFork action (RunInIO mrun) exHandler =
+doFork = doFork' False
+
+-- | Similar to 'doFork', but has a "bound" boolean parameter for specifying
+-- whether 'forkOS' should be used instead of 'rawForkIO'.
+{-# INLINE doFork' #-}
+doFork' :: MonadRunInIO m
+    => Bool
+    -> m ()
+    -> RunInIO m
+    -> (SomeException -> IO ())
+    -> m ThreadId
+doFork' bound action (RunInIO mrun) exHandler =
     withRunInIO $ \run ->
         mask $ \restore -> do
-                tid <- rawForkIO $ catch (restore $ void $ mrun action)
-                                         exHandler
+                tid <- (if bound then forkOS else rawForkIO) $
+                    catch (restore $ void $ mrun action)
+                          exHandler
                 run (return tid)
 
 -- | 'fork' lifted to any monad with 'MonadBaseControl IO m' capability.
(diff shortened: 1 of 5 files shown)


src/Streamly/Internal/Control/ForkLifted.hs
```diff
@@ -32,11 +33,23 @@ doFork :: MonadRunInIO m
...
     -> (SomeException -> IO ())
     -> m ThreadId
-doFork action (RunInIO mrun) exHandler =
+doFork = doFork' False
+
+-- | Similar to 'doFork', but has a "bound" boolean parameter for specifying
+-- whether 'forkOS' should be used instead of 'rawForkIO'.
+{-# INLINE doFork' #-}
+doFork' :: MonadRunInIO m
+    => Bool
```
adithyaov, line 42: This is like some type of configuration for `doFork`
Maybe rename this to `doForkWith` where `Bool` is the config.

src/Streamly/Internal/Control/ForkLifted.hs
```diff
@@ -32,11 +33,23 @@ doFork :: MonadRunInIO m
     -> RunInIO m
     -> (SomeException -> IO ())
     -> m ThreadId
-doFork action (RunInIO mrun) exHandler =
+doFork = doFork' False
+
+-- | Similar to 'doFork', but has a "bound" boolean parameter for specifying
```
adithyaov, line 38: Double quote in haddocks has the meaning of pointing to a module I believe.

src/Streamly/Internal/Data/Stream/Channel/Types.hs
```diff
@@ -542,6 +546,14 @@ interleaved flag st = st { _interleaved = flag }
 getInterleaved :: Config -> Bool
 getInterleaved = _interleaved
 
+-- | Spawn bound threads (i.e., spawn threads using 'forkOS' instead of
+-- 'forkIO'). The default value is 'False'.
```
adithyaov, line 550: Can you make a small TODO note here that this only takes effect for concurrent folds as of now?
And open an issue to implement support for `bound` everywhere.

adithyaov (commented):
Sorry for the late review. Look good to me overall. Some minor comments.

harendra-kumar:
> Adding bound support everywhere would apparently require more sweeping changes in the codebase (adding an extra Bool parameter to all sorts of functions, etc.), so I hesitated to do this for now. If this is something you really would prefer at this time, or if you have any other suggestions for improvements, feel free to let me know.

It would be better to have it supported consistently everywhere so that there are no surprises. However, I will merge this PR right away to not let it languish.

shlok:
@harendra-kumar @adithyaov Thanks for the review and merge, and apologies for not getting to this earlier. If you guys want, I can incorporate @adithyaov’s suggestions in a separate PR.

adithyaov:
@shlok It would be great if you can incorporate the suggestions into a separate PR.

src/Streamly/Internal/Control/ForkLifted.hs
shlok: Sure; resolved in separate PR: https://github.com/composewell/streamly/pull/2406

src/Streamly/Internal/Control/ForkLifted.hs
shlok: Thanks; you’re right. Resolved in separate PR: https://github.com/composewell/streamly/pull/2406

src/Streamly/Internal/Data/Stream/Channel/Types.hs
shlok: Sure; resolved in separate PR: https://github.com/composewell/streamly/pull/2406

shlok:
@adithyaov Sure; no problem. See https://github.com/composewell/streamly/pull/2406

PR #2402, opened 2023-05-10
Add warning to demux documentation

shlok:
I apparently need to do this at one point for things to go smoothly. (However, if this makes no sense (based on the way `demux` is known to work), feel free to reject.)

### DIFF
--- a/core/src/Streamly/Internal/Data/Fold/Container.hs
+++ b/core/src/Streamly/Internal/Data/Fold/Container.hs
@@ -326,6 +326,10 @@ demuxGeneric getKey getFold = fmap extract $ foldlM' step initial
 -- This can be used to scan a stream and collect the results from the scan
 -- output.
 --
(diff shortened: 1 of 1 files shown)


core/src/Streamly/Internal/Data/Fold/Container.hs
```diff
@@ -326,6 +326,10 @@ demuxGeneric getKey getFold = fmap extract $ foldlM' step initial
 -- This can be used to scan a stream and collect the results from the scan
 -- output.
 --
+-- /Warning/: One should call the returned monadic action to make sure the
+-- folds fully complete—even if the action returns nothing useful (e.g., a map
+-- of @()@ values).
```
harendra-kumar, line 331: This should not be the case unless there is some laziness somewhere. Can you test and confirm that this makes a difference?

harendra-kumar:
> I apparently need to do this at one point for things to go smoothly. 

Can you recheck/reproduce this?

Issue #2386, opened 2023-03-25
Remove the capacity field in the unboxed mutable arrays?


harendra-kumar:
The capacity is available in the ByteArray# itself, we do not need to duplicate it.

PR #2404, opened 2023-05-16
Remove capacity field from MutArray

rnjtranjan:


rnjtranjan:
Fixes : https://github.com/composewell/streamly/issues/2386

benchmark/streamly-benchmarks.cabal
```diff
@@ -401,7 +401,7 @@ benchmark Data.Parser
     if flag(dev)
       ghc-options: +RTS -M3000M -RTS
     else
-      ghc-options: +RTS -M2000M -RTS
+      ghc-options: +RTS -M3000M -RTS
```
adithyaov, line 404: Why was this required now and not earlier?

core/src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -1329,15 +1340,15 @@ chunksOf n (D.Stream step state) =
                     then D.Skip
                             (GroupYield
                                 contents start end1 bound (GroupStart s))
-                    else D.Skip (GroupBuffer s contents start end1 bound)
+                     else D.Skip (GroupBuffer s contents start end1 bound)
```
adithyaov, line 1343: Added a space here by mistake

core/src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -1329,15 +1340,15 @@ chunksOf n (D.Stream step state) =
...
             D.Skip s ->
                 return $ D.Skip (GroupBuffer s contents start end bound)
             D.Stop ->
                 return
                     $ D.Skip (GroupYield contents start end bound GroupFinish)
 
-    step' _ (GroupYield contents start end bound next) =
-        return $ D.Yield (MutArray contents start end bound) next
+    step' _ (GroupYield contents start end _bound next) =
```
adithyaov, line 1350: Write a comment near the datatype to remove the bound from the datatype

Issue #2385, opened 2023-03-21
API to create arrays from pure streams and string literals


harendra-kumar:
We do not have pure folds in arrays. But this https://streamly.composewell.com/haddocks/streamly-core-0.1.0/Streamly-Data-Array.html#t:Array documents a way to create them.
Essentially you could say:

```
writePure = Fold.morphInner (Identity . unsafePerformIO) Array.write
Stream.fold writePure $ Stream.fromList "hello"
```

Knowing that GHC strings are already Utf8 encoded we could create a Utf8 array like this:

```
Stream.fold writePure $ Stream.fromByteStr# "hello"
```

We could have a `fromPure` and a `fromByteStr#` in Array.

PR #2407, opened 2023-05-22
API to create arrays from pure streams and string literals

rnjtranjan:


rnjtranjan:
Fixes: https://github.com/composewell/streamly/issues/2385

core/src/Streamly/Internal/Data/Array/Mut.hs
```diff
@@ -84,3 +91,10 @@ getSlicesFromLen from len =
 fromStream :: (MonadIO m, Unbox a) => Stream m a -> m (MutArray a)
 fromStream = fromStreamD
 -- fromStream (Stream m) = P.fold write m
+
+fromByteStr# :: Addr# -> Identity (MutArray Word8)
```
adithyaov, line 95: Why `Identity`?

core/src/Streamly/Internal/Data/Array/Mut/Type.hs
```diff
@@ -1908,6 +1912,12 @@ writeWith elemCount =
 write :: forall m a. (MonadIO m, Unbox a) => Fold m a (MutArray a)
 write = writeWith (allocBytesToElemCount (undefined :: a) arrayChunkBytes)
 
+writePure :: Unbox a => Fold Identity a (MutArray a)
+writePure = FL.morphInner (Identity . unsafePerformIO) write
+
+fromPure :: Unbox a => Stream Identity a -> Identity (MutArray a)
+fromPure = D.fold writePure
```
adithyaov, line 1919: Is this safe to do?

core/src/Streamly/Internal/Data/Array/Generic.hs
```diff
@@ -113,6 +118,15 @@ writeWith elemCount = unsafeFreeze <$> MArray.writeWith elemCount
 write :: MonadIO m => Fold m a (Array a)
 write = fmap unsafeFreeze MArray.write
 
+fromPureStream :: Stream Identity a -> Array a
+fromPureStream x =
+    unsafePerformIO $ fmap (unsafeFreeze) (MArray.fromPureStream x)
```
adithyaov, line 123: Why can't you just use `MArray.fromStream` here instead?

core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -495,6 +499,60 @@ writeWith elemCount = unsafeFreeze <$> MA.writeWith elemCount
...
+-- expression in a particular case?
+--
+-- For this reason array creation folds have a MonadIO constraint. Pure folds
+-- could be unsafe and dangerous. This is dangerous especially when used with
+-- foldMany like operations.
+--
+-- >>> unsafePureWrite = Array.unsafeMakePure Array.write
+--
+{-# INLINE unsafeMakePure #-}
+unsafeMakePure :: Monad m => Fold IO a b -> Fold m a b
```
adithyaov, line 525: We shouldn't expose this.

adithyaov: This can be misused.

core/src/Streamly/Internal/Data/Stream/StreamD/Generate.hs
```diff
@@ -1143,35 +1144,35 @@ fromFoldableM = Prelude.foldr consM nil
...
 
--- | Keep reading 'Storable' elements from 'Ptr' onwards.
+-- | Keep reading 'Storable' elements from an immutable 'Ptr' onwards.
 --
 -- /Unsafe:/ The caller is responsible for safe addressing.
 --
 -- /Pre-release/
 {-# INLINE fromPtr #-}
-fromPtr :: forall m a. (MonadIO m, Storable a) => Ptr a -> Stream m a
+fromPtr :: forall m a. (Monad m, Storable a) => Ptr a -> Stream m a
```
adithyaov, line 1153: Why make this change? What is the use-case for this?

adithyaov, line 1153: I have an eerie feeling about it. Can't explain it.

adithyaov (approved):

adithyaov:
Some comments.

core/src/Streamly/Internal/Data/Array/Type.hs
harendra-kumar: not exposed, not meant to be exposed. Experimental and to be used internally where it is safe.

PR #2414, opened 2023-06-04
Future-proof against potential Prelude.foldl'

Bodigrim:
See https://github.com/haskell/core-libraries-committee/issues/167

This is, of course, a bit speculative at the moment, but given that the change does not involve CPP, it should not hurt to merge it early to simplify further impact assessment.

This PR covers `streamly-core` only: `streamly` itself does not seem to be buildable with GHC 9.6, even less so with GHC HEAD.

PR #2416, opened 2023-06-06
Future-proof against potential Prelude.foldl' - part 2

Bodigrim:
Continuing #2414. 

I built a patched GHC 9.4 to be able to complete an impact assessment without waiting for #2415. This patch now covers `streamly` package itself and all the rest of the repo.

### DIFF
--- a/benchmark/Streamly/Benchmark/Data/Stream/Common.hs
+++ b/benchmark/Streamly/Benchmark/Data/Stream/Common.hs
@@ -103,7 +103,7 @@ import qualified Streamly.Internal.Data.Stream.StreamD as StreamK
 #endif
 
 import Gauge
-import Prelude hiding (mapM, replicate)
+import Prelude hiding (Foldable(..), mapM, replicate)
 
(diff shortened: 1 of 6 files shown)


harendra-kumar (approved):

harendra-kumar:
Thanks again. Looks good.

How do we ensure that we do not break it? Should we work on the HEAD CI after fixing 9.6?

Bodigrim:
> How do we ensure that we do not break it? Should we work on the HEAD CI after fixing 9.6?

There is an ongoing work to make GHC nightlies easily available for consumption via `ghcup`, but I'm not sure about the status. CC @hasufell

hasufell:
> > How do we ensure that we do not break it? Should we work on the HEAD CI after fixing 9.6?
> 
> There is an ongoing work to make GHC nightlies easily available for consumption via `ghcup`, but I'm not sure about the status. CC @hasufell 

It's almost complete. GHC is already producing automated metadata yaml. GHCup will need a new release to deal with it properly. Could be end of this month.

PR #2417, opened 2023-06-13
Unbox variable length structures

adithyaov:


core/src/Streamly/Internal/Data/Unboxed.hs
```diff
@@ -290,6 +294,57 @@ class Unbox a where
         Int -> MutableByteArray -> a -> IO ()
     pokeByteIndex i arr = genericPokeByteIndex arr i
 
+data UnboxVSize a
```
harendra-kumar, line 297: Move to a separate module `Streamly.Data.Serialize`. Rename this to `Serialize`.

core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -0,0 +1,99 @@
...
+DERIVE_SERIALIZE_FROM_UNBOX(Word64)
+DERIVE_SERIALIZE_FROM_UNBOX(Double)
+DERIVE_SERIALIZE_FROM_UNBOX(Float)
+DERIVE_SERIALIZE_FROM_UNBOX((StablePtr a))
+DERIVE_SERIALIZE_FROM_UNBOX((Ptr a))
+DERIVE_SERIALIZE_FROM_UNBOX((FunPtr a))
+
+instance forall a. Serialize a => Serialize [a] where
+
+    size = VarSize $ \lst ->
```
adithyaov, line 71: Add `INLINE` pragmas

test/Streamly/Test/Data/Unbox.hs
```diff
@@ -146,6 +148,27 @@ testGenericConsistency val = do
 checkSizeOf :: forall a. Unbox a => Proxy a -> Int -> IO ()
 checkSizeOf _ size = sizeOf (Proxy :: Proxy a) `shouldBe` size
 
+testUnboxVList
```
adithyaov, line 151: Rename this function

core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -0,0 +1,99 @@
...
+-- Types
+--------------------------------------------------------------------------------
+
+data Size a
+    = VarSize (a -> Int)
+    | ConstSize (Proxy a -> Int)
+
+class Serialize a where
+    size :: Size a
+    peekByteIndex :: Int -> MutableByteArray -> IO (Int, a)
```
harendra-kumar, line 33: Need to document the APIs.

Also, we can put a dev comment here. We can implement this without returning the `Int` offset but that may require traversing the Haskell structure again to get the size. Therefore, this is a performance optimization.

core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -0,0 +1,99 @@
...
+                    lst
+            ConstSize f ->
+                length lst
+                    * f (Proxy :: Proxy a)
+                    + Unbox.sizeOf (Proxy :: Proxy Int)
+
+    peekByteIndex off arr = do
+        len <- Unbox.peekByteIndex off arr :: IO Int
+        let off1 = off + Unbox.sizeOf (Proxy :: Proxy Int)
+        let peekList buf o 0 = pure (o, reverse buf)
```
harendra-kumar, line 86: In future, for the cases when the order does not matter we can define an alternative instance that does not reverse. That can be done on a newtype wrapper over list.

harendra-kumar (changes requested):
* Document the APIs
* Add INLINE

core/src/Streamly/Internal/Data/Serialize.hs
harendra-kumar: We can name these `deserialize` and `serialize` to avoid naming conflict with Unbox?

rnjtranjan (approved):

core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -0,0 +1,152 @@
...
+-- value or can be independent.
+data Size a
+    = VarSize (a -> Int)
+    | ConstSize !Int
+
+-- | A type implementing the 'Serialize' interface supplies operations for
+-- reading and writing the type from and to a mutable byte array (an unboxed
+-- representation of the type) in memory. The read operation 'deserialize'
+-- deserializes the boxed type from the mutable byte array. The write operation
+-- 'serialize' serializes the boxed type to the mutable byte array.
```
harendra-kumar, line 37: We should mention that the byte ordering in the serialized data is the same as the machine byte order. Therefore, it cannot be compatible across machines with different byte ordering.

harendra-kumar (approved):

PR #2426, opened 2023-07-11
Remove deprecated modules from streamly version <= 0.8.1

rnjtranjan:


rnjtranjan:
There are no modules in the current repo which is deprecated on or before **v.0.7.3**.
Streamly module is deprecated since v0.8.0.

harendra-kumar:
Let's keep these around for one more major release then.

Issue #2361, opened 2023-03-16
Internal Stream module cleanup


harendra-kumar:
* Rename Internal.Data.Stream.StreamD to Internal.Data.Stream
* Rename Internal.Data.Stream.StreamK to Internal.Data.StreamK

Do the same for benchmarks and tests as well.

PR #2428, opened 2023-07-14
Rename Internal.Data.Stream.StreamD to Internal.Data.Stream

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2361

core/src/Streamly/Internal/Data/Stream/StreamD.hs
```diff
@@ -1,42 +0,0 @@
--- |
```
harendra-kumar, line 1: We can deprecate this module and export Streamly.Internal.Data.Stream from it. So that if anyone has imported this, it continues working.

docs/User/Project/Upgrading.md
```diff
@@ -94,7 +94,7 @@ The `CrossStream` type in `Streamly.Internal.Data.Stream` is a
 wrapper over `Stream` type supplying the Monad instance. However, see
 the "Stream Fusion" section in the `Streamly.Data.Stream` module for
 limitations of the `Stream` type. The `StreamK` type and `CrossStreamK`
-(in Streamly.Internal.Data.Stream.StreamK) could be used to overcome the
+(in Streamly.Internal.Data.StreamK) could be used to overcome the
```
harendra-kumar, line 97: This guide is to upgrade from 0.8 to 0.9, therefore, should not reflect changes in 0.10 .

harendra-kumar (changes requested):
We should keep Streamly.Internal.Data.Stream.StreamD as well but deprecate it. We can just export Stream.Internal.Data.Stream from it.

PR #2431, opened 2023-07-14
Signature of getIndex of mutable array is inconsistent

rnjtranjan:


core/src/Streamly/Data/MutArray/Generic.hs
```diff
@@ -37,6 +37,7 @@ module Streamly.Data.MutArray.Generic
 
     -- * Random reads
     , getIndex
+    , getIndexUnsafe
```
harendra-kumar, line 40: This may conflict with the other PR.

adithyaov (approved):

PR #2432, opened 2023-07-14
Rename chunksOfTimeout to groupsOfTimeout

rnjtranjan:


src/Streamly/Internal/Data/Stream/IsStream/Reduce.hs
```diff
@@ -41,7 +41,7 @@ module Streamly.Internal.Data.Stream.IsStream.Reduce {-# DEPRECATED "Please use
     , chunksOf
     , arraysOf
     , intervalsOf
-    , chunksOfTimeout
+    , groupsOfTimeout
```
harendra-kumar, line 44: This is a deprecated module. Do not make any changes in this.

harendra-kumar (changes requested):

PR #2439, opened 2023-07-16
Rename partitionBy to partitionWith in Data.Fold

rnjtranjan:


harendra-kumar:
Ah, adithya also fixed this. And we decided to leave it be.

PR #2394, opened 2023-04-20
Have streamly 0.7.3 build with ghc 9.4.

philderbeast:
Similar to #2167, this allows the `streamly-0.7.3.*` branch to be used with `ghc-9.4.5`.

PR #2450, opened 2023-07-20
PR only for testing the CIs.

adithyaov:
This is not meant to be merged.


adithyaov:
See https://github.com/composewell/streamly/pull/2394

Issue #2451, opened 2023-07-20
parseBreakD loses the backtracking buffer on error


harendra-kumar:
In the Error constructor handling we should return the entire backtracking buffer in the remaining stream. 

Also look at all other places where the parser driver is handling the Error constructor, for a similar class of bug.

PR #2453, opened 2023-07-20
Fix parseBreak lose backtracking Buffer

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2451

core/src/Streamly/Internal/Data/Producer/Source.hs
```diff
@@ -238,8 +238,11 @@ parse
                 let src0 = Prelude.take n (getList buf)
                     src  = Prelude.reverse src0
                 return (Right b, unread src (source Nothing))
-            Error err ->
-                return (Left (ParseError err), source Nothing)
+            Error err -> do
+                let n = length (getList buf)
+                    src0 = Prelude.take n (getList buf)
```
adithyaov, line 243:
```suggestion
                let src0 = Prelude.take n (getList buf)
```

core/src/Streamly/Internal/Data/Producer/Source.hs
```diff
@@ -238,8 +238,11 @@ parse
...
+                    src  = Prelude.reverse src0
```
adithyaov, line 244:
```suggestion
                let src  = Prelude.reverse $ getList buf
```

PR #2458, opened 2023-07-24
Merge Data.Parser.ParserD test cases to Data.Parser

rnjtranjan:

PR #2460, opened 2023-07-24
Rename Test.Data.Parser.ParserD to Test.Data.ParserK

rnjtranjan:


adithyaov:
Why are we renaming the `ParserD` tests to `ParserK`?
It makes sense if we are renaming it to `Parser`
What is the issue associated with this PR?

rnjtranjan:
taken care in https://github.com/composewell/streamly/pull/2458

PR #2461, opened 2023-07-24
Add an example to make a scientific parser

adithyaov:


core/src/Streamly/Internal/Unicode/Parser.hs
```diff
@@ -290,6 +290,7 @@ data ScientificParseState
 -- exponent) tuple. The result can be mapped to 'Double' or any other number
 -- representation e.g. @Scientific@.
 --
+-- >> parserScientific = uncurry Data.Scientific.scientific <$> 'number'
```
harendra-kumar, line 293: We can add - "For example using the `scientifc` package:"

harendra-kumar (approved):

Issue #2479, opened 2023-07-28
Expose chunksOf from the array modules instead of Stream


harendra-kumar:
So that the chunks could be of specific array type e.g. Array or MutArray.

Deprecate/remove it from the Data.Stream module.

PR #2482, opened 2023-07-29
Expose chunksOf from the array modules instead of Stream

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2479

PR #2490, opened 2023-08-02
Use Array directly as input instead of a Chunk wrapper

harendra-kumar:


adithyaov (approved):
Can we use `neverArray :: Array a` instead?
Where `arrEnd == -1`?
Kind of synonymous with error in arrays

Issue #2051, opened 2022-10-27
Expose the `setup` stanza in every module.


adithyaov:

adithyaov:
Since you'll be updating the docs, reassigning this to you @harendra-kumar

harendra-kumar:
See https://github.com/composewell/streamly/pull/2328.

The following modules require similar doctest changes:

* FileSystem modules
* Unicode modules
* All of streamly package modules

harendra-kumar:
1. We need this only for exposed modules.
2. We should not combine setup of multiple exposed modules together.

PR #2493, opened 2023-08-03
Expose setup stanza

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2051

core/src/DocTestDataFileSystem.hs
```diff
@@ -0,0 +1,13 @@
+{- $setup
+>>> :m
+>>> import qualified Streamly.Data.Array as Array
+>>> import qualified Streamly.Data.Fold as Fold
+>>> import qualified Streamly.Data.Unfold as Unfold
+>>> import qualified Streamly.Data.Stream as Stream
```
harendra-kumar, line 6: Sort the modules.

core/src/DocTestDataUnicode.hs
```diff
@@ -0,0 +1,18 @@
...
+>>> :set -XOverloadedStrings
+
+>>> import Streamly.Internal.Unicode.Stream
+>>> import Streamly.Internal.Unicode.String
+
+>>> import qualified Streamly.Data.Fold as Fold
+>>> import qualified Streamly.Data.Stream as Stream
+>>> import qualified Streamly.Internal.Unicode.Array as Array
+>>> import qualified Streamly.Internal.Unicode.Stream as Unicode
+>>> import qualified Streamly.Unicode.Parser as Unicode
```
harendra-kumar, line 14: Separate non-internal imports from internal ones. Add a comment before internal imports - see Streamly.Data.Stream.

harendra-kumar (changes requested):
Need to include the same setup stanza in corresponding released modules as well. Otherwise there is not much use of this.

core/src/DocTestDataUnicode.hs
harendra-kumar: Do this in other modules as well.

harendra-kumar (changes requested):
Need to include the same setup stanza in corresponding released modules as well. Otherwise there is not much use of this.

harendra-kumar (commented):
1. We need this only for exposed modules.
2. We should not combine setup of multiple exposed modules together.

PR #2496, opened 2023-08-03
Update comments and docs about the Unbox type class

harendra-kumar:


benchmark/Streamly/Benchmark/Data/Unbox.hs
```diff
@@ -68,35 +87,46 @@ benchSink name times f = bench name (nfIO (randomRIO (times, times) >>= f))
...
 -------------------------------------------------------------------------------
 
 main :: IO ()
-main = runWithCLIOpts defaultStreamSize allBenchmarks
+main = do
+#ifndef FUSION_CHECK
+    runWithCLIOpts defaultStreamSize allBenchmarks
+#else
+    -- Enable FUSION_CHECK macro at the beginning of the file
+    -- Enable one benchmark below, and run the benchmark
```
adithyaov, line 126: What benchmark below?

targets/Targets.hs
```diff
@@ -187,7 +187,8 @@ targets =
     , ("Data.Fold.Window", [ "parser_grp", "infinite_grp" ])
     , ("Data.Parser", [ "parser_grp", "infinite_grp" ])
 
-    , ("Data.Unbox", ["noBench"])
+    , ("Data.Unbox", [""])
+    , ("Data.Unbox.Derive.TH", [""])
```
adithyaov, line 191: What's with the empty string?

adithyaov (approved):

targets/Targets.hs
harendra-kumar: accidental.

Issue #2499, opened 2023-08-05
Put exposed-modules under flag dev in other-modules


harendra-kumar:
Otherwise they still show up in the list of modules on hackage.

PR #2503, opened 2023-08-07
Put exposed-modules under flag dev in other-modules

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2499

Issue #2495, opened 2023-08-03
Change the deprecation message to remove multiple indirections


harendra-kumar:
In the following modules:

* Streamly
* Streamly.Memory.Array

PR #2504, opened 2023-08-07
Change the deprecation message to remove multiple indirections

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2495

Issue #2477, opened 2023-07-27
Re-export entire submodules from the higher level module


harendra-kumar:
For the following modules in streamly-core package:

* Fold
* StreamK
* Unfold
* Parser
* Refold
* Producer
* Pipe
* Array
* MutArray

Once we do that we can hide the submodules under them as they will be entirely exported via higher level modules.

Also

* Remove ParserDK module
* Move Data.Parser.ParserK to Data.ParserK
* Move Data.Ring to Data.Ring.Generic
* Move Data.Ring.Unboxed to Data.Ring

Also:

* hide the internal modules corresponding to deprecated modules.
* hide Streamly.Internal.Data.Stream.IsStream.*
* hide Streamly.Internal.FileSystem.FD, Streamly.Internal.FileSystem.FDIO
* hide Streamly.Internal.System.IOVec, Streamly.Internal.System.IOVec.Type

harendra-kumar:
We can export core/src/Streamly/Internal/Data/Stream/StreamK/Transformer.hs from core/src/Streamly/Internal/Data/Stream/StreamK.hs  and hide Transformer.hs. Also, we can move core/src/Streamly/Internal/Data/Stream/StreamK.hs to core/src/Streamly/Internal/Data/StreamK.hs .

PR #2505, opened 2023-08-07
Re-export entire submodules from the higher level module Streamly.Internal.Data.Fold

rnjtranjan:


rnjtranjan:
Fixes:https://github.com/composewell/streamly/issues/2477

PR #2509, opened 2023-08-08
Re-export entire submodules from MutArray

rnjtranjan:


harendra-kumar (changes requested):
1. undo the import changes in streamly-core package
2. hide the MutArray.Type module

harendra-kumar (changes requested):
hide MutArray.Type

harendra-kumar (approved):

PR #2510, opened 2023-08-08
Re-export entire submodules from Array

rnjtranjan:


core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -241,7 +216,7 @@ readerUnsafe = Unfold step inject
 
 -- |
 --
--- >>> import qualified Streamly.Internal.Data.Array.Type as Array
+-- >>> import qualified Streamly.Internal.Data.Array as Array
```
harendra-kumar, line 219: Do we need this? This should be in the setup stanza already.

core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -567,7 +567,7 @@ pinnedWrite = fmap unsafeFreeze MA.pinnedWrite
 -- could be unsafe and dangerous. This is dangerous especially when used with
 -- foldMany like operations.
 --
--- >>> import qualified Streamly.Internal.Data.Array.Type as Array
+-- >>> import qualified Streamly.Internal.Data.Array as Array
```
harendra-kumar, line 570: Do we need this? This should be in the setup stanza already.

core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -596,7 +596,7 @@ fromPureStream x = unsafePerformIO $ fmap (unsafeFreeze) (MA.fromPureStream x)
 -- literals because they are guaranteed to be NULL terminated:
 --
 -- >>> :set -XMagicHash
--- >>> import qualified Streamly.Internal.Data.Array.Type as Array
+-- >>> import qualified Streamly.Internal.Data.Array as Array
```
harendra-kumar, line 599: Do we need this? This should be in the setup stanza already.

core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -17,42 +17,26 @@ module Streamly.Internal.Data.Array
     -- $design
 
     -- * The Array Type
-      Array
-
-    -- * Construction
```
harendra-kumar, line 22: We can keep this heading.

harendra-kumar (changes requested):
Undo the import changes in streamly-core module.
hide the Array.Type module

harendra-kumar (approved):

PR #2512, opened 2023-08-08
Re-export entire submodules from the StreamK

rnjtranjan:


core/src/Streamly/Internal/Data/Stream/StreamK.hs
```diff
@@ -0,0 +1,15 @@
+-- |
```
harendra-kumar, line 1: Remove this module. We already have Data.StreamK .

harendra-kumar (changes requested):
1. streamly-core should keep importing from the lower level module instead of using the higher level one.
2. We can move Data/Stream/StreamK/Type to Data/StreamK/Type .
3. hide the Type module

harendra-kumar (approved):

Issue #2515, opened 2023-08-08
Add window prefix to combinators in Fold.Window module


harendra-kumar:
And export this module via Internal.Data.Fold .

PR #2517, opened 2023-08-09
Add window prefix to combinators in Fold.Window module

rnjtranjan:


rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2515

core/src/Streamly/Internal/Data/Fold/Combinators.hs
```diff
@@ -820,7 +820,7 @@ length = lengthGeneric
 -- identity (@0@) when the stream is empty. Note that this is not numerically
 -- stable for floating point numbers.
 --
--- >>> sum = FoldW.cumulative FoldW.sum
+-- >>> sum = FoldW.cumulative FoldW.windowSum
```
harendra-kumar, line 823: We can now import this as `Fold` instead of `FoldW`.

PR #2519, opened 2023-08-09
Remove redundant exports from Fold.Combinators

harendra-kumar:


rnjtranjan (commented):
import Streamly.Internal.Data.Fold.Step in module Streamly.Internal.Data.Fold is redundant.

harendra-kumar:
> import Streamly.Internal.Data.Fold.Step in module Streamly.Internal.Data.Fold is redundant.

fixed.

PR #2525, opened 2023-08-11
Re-export entire submodules from Parser

rnjtranjan:


core/streamly-core.cabal
```diff
@@ -412,6 +410,10 @@ library
                     , Streamly.Internal.Data.Fold.Container
                     , Streamly.Internal.Data.Fold.Window
 
+                    , Streamly.Internal.Data.Parser.Tee
+                    , Streamly.Internal.Data.Parser.Type
+                    , Streamly.Internal.Data.ParserK.Type
```
harendra-kumar, line 415: Cannot hide ParserK.Type as it is not exported from any unhidden module.

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -12,8 +12,10 @@ module Streamly.Internal.Data.Parser
     -- * Setup
     -- $setup
 
+    module Streamly.Internal.Data.Parser.Type
```
harendra-kumar, line 15: Parser.Tee should also be exported.

rnjtranjan: It's empty , getting werror.

harendra-kumar (changes requested):

Issue #2521, opened 2023-08-09
Move chunksSplitOn from Data.Stream module to Array module


harendra-kumar:

adithyaov:
Is this issue obsolete?

adithyaov:
`chunksSplitOn` is now `MutArray.splitOn`
This issue can be closed.

adithyaov:
Deprecate `chunksOf` in `Data.Stream`
This will be exported via `Array`.

PR #2532, opened 2023-08-14
Move chunksSplitOn from Data.Stream module to Array module

rnjtranjan:


rnjtranjan:
Fix : https://github.com/composewell/streamly/issues/2521

PR #2535, opened 2023-08-15
Improve performance of the Serialize::size operation

harendra-kumar:


benchmark/Streamly/Benchmark/Data/Serialize.hs
```diff
@@ -244,30 +232,31 @@ mkBinTree = go (generate $ arbitrary)
...
-                    VarSize f -> f t
-        arr <- newBytes n
-        _ <- serialize 0 arr t
-        (_, t1) <- deserialize 0 arr
-        return $ t1 == t
-
-runQC :: IO ()
-runQC = quickCheckWith stdArgs {maxSuccess = 100} prop
--}
+-- XXX Why do we need this? Why can't we directly use getSize?
```
harendra-kumar, line 251: Earlier there was no getSize, this became so after refactoring.

harendra-kumar (commented):
Did you do benchmark diff before and after refactoring? That's very important, sometimes there is a drastic change.

adithyaov:
See #2553 for further dev.

Issue #2003, opened 2022-10-19
Rename unsafeIndex in Array module to getIndexUnsafe


harendra-kumar:
To keep it consistent with mutable array API.

PR #2540, opened 2023-08-17
Rename unsafeIndex in Array module to getIndexUnsafe

rnjtranjan:


rnjtranjan:
Fix : https://github.com/composewell/streamly/issues/2003

core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -46,7 +46,7 @@ module Streamly.Internal.Data.Array.Type
 
     -- * Elimination
     , unsafeIndexIO
-    , unsafeIndex -- getIndexUnsafe
+    , getIndexUnsafe
```
harendra-kumar, line 49: Deprecate the old API instead of removing.

harendra-kumar (approved):

PR #2552, opened 2023-08-19
Enable array bound check during deserialize

harendra-kumar:


core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -132,6 +132,7 @@ checkBounds _label _off _arr = do
     else return ()
 #endif
 
+-- XXX Is it still unsafe if we are checking the bounds?
```
harendra-kumar, line 135: It is safe but I did not change the name.

adithyaov (approved):

PR #2554, opened 2023-08-21
Improve pin/unpin in MutArray

adithyaov:


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2086,24 +2092,30 @@ fromListRev :: (MonadIO m, Unbox a) => [a] -> m (MutArray a)
...
+    liftIO $ do
+        let startSrc = arrStart src
+            srcLen = arrEnd src - startSrc
+        newArrContents <- Unboxed.newBytesAs ps srcLen
+        putSliceUnsafe (arrContents src) startSrc newArrContents 0 srcLen
+        return $ MutArray newArrContents 0 srcLen srcLen
+
+{-# INLINE clone #-}
+clone :: MonadIO m => MutArray a -> m (MutArray a)
+clone = cloneAs Unpinned
```
harendra-kumar, line 2110: Is this operation faster than streaming the array and writing it to a new array? Why do we need this?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -242,6 +245,7 @@ import Streamly.Internal.Data.Unbox
     , PinnedState(..)
     , getMutableByteArray#
     , touch
+    , putSliceUnsafe
```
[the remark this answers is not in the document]
adithyaov: Can remove this.

core/src/Streamly/Internal/Data/MutArray/Type.hs
adithyaov: I'll need to check if it's faster.

core/src/Streamly/Internal/Data/MutArray/Type.hs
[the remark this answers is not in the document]
adithyaov: Wait. Cannot remove this?

core/src/Streamly/Internal/Data/MutArray/Type.hs
adithyaov: See https://github.com/composewell/streamly/issues/2573

PR #2556, opened 2023-08-21
Add debug mode array bound checks to Unbox methods

harendra-kumar:


core/src/Streamly/Internal/Data/Unbox.hs
```diff
@@ -350,8 +350,8 @@ class Unbox a where
     -- | Read an element of type "a" from a MutableByteArray given the byte
     -- index.
     --
-    -- IMPORTANT: The implementation of this interface may not check the bounds
-    -- of the array, the caller must not assume that.
+    -- IMPORTANT: The implementation of this interface does not check the
```
adithyaov, line 353: Depends on the implementation.

core/src/Streamly/Internal/Data/Unbox.hs
```diff
@@ -362,48 +362,71 @@ class Unbox a where
     -- | Write an element of type "a" to a MutableByteArray given the byte
     -- index.
     --
-    -- IMPORTANT: The implementation of this interface may not check the bounds
-    -- of the array, the caller must not assume that.
+    -- IMPORTANT: The implementation of this interface does not check the
```
adithyaov, line 365: Depends on how we or someone else implements it in their code.

core/src/Streamly/Internal/Data/Unbox.hs
```diff
@@ -362,48 +362,71 @@ class Unbox a where
...
     {-# INLINE pokeByteIndex #-}
     default pokeByteIndex :: (Generic a, PokeRep (Rep a)) =>
         Int -> MutableByteArray -> a -> IO ()
     pokeByteIndex i arr = genericPokeByteIndex arr i
 
--- XXX Add asserts to check bounds
+{-# INLINE checkBounds #-}
+checkBounds :: String -> Int -> MutableByteArray -> IO ()
+checkBounds _label _off _arr = do
+#ifdef DEBUG
```
adithyaov, line 377: Did you check the compilation with this flag enabled?

core/src/Streamly/Internal/Data/Unbox.hs
```diff
@@ -362,48 +362,71 @@ class Unbox a where
...
+    sz <- sizeOfMutableByteArray _arr
+    if (_off > sz)
```
adithyaov, line 379: This can be `>=`.

core/src/Streamly/Internal/Data/Unbox.hs
```diff
@@ -362,48 +362,71 @@ class Unbox a where
...
 
     {-# INLINE pokeByteIndex #-}
     default pokeByteIndex :: (Generic a, PokeRep (Rep a)) =>
         Int -> MutableByteArray -> a -> IO ()
     pokeByteIndex i arr = genericPokeByteIndex arr i
 
--- XXX Add asserts to check bounds
+{-# INLINE checkBounds #-}
+checkBounds :: String -> Int -> MutableByteArray -> IO ()
+checkBounds _label _off _arr = do
```
adithyaov, line 376: Rename `_off` to `_firstInvalidOff` or document it like so.

adithyaov: This can probably act like a wrapper function maybe?

core/src/Streamly/Internal/Data/Unbox.hs
harendra-kumar: reverted the change.

core/src/Streamly/Internal/Data/Unbox.hs
harendra-kumar: reverted.

core/src/Streamly/Internal/Data/Unbox.hs
harendra-kumar: Added a -DDEBUG cpp macro during compliation when debug build flag is on.

Issue #2557, opened 2023-08-21
Change the builder type


harendra-kumar:
from `Builder (s -> m (s, a))` to `Builder (s -> m (a, s))` to make it consistent with the state monad.

PR #2561, opened 2023-08-21
Change the Builder type

rnjtranjan:


core/src/Streamly/Internal/Data/Builder.hs
```diff
@@ -27,16 +27,16 @@ import Control.Applicative (liftA2)
...
 -- a loop.
 newtype Builder s m a =
-  Builder (s -> m (s, a))
+  Builder (s -> m (a, s))
 
 -- | Maps a function on the output of the fold (the type @b@).
 instance Functor m => Functor (Builder s m) where
     {-# INLINE fmap #-}
-    fmap f (Builder step1) = Builder (fmap (fmap f) . step1)
+    fmap f (Builder step1) = Builder (fmap (\ ~(a, s') -> (f a, s')) . step1)
```
harendra-kumar, line 35: @adithyaov this is how we have to map if we have value as the first parameter. Should we keep API divergence from StateT for mapping convenience?

rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2557

core/src/Streamly/Internal/Data/Builder.hs
adithyaov: What is the reason for diverging in the first place? To conform with `StateT` should not be a reason.

harendra-kumar (approved):

PR #2563, opened 2023-08-22
Add an API to wrap an index of a MutArray into an Unboxed IORef

rnjtranjan:


rnjtranjan:
Fix : https://github.com/composewell/streamly/issues/2558

PR #2568, opened 2023-08-23
Remove the no-fusion build flag and DISABLE_FUSION cpp macro

rnjtranjan:


harendra-kumar (changes requested):
Remove from ParserK as well:

```
core/src/Streamly/Internal/Data/ParserK/Type.hs
785:#ifndef DISABLE_FUSION
```

rnjtranjan:
Fix:https://github.com/composewell/streamly/issues/2565

PR #2571, opened 2023-08-24
Reduce internal imports of "streamly-core" in "streamly" package

rnjtranjan:


src/Streamly/Data/Stream/Concurrent.hs
```diff
@@ -106,6 +106,40 @@ module Streamly.Data.Stream.Concurrent
...
+  , parMergeBy
+  , parMergeByM
+  , parRepeatM
+  , parReplicateM
+  , parSequence
+  , parTapCount
+  , parZipWith
+  , parZipWithM
+  , tapCount
+  )
```
harendra-kumar, line 142: Why explicit imports when there are so many? Also, these are not from streamly-core.

harendra-kumar (changes requested):

PR #2576, opened 2023-08-31
Update deriveSeraializeWith

adithyaov:


core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,69 +362,65 @@ deriveSerializeInternal preds headTy cons = do
             ]
     return [plainInstanceD preds (AppT (ConT ''Serialize) headTy) methods]
 
--- | Template haskell helper to create instances of 'Serialize' automatically.
+-- | Like 'deriveSerialize' but control which types variables should not get the
+-- 'Serialize' constraint and how to substitute the variables while applying to
+-- the main type.
```
harendra-kumar, line 367: it is not obvious what substitution means here. Can it be explained better with a use case/example?

harendra-kumar (approved):

core/src/Streamly/Internal/Data/Serialize/TH.hs
adithyaov: Updated the doc.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,69 +362,69 @@ deriveSerializeInternal preds headTy cons = do
             ]
     return [plainInstanceD preds (AppT (ConT ''Serialize) headTy) methods]
 
--- | Template haskell helper to create instances of 'Serialize' automatically.
+-- Similar to 'deriveSerialize,' but with the ability to specify which type
```
harendra-kumar, line 365: Need haddock markup here.

harendra-kumar (approved):

PR #2577, opened 2023-09-01
Use a configuration to control functionality of deriveSerializeWith

adithyaov:


core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -10,6 +10,8 @@
 --
 module Streamly.Internal.Data.Serialize.TH
     ( deriveSerialize
+    , SerializeTHConfig(..)
+    , defaultSerializeTHConfig
```
harendra-kumar, line 14: I would prefer to call these just `Config` and `defaultConfig`?

In fact we can assume this module to be imported qualified and drop the `Serialize` suffix from `deriveSerialize` as well.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,40 +366,94 @@ deriveSerializeInternal preds headTy cons = do
...
+          -- ^ If True, constructors with record syntax will be encoded in a
+          -- more compatible way. Allows addition, removal, and reordering of
+          -- fields. The default value is 'False'.
+-}
+        }
+
+defaultSerializeTHConfig :: SerializeTHConfig
+defaultSerializeTHConfig =
+    SerializeTHConfig
+        { unconstrainedTypeVars = []
```
harendra-kumar, line 429: `typeVarConstraints` or just `constraints`.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,40 +366,94 @@ deriveSerializeInternal preds headTy cons = do
...
+        , typeVarSubstitutions = []
```
harendra-kumar, line 430: `typeVarSpecializations` or just `specializations`.

adithyaov: `substitutions`*

core/src/Streamly/Internal/Data/Serialize/TH.hs
adithyaov: `constraints` sounds good.

core/src/Streamly/Internal/Data/Serialize/TH.hs
adithyaov: Gotcha.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,40 +372,94 @@ deriveSerializeInternal preds headTy cons = do
...
+          -- @f@ is replaced with 'Identity' and becomes unconstrained.
+        , inlineSize :: Inline
+          -- ^ Inline value for 'size'. Default is Inline.
+        , inlineSerialize :: Inline
+          -- ^ Inline value for 'serialize'. Default is Inline.
+        , inlineDeserialize :: Inline
+          -- ^ Inline value for 'deserialize'. Default is Inline.
+        , constructorTagAsString :: Bool
+          -- ^ If True, encode constructors as Latin-1 byte sequence. This
+          -- allows addition, removal, and reordering of constructors. If False
```
harendra-kumar, line 422: encode constructors using the constructor names as Latin-1 byte sequence.

> This allows addition, removal, and reordering of constructors.

Without the context for backward compatibility this may be confusing. the user should know how they can use this feature.

Should probably mark this as experimental feature.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -362,40 +372,94 @@ deriveSerializeInternal preds headTy cons = do
...
+          -- encode them as numbers. The default value is 'False'.
+        , recordSyntaxWithHeader :: Bool
+          -- ^ If True, constructors with record syntax will be encoded in a
+          -- more compatible way. Allows addition, removal, and reordering of
```
harendra-kumar, line 426: "more compatible way" is vague.

harendra-kumar (approved):

harendra-kumar (approved):

PR #2578, opened 2023-09-06
Implement the constructorTagAsString config option in Serialize.TH

adithyaov:


core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -136,8 +144,185 @@ isUnitType _ = False
...
+--------------------------------------------------------------------------------
+-- Primitive serialization
+--------------------------------------------------------------------------------
+
+-- XXX Batch serialize? See xorCmp
+serializeW8List :: Name -> Name -> [Word8] -> Q Exp
+serializeW8List off arr w8List = do
+    [|let $(varP (makeN 0)) = $(varE off)
+       in $(doE (fmap makeBind [0 .. (lenW8List - 1)] ++
+                 [noBindS ([|pure $(varE (makeN lenW8List))|])]))|]
```
harendra-kumar, line 261: May be too much code bloat. Not sure if this could be a real issue. Can we use a loop with the loop body unrolled?

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -136,8 +144,185 @@ isUnitType _ = False
...
+c2w = fromIntegral . ord
+
+shiftAdd :: Bits a => (b -> a) -> [b] -> a
+shiftAdd conv xs =
+    foldl' (.|.) zeroBits $
+    map (\(j, x) -> shiftL x (j * 8)) $ zip [0 ..] $ map conv xs
+
+-- Note: This only works in little endian machines
+xorCmp :: [Word8] -> Name -> Name -> Q Exp
+xorCmp tag off arr =
```
harendra-kumar, line 176: Instead of generating this via TH can't we write it directly in Haskell and use that? Creating one comparison function for each deserialization may be too much code and may not be necessary.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -136,8 +144,185 @@ isUnitType _ = False
...
 --------------------------------------------------------------------------------
 
-mkSizeOfExpr :: TypeOfType -> Q Exp
-mkSizeOfExpr tyOfTy =
+{-
+int_w8 :: Int -> Word8
+int_w8 = fromIntegral
+-}
+
+w8_w16 :: Word8 -> Word16
```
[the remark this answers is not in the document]
adithyaov: Snake case sometimes is more clear. IMO, Sometimes neater. we can ignore it in hlint.

core/src/Streamly/Internal/Data/Serialize/TH.hs
adithyaov: I'm going to add this as a developer comment. We can fix this if any issue arises.

core/src/Streamly/Internal/Data/Serialize/TH.hs
adithyaov: We can do that. I'm going to add this as a developer comment and open an issue for keeping track of these comments. These will be tested with a bigger project soon. We can fix these according to our experience with that.

adithyaov:
See: https://github.com/composewell/streamly/issues/2579

adithyaov:
See https://github.com/composewell/streamly/compare/th-dev-3?expand=1

PR #2580, opened 2023-09-09
Implement the constructorTagAsString config option in Serialize.TH

adithyaov:


core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -42,6 +42,19 @@ import qualified Streamly.Internal.Data.Array as Array
...
 
+--------------------------------------------------------------------------------
+-- Developer Note
+--------------------------------------------------------------------------------
+
+-- IMPORTANT
+-- =========
+--
+-- Don't ever serialize the offsets in the encoding. Serialize length instead.
+-- Offsets are NOT stable.
```
harendra-kumar, line 53: Offsets in general are perfectly ok in persistent structures. But I you are talking about the specific case of array implementation and storing the array's offset which is obviously not correct, so you should say so.

harendra-kumar (approved):

core/src/Streamly/Internal/Data/Serialize.hs
adithyaov: By offsets, I mean absolute offsets. That is like "JUMP" to an address. But in the case of serialization the address space keeps changing.

PR #2581, opened 2023-09-10
Add more instances to Serialize

adithyaov:


core/src/Streamly/Internal/Data/Serialize.hs
```diff
@@ -7,272 +12,107 @@
...
-encode = encodeAs Unpinned
+instance Serialize Integer where
+    {-# INLINE size #-}
+    size i a = size i (liftInteger a)
 
-{-# INLINE pinnedEncode #-}
-pinnedEncode :: Serialize a => a -> Array Word8
-pinnedEncode = encodeAs Pinned
+    {-# INLINE deserialize #-}
+    deserialize off arr end = fmap unliftInteger <$> deserialize off arr end
```
harendra-kumar, line 115: I am wondering if it will be simpler to implement it directly rather than using lift/unlift.

harendra-kumar (approved):

core/src/Streamly/Internal/Data/Serialize.hs
adithyaov: I'll try implementing it directly. This is just an indirection for convinence.

adithyaov: See : https://github.com/composewell/streamly/issues/2582

PR #2588, opened 2023-09-14
Use more expressive notation for deriveSerializeWith

adithyaov:


core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -381,44 +380,47 @@ mkSerializeDec (Config {..}) headTy cons = do
...
+  where
+
+    next (InstanceD mo preds headTy []) methods =
+        pure [InstanceD mo preds headTy methods]
+    next _ _ = errorUnsupported
+
+    getMainTypeName = go . unwrap
+
+    unwrap (AppT (ConT _) r) = r
+    unwrap _ = error "getMainTypeName.unwrap: Undefined instance declaration."
```
harendra-kumar, line 418: Should use the prefix `deriveSerializeWith` as that is what the user knows about. Also, does the message convey the specific error information that the user can understand, act upon and correct.

adithyaov: Yes, we should be more specific.

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -412,14 +412,31 @@ deriveSerializeWith conf mDecs = do
...
-    go (ConT nm) = nm
-    go (AppT l _) = go l
-    go _ = error "getMainTypeName.go: Undefined instance declaration."
+    errorMessage dec =
+        error $ unlines
+            [ "Error: deriveSerializeWith:"
+            , ""
+            , ">> " ++ pprint dec
+            , ""
+            , "The above is not a proper instance declaration."
```
harendra-kumar, line 421: "proper" => valid

core/src/Streamly/Internal/Data/Serialize/TH.hs
```diff
@@ -412,14 +412,31 @@ deriveSerializeWith conf mDecs = do
...
+            , "Any proper instance declaration without a body is valid."
```
harendra-kumar, line 422: proper => Haskell

harendra-kumar (approved):

PR #2591, opened 2023-09-21
Improve list Serialization Instance

adithyaov:


core/src/Streamly/Internal/Data/Serialize/Type.hs
```diff
@@ -240,13 +240,13 @@ instance forall a. Serialize a => Serialize [a] where
     -- Inlining this causes large compilation times for tests
     {-# INLINABLE serialize #-}
     serialize off arr val = do
-        void $ serialize off arr (length val)
         let off1 = off + Unbox.sizeOf (Proxy :: Proxy Int)
```
harendra-kumar, line 243: This should be sizeOf Int64.

harendra-kumar (changes requested):

PR #2602, opened 2023-11-11
Fix a bug in Fold.Window.windowRange

adithyaov:


core/src/Streamly/Internal/Data/Fold/Window.hs
```diff
@@ -300,7 +302,11 @@ windowRange n = Fold step initial extract
         if i == 0
         then return Nothing
         else do
-            x <- liftIO $ peek rh
+            -- Here we use "ringStart" over "ringHead" as "ringHead" will be
+            -- uninitialized if the ring is not full.
+            -- Using "unsafeForeignPtrToPtr" here is safe as we touch the ring
+            -- again in "foldFunc".
+            x <- liftIO $ peek (unsafeForeignPtrToPtr (Ring.ringStart rb))
```
harendra-kumar, line 309: But ringStart will be incorrect when the ring is full. We should use a condition similar to foldFunc for both cases.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/Fold/Window.hs
adithyaov: No, it'll be correct. range is position-independent. It does not matter what element we initially take.

harendra-kumar: Ok, rebase it so that we can merge.

harendra-kumar (approved):

PR #2612, opened 2023-11-21
Update changelog for 0.10.0

harendra-kumar:


core/docs/Changelog.md
```diff
@@ -2,27 +2,45 @@
...
+  `Array` comes from `Streamly.Data.Array`. This change also changed the
+  signatures of `parseChunks` and `parseBreakChunks`.
+* Changed the signature of 'Streamly.Data.Stream.handle' to make the
+  exception handler monadic.
+* Rethrow exceptions promptly in `bracketIO`.
+* `getIndex` signature changed in `Streamly.Data.MutArray` and
+  `Streamly.Data.MutArray.Generic`.
+
+### Internal Changes
+* Fold constructor has changed, added a `final` field to support
```
adithyaov, line 18: We should remove `Fold.rmapM`. Can unintentionally create an improper extract function.
Actually, we'll almost always create an improper extract function.
We can make `rmapM :: (b -> c) -> (b -> m c) -> Fold m a b -> Fold m a c`.
Can always use `rmapM (const undefined) mapper fold`. If planning to not use scans.

adithyaov (commented):
One comment.

PR #2613, opened 2023-11-21
Add Config setter functions for Serialize derivation

harendra-kumar:


core/src/Streamly/Data/Serialize.hs
```diff
@@ -51,14 +51,20 @@ module Streamly.Data.Serialize
...
     , Serialize(..)
-    -- XXX Need to export the config setters as well.
-    , defaultConfig
+
+    -- Deriving instances
+    , Config -- XXX rename to SerializeConfig
+    , defaultConfig -- XXX rename to defaultSerializeConfig
+    , inlineSize
+    , inlineSerialize
+    , inlineDeserialize
```
adithyaov, line 60: Even `deriveSerialize` looks like a config option.
You can only distinguish `inlineSerialize` and `deriveSerialize` only by looking at the signature.

core/src/Streamly/Internal/Data/Serialize/TH/Bottom.hs
```diff
@@ -65,48 +74,122 @@ import Streamly.Internal.Data.Unbox.TH (DataCon(..))
...
+-- version of the data type can still be deserialized by the new version, the
+-- field value in the older version is assumed to be 'Nothing'. If any other
+-- type of field is added, deserialization of the older version results in an
+-- error but only when that field is accessed in the deserialized record.
+-- * __Deletion__: If a field is deleted in the new version and it is
+-- encountered in a previously serialized version then the field is discarded.
+-- TBD: We can possibly designate a catch-all field to handle this case.
+--
+-- This option has to be the same on both encoding and decoding side.
+-- The default is 'False'.
```
adithyaov, line 169: The default is 'False' is mentioned twice. Even below.

core/src/Streamly/Internal/Data/Serialize/TH/Bottom.hs
```diff
@@ -65,48 +74,122 @@ import Streamly.Internal.Data.Unbox.TH (DataCon(..))
...
+-- 'inlineSize'.
+inlineSerialize :: Inline -> Config -> Config
+inlineSerialize v cfg = cfg {cfgInlineSerialize = v}
+
+-- | How should we inline the 'deserialize' function? See guidelines in
+-- 'inlineSize'.
+inlineDeserialize :: Inline -> Config -> Config
+inlineDeserialize v cfg = cfg {cfgInlineDeserialize = v}
+
+-- XXX We can identify if it is a sum type or a product type and use a single
```
adithyaov, line 111: We need two different options.
```
data A = B {..} | C {..}
```
Encoding 'B', and 'C' as String and encoding the record syntax in a different way are 2 independent options.

adithyaov:
Setters add one more layer of indirection which might make things cleaner and easier to use in a few cases. I wonder if it helps here.

adithyaov (approved):
LGTM

core/src/Streamly/Internal/Data/Serialize/TH/Bottom.hs
adithyaov: This case is currently unsupported though.

adithyaov: I'm removing this `XXX` comment.

PR #2616, opened 2023-11-21
Implement deriveSerialize using deriveSerializeWith

harendra-kumar:


adithyaov (approved):
Looks good, need to fixup.

PR #2631, opened 2023-11-28
Add API changelog for streamly-0.10.0 and streamly-core-0.2.0

harendra-kumar:


core/docs/ApiChangelogs/0.1.0-0.2.0.txt
```diff
@@ -0,0 +1,1471 @@
+---------------------------------
+API Annotations
+---------------------------------
+
+[A] : Added
+[R] : Removed
+[C] : Signature changed
```
adithyaov, line 7: Just "Changed" should work. The tag is present on modules as well. This is basically a tree and the node has `[C]` if any descendants of it are changed.

core/docs/ApiChangelogs/0.1.0-0.2.0.txt
```diff
@@ -0,0 +1,1471 @@
...
+[O] : Old signature
+[N] : New signature
```
adithyaov, line 9: `signature` is very specific to the context.

core/docs/ApiChangelogs/0.1.0-0.2.0.txt
```diff
@@ -0,0 +1,1471 @@
...
+[C] Streamly.Data.Stream
+    [A] wordsBy :: Monad m => (a -> Bool) -> Fold m a b -> Stream m a -> Stream m b
+    [A] splitOn :: Monad m => (a -> Bool) -> Fold m a b -> Stream m a -> Stream m b
+    [C] handle
+        [O] handle :: (MonadCatch m, Exception e) => (e -> Stream m a) -> Stream m a -> Stream m a
+        [N] handle :: (MonadCatch m, Exception e) => (e -> m (Stream m a)) -> Stream m a -> Stream m a
+    [A] groupsOf :: Monad m => Int -> Fold m a b -> Stream m a -> Stream m b
+[C] Streamly.Data.ParserK
+    [D] fromParser :: (MonadIO m, Unbox a) => Parser a m b -> ParserK (Array a) m b
+    [D] fromFold :: (MonadIO m, Unbox a) => Fold m a b -> ParserK (Array a) m b
```
adithyaov, line 56: If we deprecate `fromFold` we should have corresponding required parser combinators. Like, `drain`, `all`, `any` etc.
Why can't folds be converted to parsers, again?
`Fold initial step extract final` = `Parser initial step final`
Does the Fold applicative make sense anymore, given scanning is what we have in mind?

docs/User/Project/ApiChangelogs/0.9.0-0.10.0.txt
```diff
@@ -0,0 +1,424 @@
+---------------------------------
+API Annotations
+---------------------------------
+
+[A] : Added
+[R] : Removed
+[C] : Signature changed
+[O] : Old signature
+[N] : New signature
```
adithyaov, line 9: See comments above.

core/docs/ApiChangelogs/0.1.0-0.2.0.txt
harendra-kumar: instead of K.fromFold we use D.fromFold and `K.adapt` now.

PR #2642, opened 2023-12-19
Read dir performance improvements

harendra-kumar:


harendra-kumar:
Targeting this to next major release streamly-0.11/streamly-core-0.3 as breaking changes cannot be avoided.

PR #2651, opened 2023-12-27
Some fixes and updates to Array and Array stream modules

harendra-kumar:
* Some of these are updates required for better performance for directory traversal.
* Also, fixed a bug in pinnedFromList to generate pinned arrays.
* The rest of the changes are some cleanups

### DIFF
--- a/core/src/Streamly/Internal/Data/MutArray/Type.hs
+++ b/core/src/Streamly/Internal/Data/MutArray/Type.hs
@@ -181,6 +181,8 @@ module Streamly.Internal.Data.MutArray.Type
     , foldl'
     , foldr
     , cmp
+    , byteCmp
+    , byteEq
 
     -- ** Arrays of arrays
     --  We can add dimensionality parameter to the array type to get
@@ -2424,14 +2426,18 @@ asUnpinnedPtrUnsafe arr f =
 -- Equality
 -------------------------------------------------------------------------------
 
--- | Compare the length of the arrays. If the length is equal, compare the
--- lexicographical ordering of two underlying byte arrays otherwise return the
--- result of length comparison.
+-- | Byte compare two arrays. Compare the length of the arrays. If the length
+-- is equal, compare the lexicographical ordering of two underlying byte arrays
+-- otherwise return the result of length comparison.
+--
+-- /Unsafe/: Note that the 'Unbox' instance of sum types with constructors of
+-- different sizes may leave some memory uninitialized which can make byte
+-- comparison unreliable.
 --
 -- /Pre-release/
-{-# INLINE cmp #-}
-cmp :: MonadIO m => MutArray a -> MutArray a -> m Ordering
-cmp arr1 arr2 =
+{-# INLINE byteCmp #-}
+byteCmp :: MonadIO m => MutArray a -> MutArray a -> m Ordering
+byteCmp arr1 arr2 =
     liftIO
         $ do
             let marr1 = getMutableByteArray# (arrContents arr1)
@@ -2453,6 +2459,20 @@ cmp arr1 arr2 =
                               in (# s#, res #)
                     return $ compare r 0
                 x -> return x
+
+{-# INLINE cmp #-}
+{-# DEPRECATED cmp "Please use byteCmp instead." #-}
+cmp :: MonadIO m => MutArray a -> MutArray a -> m Ordering
+cmp = byteCmp
+
+-- | Byte equality of two arrays.
+--
+-- >>> byteEq arr1 arr2 = (==) EQ $ MArray.byteCmp arr1 arr2
+--
+-- /Unsafe/: See 'byteCmp'.
+{-# INLINE byteEq #-}
+byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
+byteEq arr1 arr2 = fmap (EQ ==) $ byteCmp arr1 arr2
 
 -------------------------------------------------------------------------------
 -- NFData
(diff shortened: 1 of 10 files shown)


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2042,6 +2049,32 @@ fromPureStream :: (MonadIO m, Unbox a) => Stream Identity a -> m (MutArray a)
...
+{-# INLINABLE fromPtrN #-}
+fromPtrN :: MonadIO m => Int -> Ptr Word8 -> m (MutArray Word8)
+fromPtrN len addr = do
+    -- memcpy is better than stream copy when the size is known.
+    -- XXX We can implement a stream copy in a similar way by streaming Word64
+    -- first and then remaining Word8.
+    arr <- new len
+    _ <- asUnpinnedPtrUnsafe arr
+            (\ptr -> liftIO $ c_memcpy ptr addr (fromIntegral len))
+    return (arr {arrEnd = len})
```
adithyaov, line 2076: Is it safe to use `c_memcpy` on unpinned arrays?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2379,6 +2412,12 @@ asPtrUnsafe arr f =
     Unboxed.asPtrUnsafe
         (arrContents arr) (\ptr -> f (ptr `plusPtr` arrStart arr))
 
+{-# INLINE asUnpinnedPtrUnsafe #-}
+asUnpinnedPtrUnsafe :: MonadIO m => MutArray a -> (Ptr a -> m b) -> m b
+asUnpinnedPtrUnsafe arr f =
+    Unboxed.asUnpinnedPtrUnsafe
+        (arrContents arr) (\ptr -> f (ptr `plusPtr` arrStart arr))
```
adithyaov, line 2465: Is this function safe to use?
Under what circumstances is this safe to use?

core/src/Streamly/Internal/Data/MutByteArray/Type.hs
```diff
@@ -120,6 +121,16 @@ asPtrUnsafe arr f = do
   liftIO $ touch contents
   return r
 
+-- | For use with unsafe FFI functions. Does not force pin the array memory.
+{-# INLINE asUnpinnedPtrUnsafe #-}
+asUnpinnedPtrUnsafe :: MonadIO m => MutByteArray -> (Ptr a -> m b) -> m b
+asUnpinnedPtrUnsafe arr f = do
+  let !ptr = Ptr (byteArrayContents#
+                     (unsafeCoerce# (getMutableByteArray# arr)))
```
adithyaov, line 129: How is this safe?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2088,15 +2087,19 @@ arrayStreamKLength as = K.foldl' (+) 0 (K.map length as)
 --
 {-# INLINE fromArrayStreamK #-}
 fromArrayStreamK :: (Unbox a, MonadIO m) =>
-    StreamK m (MutArray a) -> m (MutArray a)
-fromArrayStreamK as = do
+    PinnedState -> StreamK m (MutArray a) -> m (MutArray a)
+fromArrayStreamK ps as = do
```
adithyaov: You can give as `As` suffix. Keeping the convention.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2290,10 +2290,9 @@ spliceExp = spliceWith (\l1 l2 -> max (l1 * 2) (l1 + l2))
 {-# INLINE breakOn #-}
 breakOn :: MonadIO m
     => Word8 -> MutArray Word8 -> m (MutArray Word8, Maybe (MutArray Word8))
-breakOn sep arr@MutArray{..} = asPtrUnsafe arr $ \p -> liftIO $ do
-    -- XXX Instead of using asPtrUnsafe (pinning memory) we can pass unlifted
-    -- Addr# to memchr and it should be safe (from ghc 8.4).
```
adithyaov, line 2251: Aha. How?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2087,21 +2087,29 @@ arrayStreamKLength as = K.foldl' (+) 0 (K.map length as)
...
     K.foldlM' spliceUnsafe (pure arr) as
     -- fromStreamDN len $ D.unfoldMany reader $ D.fromStreamK as
 
+-- | Convert an array stream to an array. Note that this requires peak memory
+-- that is double the size of the array stream.
+--
+{-# INLINE fromArrayStreamK #-}
+fromArrayStreamK :: (Unbox a, MonadIO m) =>
+    StreamK m (MutArray a) -> m (MutArray a)
+fromArrayStreamK = fromArrayStreamKAs Unpinned
```
adithyaov, line 2140: :+1:

adithyaov:
This is still nasty: https://github.com/composewell/streamly/blob/97657098d080bb3b54be94c2a2ead93e0a1f2fd3/core/src/Streamly/Internal/Data/MutByteArray/Type.hs#L115C1-L115C1

We can possibly return the updated pointer or the updated array. Or at least mention in the docs that this is nasty.
Also the `XXX` comment can be removed as `asUnpinnedPtrUnsafe` is now implemented.

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: Read: https://downloads.haskell.org/ghc/latest/docs/users_guide/exts/ffi.html?highlight=ffi#guaranteed-call-safety

GHC, since version 8.4, guarantees that garbage collection will never occur during an unsafe call, even in the bytecode interpreter, and further guarantees that unsafe calls will be performed in the calling thread. Making it safe to pass heap-allocated objects to unsafe functions.

PR #2658, opened 2024-01-02
Deprecate MutArray/Stream module, move funcs to MutArray module

harendra-kumar:
Move the mut array compaction funcs from MutArray.Stream to MutArray module.

### DIFF
--- a/core/src/Streamly/Internal/Data/MutArray/Type.hs
+++ b/core/src/Streamly/Internal/Data/MutArray/Type.hs
@@ -216,14 +216,23 @@ module Streamly.Internal.Data.MutArray.Type
     , pinnedChunksOf
     , buildChunks
 
-    -- *** Flatten a stream of arrays
-    , concatChunks
-    , concatChunksRev
-
     -- *** Split an array into slices
     -- , getSlicesFromLenN
     , splitOn
     -- , slicesOf
+
+    -- *** Flatten a stream of arrays
+    , concatChunks
+    , concatChunksRev
+
+    -- *** Compaction
+    , SpliceState (..)
+    , pCompactChunksLE
+    , rCompactChunksLE
+    , fCompactChunksGE
+    , lCompactChunksGE
+    , compactChunksGE
+    , compactChunksEQ
 
     -- ** Utilities
     , roundUpToPower2
@@ -248,6 +257,7 @@ where
 
 import Control.Monad (when, void)
 import Control.Monad.IO.Class (MonadIO(..))
+import Data.Bifunctor (first)
 import Data.Bits (shiftR, (.|.), (.&.))
 import Data.Functor.Identity (Identity(..))
 import Data.Proxy (Proxy(..))
@@ -274,13 +284,17 @@ import GHC.Ptr (Ptr(..))
 import Streamly.Internal.Data.Fold.Type (Fold(..))
 import Streamly.Internal.Data.Producer.Type (Producer (..))
 import Streamly.Internal.Data.Stream.Type (Stream)
+import Streamly.Internal.Data.Parser.Type (Parser (..))
 import Streamly.Internal.Data.StreamK.Type (StreamK)
 import Streamly.Internal.Data.SVar.Type (adaptState, defState)
+import Streamly.Internal.Data.Tuple.Strict (Tuple'(..))
 import Streamly.Internal.Data.Unfold.Type (Unfold(..))
 import Streamly.Internal.System.IO (arrayPayloadSize, defaultChunkSize)
 
 import qualified Streamly.Internal.Data.Fold.Type as FL
 import qualified Streamly.Internal.Data.MutByteArray.Type as Unboxed
+import qualified Streamly.Internal.Data.Parser.Type as Parser
+-- import qualified Streamly.Internal.Data.Fold.Type as Fold
 import qualified Streamly.Internal.Data.Producer as Producer
 import qualified Streamly.Internal.Data.Stream.Type as D
 import qualified Streamly.Internal.Data.Stream.Lift as D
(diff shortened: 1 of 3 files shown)


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2582,7 +2596,240 @@ byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
...
--- NFData
+-- Compact
+-------------------------------------------------------------------------------
+
+-- | Parser @pCompactChunksLE n@ coalesces adjacent arrays in the input stream
+-- only if the combined size would be less than or equal to n.
+--
+-- /Internal/
+{-# INLINE_NORMAL pCompactChunksLE #-}
+pCompactChunksLE ::
```
adithyaov, line 2607: Does `p` here signify parser?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2582,7 +2596,240 @@ byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
...
+    step' _ (SpliceYielding arr next) = return $ D.Yield arr next
+
+-- | Fold @fCompactChunksGE n@ coalesces adjacent arrays in the input stream
+-- until the size becomes greater than or equal to n.
+--
+{-# INLINE_NORMAL fCompactChunksGE #-}
+fCompactChunksGE ::
+       forall m a. (MonadIO m, Unbox a)
+    => Int -> FL.Fold m (MutArray a) (MutArray a)
+fCompactChunksGE n = Fold step initial extract extract
```
adithyaov, line 2710: The name `Chunks` is already here. Can we move this to the fold module?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2582,7 +2596,240 @@ byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
...
+
+    extract Nothing = return nil
+    extract (Just buf) = return buf
+
+-- | Like 'compactChunksGE' but for transforming folds instead of stream.
+--
+-- >>> lCompactChunksGE n = Fold.many (MutArray.fCompactChunksGE n)
+--
+{-# INLINE_NORMAL lCompactChunksGE #-}
+lCompactChunksGE :: (MonadIO m, Unbox a)
```
adithyaov, line 2748: Can we move this to the fold module?

core/src/Streamly/Internal/Data/MutArray/Type.hs
adithyaov: Can we move this to the parser module? `Chunks` is already in the name.

adithyaov: If we want to keep it here, can we rename it to `compactChunksParserLE`? Although long it's explicit.
The prefix `p` is confusing.

core/src/Streamly/Internal/Data/MutArray/Type.hs
adithyaov: Rename it as it was renamed earlier. The prefix `f` is confusing.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2582,7 +2596,240 @@ byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
...
+-- instead of mutating it.
+
+-- | Scan @rCompactChunksLE n@ coalesces adjacent arrays in the input stream
+-- only if the combined size would be less than or equal to n.
+--
+-- /Internal/
+{-# INLINE_NORMAL rCompactChunksLE #-}
+rCompactChunksLE :: (MonadIO m, Unbox a)
+    => Int -> D.Stream m (MutArray a) -> D.Stream m (MutArray a)
+rCompactChunksLE n (D.Stream step state) =
```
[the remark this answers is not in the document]
adithyaov: Why do you call this a `Scan`?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2582,7 +2596,240 @@ byteEq :: MonadIO m => MutArray a -> MutArray a -> m Bool
...
+    | SpliceBuffering s arr
+    | SpliceYielding arr (SpliceState s arr)
+    | SpliceFinish
+
+-- This mutates the first array (if it has space) to append values from the
+-- second one. This would work for immutable arrays as well because an
+-- immutable array never has additional space so a new array is allocated
+-- instead of mutating it.
+
+-- | Scan @rCompactChunksLE n@ coalesces adjacent arrays in the input stream
```
adithyaov, line 2655: `Scan`?

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: it's been renamed to pCompactLE in a later PR. adding Parser/Fold suffix may be verbose especially if we follow this convention in more places.

core/src/Streamly/Internal/Data/MutArray/Type.hs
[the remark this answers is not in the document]
harendra-kumar: r => right, l => left (e.g. rMapM and lmap in fold module), used when we need to differentiate between covariant and contravariant style mapping.

In general, I use the term scan for any stateful transformation.

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: No. this is specific to an array type (e.g. Array, Array.Generic, MutArray, MutArray.Generic etc), you cannot have all these functions in Fold module. Also Fold module should not have functions for all types, the module for MutArray type should have folds operating on this type.

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: stateful transformation.

PR #2662, opened 2024-01-02
Add documentation to Array.clone and Array.pinnedClone

adithyaov:


core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -324,6 +324,19 @@ fromStreamD str = unsafeFreeze <$> MA.fromStreamD str
...
+-- >>> import Streamly.Internal.Data.MutByteArray as MBA
+-- >>> import Streamly.Internal.Data.Array as Array
+-- >>> array = Array.fromList ['a', 'b', 'c', 'd', 'e', 'f']
+-- >>> slice1 = Array.getSliceUnsafe 1 4 array
+-- >>> slice2 = Array.clone slice1
+-- >>> MBA.sizeOfMutableByteArray (Array.arrContents slice1)
+-- 6
+-- >>> MBA.sizeOfMutableByteArray (Array.arrContents slice2)
+-- 4
+--
```
harendra-kumar, line 339: We can pin or unpin a slice rather than cloning it.

adithyaov: Currently, `pin` on a pinned array is noop and `unpin` on an unpinned array is a noop.
`pin` on an unpinned array == `pinnedClone`
`unpin` on a pinned array == `clone`

Do we want `pin` and `unpin` to allocate array everytime?

harendra-kumar: The original question was why would you want to clone an immutable array?

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +745,100 @@ decodeUtf8Lax = decodeUtf8
...
+                    True -> ParserD.Continue 1 UTF16CharDecodeInit
+                    False -> ParserD.Continue 0 UTF16CharDecodeInit
+
+    {-# INLINE combineSurrogates #-}
+    combineSurrogates hi lo =
+        let first10 = fromIntegral (hi - utf16HighSurrogate) `shiftL` 10
+            second10 = fromIntegral (lo - utf16LowSurrogate)
+         in unsafeChr (0x10000 + (first10 .|. second10))
+
+    -- In little endian we should get low surrogate first.
```
adithyaov, line 782: Little endian does not matter here. We are dealing with `Word16` directly.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -965,13 +1114,52 @@ encodeUtf8Lax = encodeUtf8
...
+utf16LowSurrogate :: Word16
+utf16LowSurrogate = 0xDC00
+
+{-# INLINE utf16HighSurrogate #-}
+utf16HighSurrogate :: Word16
+utf16HighSurrogate = 0xD800
+
+{-# INLINE_NORMAL readCharUtf16With #-}
+readCharUtf16With :: Monad m => WList Word16 -> Unfold m Char Word16
+readCharUtf16With invalidReplavement = Unfold step inject
```
adithyaov, line 1127: `invalidReplacement` -- spelling.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +745,100 @@ decodeUtf8Lax = decodeUtf8
...
+         | x < 0xD800 = pure $ ParserD.Done 0 $ unsafeChr $ fromIntegral x
+         | x > 0xDFFF = pure $ ParserD.Done 0 $ unsafeChr $ fromIntegral x
+         | x >= 0xD800 && x <= 0xDBFF = -- high surrogate
+               pure $ ParserD.Partial 0 $ UTF16CharDecoding x
+         | otherwise =
+               let msg = prefix ++ "Invalid first UTF16 word " ++ show x
+                in pure $ handleError msg False
+
+    -- We backtrack here as there is a chance the word in the buffer will
+    -- combine with the word after the current word.
```
harendra-kumar, line 795: We don't want that. A local problem with this one word will propagate in the entire stream by shifting the stream by one word. We should just drop the char.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +745,100 @@ decodeUtf8Lax = decodeUtf8
...
+         | x >= 0xDC00 && x <= 0xDFFF = -- low surrogate
+               pure $ ParserD.Done 0 $ combineSurrogates prev x
+         | otherwise =
+               let msg = prefix
+                       ++ "Invalid subsequent UTF16 word " ++ show x
+                       ++ " in state " ++ show prev
+                in pure $ handleError msg True
+
+    {-# INLINE extract #-}
+    extract UTF16CharDecodeInit =  error $ prefix ++ "Not enough input"
```
harendra-kumar, line 806: We should use Error rather than error, so that the parser can handle the case.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +745,100 @@ decodeUtf8Lax = decodeUtf8
...
+    extract (UTF16CharDecoding _) =
+        case cfm of
+            ErrorOnCodingFailure ->
+                return $ ParserD.Error $ prefix ++ "Not enough input"
+            TransliterateCodingFailure ->
+                return (ParserD.Done 0 replacementChar)
+            -- XXX We shouldn't error out here. There is no way to represent an
+            -- empty parser result unless we return a "Maybe" type.
+            DropOnCodingFailure -> error $ prefix ++ "Not enough input"
```
harendra-kumar, line 815: Should use parser Error.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +745,100 @@ decodeUtf8Lax = decodeUtf8
...
 --
--- /Unimplemented/
-{-# INLINE decodeUtf16le' #-}
-decodeUtf16le' :: Stream m Word16 -> Stream m Char
-decodeUtf16le' = undefined
+{-# INLINE decodeUtf16' #-}
+decodeUtf16' :: Monad m => Stream m Word16 -> Stream m Char
+decodeUtf16' =
+    fmap fromEither
+        . Stream.parseMany (parseCharUtf16With ErrorOnCodingFailure)
```
harendra-kumar, line 825: How do you handle empty stream? Do we have a test for that? The parser would error out. You can try alternative parser but need to check perf.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -134,6 +144,59 @@ import Prelude hiding (lines, words, unlines, unwords)
...
+-- Word16 to Word8 helpers
+-------------------------------------------------------------------------------
+
+data ToW16leState s x
+    = T16Nothing !s
+    | T16Just !s !x
+
+{-# INLINE scanWord16le #-}
+scanWord16le :: Monad m => Stream m Word8 -> Stream m Word16
+scanWord16le (D.Stream step state) = D.Stream step1 (T16Nothing state)
```
harendra-kumar, line 157: We can have two separate functions, one to convert Word8 stream to Word16, another one to swap the byte order. Because a lot of times it is more efficient to generate Word16 stream directly from arrays rather than combining words.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -134,6 +144,59 @@ import Prelude hiding (lines, words, unlines, unwords)
...
+                Yield
+                    ((fromIntegral x `shiftL` 8) .|. fromIntegral prev)
+                    (T16Nothing st1)
+            Skip st1 -> Skip (T16Just st1 prev)
+            -- XXX This stop condition should be configurable. Currently there
+            -- is loss of information.
+            Stop   -> Stop
+
+{-# INLINE splitWord16le #-}
+splitWord16le :: Monad m => Stream m Word16 -> Stream m Word8
```
harendra-kumar, line 182: Use a Word16 stream byte-swap function. And a separate converter to Word8 stream. Most of the times we can directly write the Word16 stream to arrays which would be more efficient rather than converting it to Wor8 first.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -965,13 +1114,52 @@ encodeUtf8Lax = encodeUtf8
...
+
+{-# INLINE_NORMAL readCharUtf16With #-}
+readCharUtf16With :: Monad m => WList Word16 -> Unfold m Char Word16
+readCharUtf16With invalidReplavement = Unfold step inject
+
+    where
+
+    inject c =
+        return $ case ord c of
+            x | x <= 0xD7FF -> fromIntegral x `WCons` WNil
```
harendra-kumar, line 1133: Should use the same numbers everywhere e.g. you used 0xD800 earlier, here also we should use `x < 0xD800`.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -965,13 +1114,52 @@ encodeUtf8Lax = encodeUtf8
...
+              | x >= 0xE000 && x <= 0xFFFF -> fromIntegral x `WCons` WNil
```
harendra-kumar, line 1134: Use `x > 0xDFFF` here to keep it consistent with the value 0xDFFF used elsewhere. or change that.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: We do the same in utf8 encoding though. Might need to change it there as well?

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: This applies to UTF8 as well then

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: This applies to UTF8 as well then

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: Ah, this comment is actually incorrect. We backtrack and restart parsing. We drop the element in the buffer.

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: Hmm. We cannot use a parser then. This should be some sort of a stateful transformation.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +695,224 @@ decodeUtf8Lax = decodeUtf8
...
+                    let len = Array.length arr
+                     in if (len .&. 1) == 1
+                        then let arr1 = Array.getSliceUnsafe 1 (len - 1) arr
+                                 fstElem = Array.getIndexUnsafe 0 arr
+                                 w16 = (remElem, fstElem)
+                              in Yield (Just w16, arr1) (Nothing, st1)
+                        else let arr1 = Array.getSliceUnsafe 1 (len - 2) arr
+                                 fstElem = Array.getIndexUnsafe 0 arr
+                                 lstElem = Array.getIndexUnsafe (len - 1) arr
+                                 w16 = (remElem, fstElem)
```
adithyaov, line 734: This will fail if the array is empty.
1. We need a test case that catches that.
2. Fix it.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +695,224 @@ decodeUtf8Lax = decodeUtf8
...
+        r <- step (adaptState gst) st
+        return $
+            case r of
+                Yield arr st1 ->
+                    let len = Array.length arr
+                     in if (len .&. 1) == 1
+                        then let arr1 = Array.getSliceUnsafe 0 (len - 1) arr
+                                 remElem = Array.getIndexUnsafe (len - 1) arr
+                              in Yield (Nothing, arr1) (Just remElem, st1)
+                        else Yield (Nothing, arr) (Nothing, st1)
```
adithyaov, line 717: This will fail if the array is empty.
1. We need a test case that catches that.
2. Fix it.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +695,234 @@ decodeUtf8Lax = decodeUtf8
...
+
+    {-# INLINE_LATE step1 #-}
+    step1 gst (Nothing, st) = do
+        r <- step (adaptState gst) st
+        return $
+            case r of
+                Yield arr st1 | Array.length arr == 0 -> Skip (Nothing, st1)
+                Yield arr st1 | Array.length arr == 1 ->
+                    let remElem = Array.getIndexUnsafe 0 arr
+                     in Skip (Just remElem, st1)
```
adithyaov, line 714: This is actually not required.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,254 @@ decodeUtf8Lax = decodeUtf8
...
+    step1 _ (SFCInner st _ i end) | i == end = pure $ Skip $ SFCOuter st
+    step1 _ (SFCInner st arrC i end) =
+        pure $
+            if hasSameByteOrdering
+            then
+                let !v = unsafeInlineIO $ Unbox.peekAt i arrC
+                 in Yield v (SFCInner st arrC (i + 2) end)
+            else
+                let !w1 = unsafeInlineIO $ Unbox.peekAt i arrC
+                    !w2 = unsafeInlineIO $ Unbox.peekAt (i + 1) arrC
```
harendra-kumar, line 814: Just read the Word16 and swap its first and second byte.

```
swapByteOrder :: Word16 -> Word16
swapByteOrder w = (w `shiftL` 8) .|. (w `shiftR` 8)
```

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,254 @@ decodeUtf8Lax = decodeUtf8
...
+    => ByteOrdering -- ^ Byte orderdering of the machine where the value is
+                    -- being deserialized.
+    -> ByteOrdering -- ^ Byte orderdering of the machine where the value is
+                    -- serialized.
+    -> Stream m (Array Word8)
+    -> Stream m Word16
+scanW16FromChunks dBO sBO strm =
+    scanW16FromEvenChunks dBO sBO (mkEvenW8Chunks strm)
+
+-- | Create a Word16 Stream from Word8 chunks with little endian byte ordering.
```
adithyaov, line 838: Create a Word16 stream from Word8 chunks interpreted in little-endian byte ordering.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,254 @@ decodeUtf8Lax = decodeUtf8
...
+    = SFCOuter s
+    | SFCInner s arr i end
+
+{-# INLINE_NORMAL scanW16FromEvenChunks #-}
+scanW16FromEvenChunks
+    :: Monad m
+    => ByteOrdering -- ^ Byte orderdering of the machine where the value is
+                    -- being deserialized.
+    -> ByteOrdering -- ^ Byte orderdering of the machine where the value is
+                    -- serialized.
```
harendra-kumar, line 772: Just pass a `Bool` whether to swap the order or not.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,229 @@ decodeUtf8Lax = decodeUtf8
 -- Decoding Utf16
 -------------------------------------------------------------------------------
 
--- | Decode a UTF-16 little endian encoded bytestream to a stream of Unicode
--- characters. The function throws an error if an invalid codepoint is
--- encountered.
+{-# INLINE_NORMAL mkEvenW8Chunks #-}
+mkEvenW8Chunks
+    :: Monad m
+    => Stream m (Array Word8) -> Stream m (Maybe (Word8, Word8), Array Word8)
```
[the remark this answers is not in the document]
harendra-kumar: We just need to carry over at most one Word8 from the previous array. If we have a carry over word, we create a new array of two Word8 and combine the carry over with the first word of the next array. The even length of the remaining array is output as it is, if the remaining length is odd we have a carry over again.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Stream.hs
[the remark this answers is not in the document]
adithyaov: Yeah, this is what I wanted to do but left it as is as it was an internal API. I'll make this change and we can merge it post that.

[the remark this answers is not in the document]
adithyaov: I was thinking in an imperative way. This seemed like the straightward approach.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,226 @@ decodeUtf8Lax = decodeUtf8
...
+        r <- step (adaptState gst) st
+        return $
+            case r of
+                Yield arr st1 | Array.length arr == 0 -> Skip (MECSInit st1)
+                Yield arr st1 | Array.length arr == 1 ->
+                    let remElem = Array.getIndexUnsafe 0 arr
+                     in Skip (MECSBuffer remElem st1)
+                Yield arr st1 ->
+                    let len = Array.length arr
+                     in if (len .&. 1) == 1
```
harendra-kumar, line 726: Use `odd` instead.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Unicode/Stream.hs
adithyaov: `odd` uses `div`

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -682,14 +700,226 @@ decodeUtf8Lax = decodeUtf8
...
+                                 w16 = Array.fromList [remElem, fstElem]
+                              in Yield w16 (MECSYieldAndInit arr1 st1)
+                        else let arr1 = Array.getSliceUnsafe 1 (len - 2) arr
+                                 fstElem = Array.getIndexUnsafe 0 arr
+                                 lstElem = Array.getIndexUnsafe (len - 1) arr
+                                 w16 = Array.fromList [remElem, fstElem]
+                              in Yield w16
+                                     (MECSYieldAndBuffer arr1 lstElem st1)
+                Skip s -> Skip (MECSBuffer remElem s)
+                Stop -> Stop -- Here the last Word8 is lost
```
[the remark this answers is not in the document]
adithyaov: This has nothing to do with utf16 decoding.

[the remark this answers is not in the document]
adithyaov: This is only used in the benchmark for now anyway, so I don't want to spend more time on this. We can change this if this use-case ever arises.

core/src/Streamly/Internal/Unicode/Stream.hs
harendra-kumar: hmm..that's odd.

core/src/Streamly/Internal/Unicode/Stream.hs
```diff
@@ -28,7 +28,12 @@ module Streamly.Internal.Unicode.Stream
     , decodeUtf8
     , decodeUtf8'
     , decodeUtf8_
+
+    -- ** UTF-16 Decoding
+    , decodeUtf16'
+    , decodeUtf16
```
adithyaov, line 34: We can keep these unexposed. Don't know if we'll ever use this without taking byte-order into consideration.

core/src/Streamly/Internal/Unicode/Stream.hs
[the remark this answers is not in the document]
harendra-kumar: If it has nothing to do with it then we should not use it in decoding :-) Because decoding should fail if it finds the input is truncated. If this drops silently, decoder won't even know that some part of the input was dropped, it will just happily use whatever it got.

benchmark/Streamly/Benchmark/Unicode/Stream.hs
```diff
@@ -258,6 +259,17 @@ _copyStreamUtf8' inh outh =
      $ Unicode.decodeUtf8'
      $ Stream.unfold Handle.reader inh
 
+-- | Copy file
+{-# NOINLINE copyStreamUtf16 #-}
+copyStreamUtf16 :: Handle -> Handle -> IO ()
+copyStreamUtf16 inh outh =
+   Stream.fold (Handle.writeChunks outh)
+     $ fmap Array.castUnsafe $ Array.chunksOf (arrayPayloadSize (16 * 1024))
```
harendra-kumar, line 267: We can use defaultChunkSize instead.

harendra-kumar (approved):

PR #2668, opened 2024-01-05
Add serialization/deserialization using Unbox

harendra-kumar:


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -491,9 +501,9 @@ newBytesAs ps bytes = do
         , arrBound = bytes
         }
 
--- | Allocates a pinned empty array that can hold 'count' items.  The memory of
--- the array is uninitialized and the allocation is aligned as per the
--- 'Unboxed' instance of the type.
+-- | Allocates a pinned empty array that with a reserved capacity of bytes.
```
adithyaov, line 504:
```suggestion
-- | Allocates a pinned empty array with a reserved capacity of bytes.
```

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -1051,6 +1061,126 @@ snoc = snocWith f
...
 
+-------------------------------------------------------------------------------
+-- Serialization/Deserialization
+-------------------------------------------------------------------------------
+
+{-# INLINE pokeNewEnd #-}
+pokeNewEnd :: (MonadIO m, Unbox a) =>
+    Int -> MutArray Word8 -> a -> m (MutArray Word8)
+pokeNewEnd newEnd arr@MutArray{..} x = liftIO $ do
+    assert (newEnd <= arrBound) (return ())
```
adithyaov, line 1072: You poke at the end of the array but set the end according to the input param?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -1051,6 +1061,126 @@ snoc = snocWith f
...
         else roundUpToPower2 oldSize * 2
 
+-------------------------------------------------------------------------------
+-- Serialization/Deserialization
+-------------------------------------------------------------------------------
+
+{-# INLINE pokeNewEnd #-}
+pokeNewEnd :: (MonadIO m, Unbox a) =>
+    Int -> MutArray Word8 -> a -> m (MutArray Word8)
+pokeNewEnd newEnd arr@MutArray{..} x = liftIO $ do
```
adithyaov, line 1071: The `Int` here is in bytes.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -1051,6 +1061,126 @@ snoc = snocWith f
...
+{-# INLINE pokeAppendUnsafe #-}
+pokeAppendUnsafe :: forall m a. (MonadIO m, Unbox a) =>
+    MutArray Word8 -> a -> m (MutArray Word8)
+pokeAppendUnsafe arr@MutArray{..} = pokeNewEnd (arrEnd + SIZE_OF(a)) arr
+
+-- | Skip the specified number of bytes in the array. The data in the skipped
+-- region remains uninitialzed.
+{-# INLINE pokeSkipUnsafe #-}
+pokeSkipUnsafe :: Int -> MutArray Word8 -> MutArray Word8
+pokeSkipUnsafe n arr@MutArray{..} =  do
```
adithyaov, line 1091: `n` is in bytes.
Should we use,
```
newtype Bytes = Bytes Int
```
Or rename this to 
`pokeSkipBytesUnsafe`?

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -1051,6 +1061,126 @@ snoc = snocWith f
...
+    else return Nothing
+
+{-# NOINLINE pokeWithRealloc #-}
+pokeWithRealloc :: forall m a. (MonadIO m, Unbox a) =>
+       (Int -> Int)
+    -> MutArray Word8
+    -> a
+    -> m (MutArray Word8)
+pokeWithRealloc sizer arr x = do
+    arr1 <- liftIO $ reallocWith "pokeWithRealloc" sizer (SIZE_OF(a)) arr
```
adithyaov, line 1115: Just saw the implementation of `reallocWith`. Seems incorrect to me. We check the length instead of capacity.

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: it is a `Word8` array, there is no other possibility.

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: looks ok. The sizing is based on the old length and not the old capacity. The doc also says the same.

PR #2669, opened 2024-01-06
Rename nil, new, write, writeAppend APIs in arrays

harendra-kumar:


core/src/Streamly/Data/MutArray.hs
```diff
@@ -88,6 +88,12 @@ module Streamly.Data.MutArray
 
     -- * Deprecated
     , newPinned
+    , new
+    , pinnedNew
+    , writeN      -- drop new
+    , write       -- full buffer
+    , writeAppendN -- XXX Rename to writeSnocN?
+    , writeAppend -- XXX Rename to writeSnoc?
```
adithyaov, line 96: We can remove the comments.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -40,50 +40,48 @@ module Streamly.Internal.Data.MutArray.Type
...
+    -- , new -- uninitialized array of specified length
 
     -- *** Cloning
     , clone
     , pinnedClone
 
     -- *** Slicing
     -- | Get a subarray without copying
-    , getSliceUnsafe
+    , getSliceUnsafe -- XXX unsafeGetSlice
```
adithyaov, line 66: We can use `Unsafe` as a suffix over a prefix.
`Unsafe` is auxiliary information, not the primary.

adithyaov (approved):

PR #2672, opened 2024-01-08
Rename write to create in array modules

harendra-kumar:


core/src/Streamly/Data/Array/Generic.hs
```diff
@@ -15,34 +15,37 @@ module Streamly.Data.Array.Generic
...
 
     -- * Construction
-    , A.fromListN
-    , A.fromList
+    , fromListN
+    , fromList
 
     -- MonadicAPIs
-    , A.writeN
-    , A.write
```
adithyaov, line 23: These are breaking changes? Is `Array.Generic` exposed?

adithyaov: Looks like it is. So the next release is going to be a major release?

adithyaov: Ahh. Nvm, `write` and `writeN` aren't deprecated.

core/src/Streamly/Internal/Data/Array/Generic.hs
```diff
@@ -116,9 +126,15 @@ writeWith elemCount = unsafeFreeze <$> MArray.createWith elemCount
 --
 -- /Caution! Do not use this on infinite streams./
 --
+{-# INLINE create #-}
+create :: MonadIO m => Fold m a (Array a)
+create = fmap unsafeFreeze MArray.create
+
+-- XXX Deprecate in major
+-- {-# DEPRECATED write "Please use create instead." #-}
```
adithyaov, line 134: Gotcha.

adithyaov (approved):

PR #2680, opened 2024-01-13
Make some changes to Path module, add absolute path check when appending

harendra-kumar:


core/src/Streamly/Internal/FileSystem/PosixPath.hs
```diff
@@ -0,0 +1,362 @@
...
+-- Maintainer  : streamly@composewell.com
+-- Portability : GHC
+--
+-- This module implements a OS_PATH type representing a file system path for
+-- OS_NAME operating systems. The only assumption about the encoding of the
+-- path is that it maps the characters SEPARATORS and @.@ to WORD_TYPE
+-- representing their ASCII values. Operations are provided to encode and
+-- decode using CODEC_NAME encoding.
+
+module Streamly.Internal.FileSystem.OS_PATH
```
adithyaov, line 36: Ah, we define `IS_WINDOWS`. Was confusing.

core/src/Streamly/Internal/FileSystem/PosixPath.hs
```diff
@@ -0,0 +1,362 @@
...
+
+import qualified Streamly.Internal.Data.Stream as Stream
+import qualified Streamly.Internal.FileSystem.Path.Common as Common
+import qualified Streamly.Internal.Unicode.Stream as Unicode
+
+import Language.Haskell.TH
+import Language.Haskell.TH.Quote
+import Streamly.Internal.Data.Path
+
+-- XXX docspec does not process CPP
```
adithyaov, line 105: Isn't CPP a step above `docspec`?

core/src/Streamly/Internal/FileSystem/PosixPath.hs
```diff
@@ -0,0 +1,362 @@
...
+------------------------------------------------------------------------------
+
+-- XXX rather have a dropEmptySegments?
+
+-- | If the path is @//@ the result is @/@. If it is @a//@ then the result is
+-- @a@.
+{-# INLINE dropTrailingSeparators #-}
+dropTrailingSeparators :: OS_PATH -> OS_PATH
+dropTrailingSeparators (OS_PATH arr) =
+    OS_PATH (Common.dropTrailingSeparators Common.Posix arr)
```
adithyaov, line 191: You have `Common.Posix` here even when `IS_WINDOWS` is defined. Will this work?

adithyaov: I'm guessing this should be `Common.OS_NAME`

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -0,0 +1,368 @@
...
+-- Chunk where we do not need to create an array.
+fromChars :: (MonadThrow m, Unbox a) =>
+       (Char -> Bool)
+    -> (Stream Identity Char -> Stream Identity a)
+    -> Stream Identity Char
+    -> m (Array a)
+fromChars p encode s =
+    -- XXX on windows terminate at first invalid char
+    let lengths = Fold.tee Fold.length (Fold.takeEndBy_ p Fold.length)
+        (n, n1) = runIdentity $ Stream.fold lengths s
```
adithyaov, line 140: `n == n1` iff the stream is completely consumed. Ie. no NULL char.
Is that right?

core/src/Streamly/Internal/FileSystem/PosixPath/FileDir.hs
```diff
@@ -0,0 +1,178 @@
...
+
+    -- Cannot have "." or ".." as last component.
+    fromPath p@(OS_PATH arr) = do
+        s1 <-
+                Stream.toList
+                    $ Stream.take 3
+                    $ Stream.takeWhile (not . Common.isSeparator OS_NAME)
+                    $ fmap Common.wordToChar
+                    $ Array.readRev arr
+        -- XXX On posix we just need to check last 3 bytes of the array
```
adithyaov, line 95: Why last 3? Not last 2?

core/src/Streamly/Internal/FileSystem/PosixPath/FileDir.hs
```diff
@@ -0,0 +1,178 @@
...
+        case s1 of
+            '.' : xs ->
```
adithyaov, line 97: Also, what if this ends with `/`. That way `s1` is an empty list.
Eg, `abc/def/`

adithyaov: We parse that as a correct file path?

core/src/Streamly/Internal/FileSystem/Path.hs
```diff
@@ -1,515 +1,194 @@
...
+-- basic path type which can represent a file, directory, absolute or relative
+-- path with no restrictions. Depending on how much type safety you want, you
+-- can choose appropriate type wrappers or a combination of those to wrap the
+-- 'Path' type.
+--
+-- The basic type-safety is provided by the
+-- "Streamly.Internal.FileSystem.PosixPath.LocSeg" module. We make a distinction
+-- between two types of paths viz. locations and segments. Locations are
+-- represented by the @Loc Path@ type and path segments are represented by the
+-- @Seg Path@ type. Locations are paths pointintg to specific objects in the
```
adithyaov, line 23:
```suggestion
-- @Seg Path@ type. Locations are paths pointing to specific objects in the
```

core/src/Streamly/Internal/FileSystem/Path.hs
```diff
@@ -1,515 +1,194 @@
...
+-- file system absolute or relative e.g. @\/usr\/bin@, @.\/local\/bin@, or @.@.
+-- Segments are a sequence of path components without any reference to a
+-- location e.g. @usr\/bin@, @local\/bin@, or @../bin@ are segments. This
```
adithyaov, line 26: A location is like an initial point and a segment is a route to some other place.

The examples are confusing. You can probably remove the examples. From a relative path, you cannot tell if it's a location or a segment. Location or a segment is what you already define. So the examples have no point?

core/src/Streamly/Internal/FileSystem/Path.hs
```diff
@@ -1,515 +1,194 @@
...
+-- an implicit root e.g. "./bin" is known as a relative location.
 --
--- /Unimplemented/
-absdir :: QuasiQuoter
-absdir = undefined
-
--- | Generates an @Rel (Dir Path)@ type from an interpolated string literal.
+-- A path that does not refer to a particular location but defines steps to go
+-- from some place to another is a path segment. For example, "local/bin" is a
+-- path segment whereas "./local/bin" is a location.
```
adithyaov, line 100: Ah, in the above doc, `../bin` is given as an example for a segment.
Why can't `./local/bin` be a segment?
Is `.` treated differently from `..`?

core/src/Streamly/Internal/FileSystem/Path.hs
```diff
@@ -1,515 +1,194 @@
...
--- /Unimplemented/
-mkRel :: String -> Q Exp
-mkRel = undefined
-
--- | Generates an @Dir Path@ type.
+-- @./bin@ may refer to a different location depending on what "." is
+-- referring to. Thus we should not allow @./bin@ to be appended to another
+-- path, @bin@ can be appended though. Similarly, we cannot compare @./bin@
+-- with @./bin@ and say that they are equal because they may be referring to
+-- different locations depending on in what context the paths were created.
```
adithyaov, line 122: The same can be said for `..`?

adithyaov (commented):
Reviewed.

core/src/Streamly/Internal/FileSystem/PosixPath.hs
adithyaov: Oh NVM. It parses the file I think.

core/src/Streamly/Internal/FileSystem/Path.hs
harendra-kumar: You can always tell. A location points to specific place, a location either starts with `/` or with `./` everything else is a segment. A location is either relative to root or explicitly relative to the current directory. `local/bin` is not relative to anything, it is a segment. `../bin` is not relative to anything, it is a segment.

core/src/Streamly/Internal/FileSystem/Path.hs
harendra-kumar: `./local/bin` points to a specific place i.e. `local/bin` in the current directory. `../bin` is not relative to anything, it is a segment but `./../bin` would be a location as it is relative to current directory.

core/src/Streamly/Internal/FileSystem/Path.hs
harendra-kumar: No. `..` means go one node up from wherever it is used, it is not necessarily from the current directory e.g. `a/../b`. it is always used as a way to specify a segment and never a location.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
harendra-kumar: yes.

core/src/Streamly/Internal/FileSystem/PosixPath.hs
harendra-kumar: This is not right, should be fixed.

core/src/Streamly/Internal/FileSystem/PosixPath/FileDir.hs
harendra-kumar: `a..` is a valid file but `/..` is not.

core/src/Streamly/Internal/FileSystem/PosixPath/FileDir.hs
harendra-kumar: `OS_PATH` cannot end with `/` we strip that. This assumes that we have a valid `OS_PATH`.

PR #2706, opened 2024-01-28
Some refactoring in the ParserK module

harendra-kumar:


core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -487,17 +491,24 @@ adaptWith
     -> Input a
     -> m (Step a m r)
 adaptWith pstep initial extract cont !relPos !usedCount !input = do
+    -- We ensure that this does not happen
+#ifdef DEBUG
+    when (relPos /= 0) $ error $ "adaptWith: entry relPos " ++ show relPos
```
adithyaov, line 501: Why can't it be `0`? Does that not mean "retry the current element"?

core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -512,30 +523,34 @@ adaptWith pstep initial extract cont !relPos !usedCount !input = do
...
 
         go !_ !pst = do
-            pRes <- pstep pst x
-            case pRes of
+            r <- pstep pst x
+            case r of
+                -- Done, call the next continuation
                 ParserD.Done 0 b ->
-                    cont (Success 1 b) (count + 1) (Chunk x)
+                    pure $ Partial 1 (cont (Success 0 b) (count + 1))
```
adithyaov, line 535: Can having backtrack on `Partial` only do the job?
Ie. Is it possible to remove the backtrack from `Result`

core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -512,30 +523,34 @@ adaptWith pstep initial extract cont !relPos !usedCount !input = do
...
+                -- Not done yet, return the parseCont continuation
                 ParserD.Partial 0 pst1 ->
                     pure $ Partial 1 (parseCont (count + 1) pst1)
                 ParserD.Partial 1 pst1 ->
-                    -- XXX Since we got Partial, the driver should drop the
-                    -- buffer, we should call the driver here?
+                    -- XXX recurse or call the driver?
                     go SPEC pst1
-                ParserD.Partial n pst1 ->
+                ParserD.Partial n pst1 -> -- n > 0
```
adithyaov, line 547: Why `n > 0`?

core/src/Streamly/Internal/Data/ParserK/Type.hs
adithyaov, line 501: I did not understand this statement.
Should `relPos` always be `0` here?

harendra-kumar: We think that it should be 0, but we are not taking the risk of being wrong. The code is written to handle even if it is not 0. The debug statement will tell us if the assumption is wrong.

core/src/Streamly/Internal/Data/ParserK/Type.hs
harendra-kumar: That's exactly what we are doing. Result has relative position. When we have backtracked in Partial, now the relative position becomes 0 (`Success 0`).

adithyaov: No, I mean. Can we remove the `Int` from `Success`? If it's always 0, we can remove it.

core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -487,17 +491,24 @@ adaptWith
...
-                    Chunk arr -> parseContChunk usedCount pst arr
+                    -- In element parser case chunk is just one element
+                    Chunk element -> parseContChunk usedCount pst element
                     None -> parseContNothing usedCount pst
             -- XXX Previous code was using Continue in this case
-            else pure $ Partial relPos (parseCont usedCount pst)
+            else
+                -- We consumed previous input, need to fetch the next
+                -- input from the driver.
+                pure $ Partial relPos (parseCont usedCount pst)
```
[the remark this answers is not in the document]
harendra-kumar: Is this a statement or is there an issue or question here?

This is ParserK.Partial - it moves as per position index. Positive means move forward, negative means move backward. This code has not changed and it seems correct. If relPos is 1 we move to the next element, because index 1 is the end in this parser.

PR #2749, opened 2024-06-11
Bump required dependencies and add a CI for ghc-9.10

adithyaov:


harendra-kumar (commented):
There are too many CIs now, we need to reduce some redundancy.

harendra-kumar (approved):

PR #2754, opened 2024-06-26
Docspec fix

adithyaov:


src/Streamly/Data/Fold/Prelude.hs
```diff
@@ -55,7 +55,7 @@ import Streamly.Internal.Data.IsMap.HashMap ()
 --
 -- >>> classify = Fold.toHashMapIO fst (Fold.lmap snd Fold.toList)
 -- >>> Stream.fold classify input :: IO (HashMap String [Double])
--- fromList [("k2",[2.0,2.2]),("k1",[1.0,1.1])]
+-- fromList [("k1",[1.0,1.1]),("k2",[2.0,2.2])]
```
harendra-kumar, line 58: For predictable results we will have to sort the output list on the key. Otherwise you will keep flip-flopping this every time it fails.

.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ jobs:
             command: |
               bash -c "$PACKCHECK cabal"
         - *save
-  cabal-ghc-9_8_1-docspec:
+  cabal-ghc-9_4_8-docspec:
```
harendra-kumar, line 166: Was it working earlier with 9.8? When did it start failing?

adithyaov: No, I was just playing around with the versions. The docspec is still failing.

PR #1408, opened 2022-01-03
Replace Ring with the experimental implmenetation

adithyaov:
```
Generating reports for Prelude.Serial...
Prelude.Serial(cpuTime)
Benchmark                                                                             default(0)(ms) default(1) - default(0)(%)
------------------------------------------------------------------------------------- -------------- --------------------------
All.Prelude.Serial/o-1-space.split.S.splitOnSeq 100k long pattern                             263.06                     +17.60
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "catcatcatcatcat" FL.drain                    257.32                     +14.48
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "abcdefghi" FL.drain                          259.26                     +13.36
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "abcdefgh" FL.drain                           139.17                     +13.13
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "abcdefghijklmnopqrstuvwxyz" FL.drain         258.78                     +13.02
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "\r\n" FL.drain                               109.71                      +6.98
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "\n" FL.drain                                 140.03                      +2.54
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "" FL.drain                                    47.94                      -2.45
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "aaaa" FL.drain                               140.05                      -4.54
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "a" FL.drain                                  124.01                     -12.03
All.Prelude.Serial/o-1-space.split.S.splitOnSeq "aa" FL.drain                                 137.54                     -27.76
```

PR #2761, opened 2024-07-04
Use MutableByteArray as the primitive for the ring buffer

adithyaov:
This looks suspiciously good.

```
Generating reports for Data.Ring.Unboxed...
Data.Ring.Unboxed(cpuTime)
Benchmark                                      default(0)(ms) default(1) - default(0)(%)
---------------------------------------------- -------------- --------------------------
All.Data.Ring.Unboxed/o-1-space.unsafeEqArray            5.55                    -566.29
All.Data.Ring.Unboxed/o-1-space.unsafeEqArrayN           1.01                    -960.54

Data.Ring.Unboxed(Allocated)
Benchmark                                      default(0)(MiB) default(1) - default(0)(%)
---------------------------------------------- --------------- --------------------------
All.Data.Ring.Unboxed/o-1-space.unsafeEqArray            24.33                  -Infinity
All.Data.Ring.Unboxed/o-1-space.unsafeEqArrayN            1.60                  -Infinity

Data.Ring.Unboxed(maxrss)
Benchmark                                      default(0)(MiB) default(1) - default(0)(%)
---------------------------------------------- --------------- --------------------------
All.Data.Ring.Unboxed/o-1-space.unsafeEqArray            15.00                       0.00
All.Data.Ring.Unboxed/o-1-space.unsafeEqArrayN           15.00                       0.00
```

```
Generating reports for Data.Array...
Data.Array(cpuTime)
Benchmark                                                        default(0)(μs) default(1) - default(0)(%)
---------------------------------------------------------------- -------------- --------------------------
All.Data.Array/o-n-space.elimination.writeLastN.Max                     1894.99                      -4.93
All.Data.Array/o-1-space.elimination.writeLastN.10                        88.07                      -6.87
All.Data.Array/o-1-space.elimination.writeLastN.1                         94.66                     -39.35

Data.Array(Allocated)
Benchmark                                                        default(0)(Bytes) default(1) - default(0)(%)
---------------------------------------------------------------- ----------------- --------------------------
All.Data.Array/o-1-space.elimination.writeLastN.1                             0.00                        NaN
All.Data.Array/o-1-space.elimination.writeLastN.10                            0.00                        NaN
All.Data.Array/o-n-space.elimination.writeLastN.Max                    15233357.00                     -12.77

Data.Array(maxrss)
Benchmark                                                        default(0)(MiB) default(1) - default(0)(%)
---------------------------------------------------------------- --------------- --------------------------
All.Data.Array/o-n-space.elimination.writeLastN.Max                        17.00                       0.00
All.Data.Array/o-1-space.elimination.writeLastN.10                         15.00                       0.00
All.Data.Array/o-1-space.elimination.writeLastN.1                          15.00                       0.00
```

This PR supersedes:
- https://github.com/composewell/streamly/pull/1364
- https://github.com/composewell/streamly/pull/1408

### DIFF
--- a/core/src/Streamly/Internal/Data/Fold/Combinators.hs
+++ b/core/src/Streamly/Internal/Data/Fold/Combinators.hs
@@ -228,14 +228,13 @@ import Data.Either (isLeft, isRight, fromLeft, fromRight)
 import Data.Int (Int64)
 import Data.Proxy (Proxy(..))
 import Data.Word (Word32)
-import Foreign.Storable (Storable, peek)
+import Streamly.Internal.Data.Unbox (Unbox(..))
 import Streamly.Internal.Data.MutArray.Type (MutArray(..))
 import Streamly.Internal.Data.Maybe.Strict (Maybe'(..), toMaybe)
 import Streamly.Internal.Data.Pipe.Type (Pipe (..))
 import Streamly.Internal.Data.Scan (Scan (..))
 import Streamly.Internal.Data.Stream.Type (Stream)
 import Streamly.Internal.Data.Tuple.Strict (Tuple'(..), Tuple3'(..))
-import Streamly.Internal.Data.Unbox (Unbox, sizeOf)
 import Streamly.Internal.Data.Unfold.Type (Unfold(..))
 
 import qualified Prelude
@@ -1564,7 +1563,7 @@ data SplitOnSeqState acc a rb rh w ck =
 --
 -- /Pre-release/
 {-# INLINE takeEndBySeq #-}
-takeEndBySeq :: forall m a b. (MonadIO m, Storable a, Unbox a, Enum a, Eq a) =>
+takeEndBySeq :: forall m a b. (MonadIO m, Unbox a, Enum a, Eq a) =>
        Array.Array a
     -> Fold m a b
     -> Fold m a b
@@ -1590,8 +1589,8 @@ takeEndBySeq patArr (Fold fstep finitial fextract ffinal) =
                 | SIZE_OF(a) * patLen <= sizeOf (Proxy :: Proxy Word) ->
                     return $ Partial $ SplitOnSeqWord acc 0 0
                 | otherwise -> do
-                    (rb, rhead) <- liftIO $ Ring.new patLen
-                    return $ Partial $ SplitOnSeqKR acc 0 rb rhead
+                    rb <- liftIO $ Ring.new patLen
+                    return $ Partial $ SplitOnSeqKR acc 0 rb 0
             Done b -> return $ Done b
 
     -- Word pattern related
@@ -1664,7 +1663,7 @@ takeEndBySeq patArr (Fold fstep finitial fextract ffinal) =
                 rh1 <- liftIO $ Ring.unsafeInsert rb rh x
                 if idx == maxIndex
                 then do
-                    let fld = Ring.unsafeFoldRing (Ring.ringBound rb)
+                    let fld = Ring.unsafeFoldRing (Ring.ringCapacity rb)
                     let !ringHash = fld addCksum 0 rb
                     if ringHash == patHash && Ring.unsafeEqArray rb rh1 patArr
                     then Done <$> ffinal s1
@@ -1676,7 +1675,7 @@ takeEndBySeq patArr (Fold fstep finitial fextract ffinal) =
         res <- fstep s x
         case res of
             Partial s1 -> do
-                old <- liftIO $ peek rh
+                (old :: a) <- liftIO $ PEEK_ELEM(a, rh, (Ring.ringContents rb))
                 rh1 <- liftIO $ Ring.unsafeInsert rb rh x
                 let ringHash = deltaCksum cksum old x
                 if ringHash == patHash && Ring.unsafeEqArray rb rh1 patArr
@@ -1704,7 +1703,7 @@ takeEndBySeq patArr (Fold fstep finitial fextract ffinal) =
 -- /Pre-release/
 --
 {-# INLINE takeEndBySeq_ #-}
-takeEndBySeq_ :: forall m a b. (MonadIO m, Storable a, Unbox a, Enum a, Eq a) =>
+takeEndBySeq_ :: forall m a b. (MonadIO m, Unbox a, Enum a, Eq a) =>
        Array.Array a
     -> Fold m a b
     -> Fold m a b
@@ -1731,8 +1730,8 @@ takeEndBySeq_ patArr (Fold fstep finitial fextract ffinal) =
                 | SIZE_OF(a) * patLen <= sizeOf (Proxy :: Proxy Word) ->
                     return $ Partial $ SplitOnSeqWord acc 0 0
                 | otherwise -> do
-                    (rb, rhead) <- liftIO $ Ring.new patLen
-                    return $ Partial $ SplitOnSeqKR acc 0 rb rhead
+                    rb <- liftIO $ Ring.new patLen
+                    return $ Partial $ SplitOnSeqKR acc 0 rb 0
             Done b -> return $ Done b
 
     -- Word pattern related
@@ -1804,14 +1803,14 @@ takeEndBySeq_ patArr (Fold fstep finitial fextract ffinal) =
         rh1 <- liftIO $ Ring.unsafeInsert rb rh x
         if idx == maxIndex
         then do
-            let fld = Ring.unsafeFoldRing (Ring.ringBound rb)
+            let fld = Ring.unsafeFoldRing (Ring.ringCapacity rb)
             let !ringHash = fld addCksum 0 rb
             if ringHash == patHash && Ring.unsafeEqArray rb rh1 patArr
             then Done <$> ffinal s
             else return $ Partial $ SplitOnSeqKRLoop s ringHash rb rh1
         else return $ Partial $ SplitOnSeqKR s (idx + 1) rb rh1
     step (SplitOnSeqKRLoop s cksum rb rh) x = do
-        old <- liftIO $ peek rh
+        old <- liftIO $ PEEK_ELEM(a, rh, (Ring.ringContents rb))
         res <- fstep s old
         case res of
             Partial s1 -> do
@@ -1841,7 +1840,7 @@ takeEndBySeq_ patArr (Fold fstep finitial fextract ffinal) =
                 if n == 0
                 then fex s
                 else do
-                    old <- liftIO $ peek rh
+                    old <- liftIO $ PEEK_ELEM(a, rh, (Ring.ringContents rb))
                     let rh1 = Ring.advance rb rh
                     r <- fstep s old
                     case r of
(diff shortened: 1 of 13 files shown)


adithyaov:
I need to add some more sanity tests before merging this.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -530,16 +479,15 @@ unsafeFoldRingFullM rh f z rb@Ring {..} =
 -- Note, this will crash on ring of 0 size.
 --
 {-# INLINE unsafeFoldRingNM #-}
-unsafeFoldRingNM :: forall m a b. (MonadIO m, Storable a)
-    => Int -> Ptr a -> (b -> a -> m b) -> b -> Ring a -> m b
-unsafeFoldRingNM count rh f z rb@Ring {..} =
-    withForeignPtrM ringStart $ \_ -> go count z rh
+unsafeFoldRingNM :: forall m a b. (MonadIO m, Unbox a)
```
harendra-kumar, line 482: Fix the docs.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -469,39 +428,30 @@ unsafeEqArray Ring{..} rh A.Array{..} =
...
-    withForeignPtrM ringStart $ \x -> go z x ptr
+unsafeFoldRingM :: forall m a b. (MonadIO m, Unbox a)
+    => Int -> (b -> a -> m b) -> b -> Ring a -> m b
+unsafeFoldRingM len f z Ring {..} = go z 0 len
   where
     go !acc !start !end
         | start == end = return acc
         | otherwise = do
-            let !x = unsafeInlineIO $ peek start
+            let !x = unsafeInlineIO $ PEEK_ELEM(a, start, ringContents)
```
harendra-kumar, line 452: No need to use unsafeInlineIO as we have MonadIO constraint.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -511,13 +461,12 @@ unsafeFoldRingM ptr f z Ring {..} =
...
-    => Ptr a -> (b -> a -> m b) -> b -> Ring a -> m b
-unsafeFoldRingFullM rh f z rb@Ring {..} =
-    withForeignPtrM ringStart $ \_ -> go z rh
+unsafeFoldRingFullM :: forall m a b. (MonadIO m, Unbox a)
+    => Int -> (b -> a -> m b) -> b -> Ring a -> m b
+unsafeFoldRingFullM rh f z rb@Ring {..} = go z rh
   where
     go !acc !start = do
-        let !x = unsafeInlineIO $ peek start
+        let !x = unsafeInlineIO $ PEEK_ELEM(a, start, ringContents)
```
harendra-kumar, line 469: Avoid unsafeInlineIO.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -530,16 +479,15 @@ unsafeFoldRingFullM rh f z rb@Ring {..} =
...
+unsafeFoldRingNM :: forall m a b. (MonadIO m, Unbox a)
+    => Int -> Int -> (b -> a -> m b) -> b -> Ring a -> m b
+unsafeFoldRingNM count rh f z rb@Ring {..} = go count z rh
 
     where
 
     go 0 acc _ = return acc
     go !n !acc !start = do
-        let !x = unsafeInlineIO $ peek start
+        let !x = unsafeInlineIO $ PEEK_ELEM(a, start, ringContents)
```
harendra-kumar, line 490: Avoid unsafeInlineIO.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -583,17 +531,15 @@ slidingWindowWith n (Fold step1 initial1 extract1 final1) =
     step (Tuple4' rb rh i st) a
         | i < n = do
             rh1 <- liftIO $ unsafeInsert rb rh a
-            liftIO $ touchForeignPtr (ringStart rb)
-            let action = toArray unsafeFoldRingM rb (PTR_NEXT(rh, a))
+            let action = toArray unsafeFoldRingM rb rh1
```
harendra-kumar, line 534: BUG: We have perhaps introduced a subtle bug here by not being equivalent to the old code. The old code was doing PTR_NEXT(rh) rather than using rh1 and that was for a good reason I guess. If we are at the last element of the ring, rh1 would become 0 and we would not fold anything.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -111,68 +106,38 @@ import Prelude hiding (length, concat, read)
...
-
-    elemSize = STORABLE_SIZE_OF(a)
-    ringStartPtr = unsafeForeignPtrToPtr ringStart
-    lenInBytes = ringBound `minusPtr` ringStartPtr
-    offInBytes = ringHead `minusPtr` ringStartPtr
-    len = assert (lenInBytes `mod` elemSize == 0) $ lenInBytes `div` elemSize
-    off = assert (offInBytes `mod` elemSize == 0) $ offInBytes `div` elemSize
-    advanceFromHead = (off + by `mod` len) * elemSize
+moveBy :: Int -> Ring a -> Int -> Int
+moveBy by rb ringHead = (ringHead + by) `mod` ringCapacity rb
```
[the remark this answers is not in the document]
adithyaov: In that case, this would become a recursive function.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -399,28 +361,25 @@ cast arr =
...
-    w8Contents = arrContents
-
     check p i = do
-        (relem :: Word8) <- peek p
-        aelem <- peekAt i w8Contents
+        (relem :: Word8) <- peekAt p ringContents
+        aelem <- peekAt i arrContents
         if relem == aelem
-        then go (p `plusPtr` 1) (i + 1)
+        then go (p + 1) (i + 1)
```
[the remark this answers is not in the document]
adithyaov: No, this is correct. We are checking byte by byte. Not element by element. 
This is exactly like the previous code.
I did not want to introduce any semantic changes.
There must have been a reason we did not check this element by element in the previous code.
Possibly because deserialization of the entire element was expensive maybe?

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -435,31 +394,31 @@ unsafeEqArrayN Ring{..} rh A.Array{..} nBytes
...
```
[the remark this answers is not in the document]
adithyaov: Not a bug. This is correct. Even I was confused looking at it initially but this replicates the previous implementation without any semantic changes.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -469,39 +428,30 @@ unsafeEqArray Ring{..} rh A.Array{..} =
...
 {-# INLINE unsafeFoldRing #-}
-unsafeFoldRing :: forall a b. Storable a
-    => Ptr a -> (b -> a -> b) -> b -> Ring a -> b
-unsafeFoldRing ptr f z Ring{..} =
-    let !res = unsafeInlineIO $ withForeignPtr ringStart $ \p ->
-                    go z p ptr
+unsafeFoldRing :: forall a b. Unbox a
+    => Int -> (b -> a -> b) -> b -> Ring a -> b
+unsafeFoldRing len f z Ring{..} =
+    let !res = unsafeInlineIO $ go z 0 len
```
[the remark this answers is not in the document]
adithyaov: We don't need to pass `len`. Non semantically similar changes I'll probably do in another commit.

core/src/Streamly/Internal/Data/Fold/Combinators.hs
```diff
@@ -1676,7 +1675,7 @@ takeEndBySeq patArr (Fold fstep finitial fextract ffinal) =
         res <- fstep s x
         case res of
             Partial s1 -> do
-                old <- liftIO $ peek rh
+                (old :: a) <- liftIO $ PEEK_ELEM(a, rh, (Ring.ringContents rb))
```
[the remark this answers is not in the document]
adithyaov: I don't see any redundant brackets.
Are the brackets here `(old :: a)` redundant?
If so, need to check hlint.

core/src/Streamly/Internal/Data/Ring.hs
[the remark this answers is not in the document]
harendra-kumar: I missed the explicit typing to Word8. I thought we are using Unbox because of peekAt. We can remove the byte byte comparison and use memcmp instead. Can you create an issue for this?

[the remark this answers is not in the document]
adithyaov: https://github.com/composewell/streamly/issues/2764

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -531,7 +531,7 @@ slidingWindowWith n (Fold step1 initial1 extract1 final1) =
     step (Tuple4' rb rh i st) a
         | i < n = do
             rh1 <- liftIO $ unsafeInsert rb rh a
-            let action = toArray unsafeFoldRingM rb rh1
+            let action = toArray unsafeFoldRingM rb (rh + 1)
```
harendra-kumar, line 553: May be a good idea to add a comment here, so that we do not fall for it again.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -137,7 +147,7 @@ advance rb ringHead =
 -- array.
 {-# INLINE moveBy #-}
 moveBy :: Int -> Ring a -> Int -> Int
-moveBy by rb ringHead = (ringHead + by) `mod` ringCapacity rb
+moveBy by rb ringHead = (ringHead + by) `mod1` ringCapacity rb
```
harendra-kumar, line 153: usually we should not be moving it by more than the ring size. But we have to handle the case if someone does that. Anyway, it seems this function is not used anywhere.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -174,7 +175,6 @@ fromArray = undefined
 
 -- | Modify a given index of a ring array using a modifier function.
 --
--- /Unimplemented/
```
harendra-kumar, line 202: Why remove this?

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -184,11 +184,20 @@ modifyIndex = undefined
...
 -- >>> putIndex arr ix val = Ring.modifyIndex arr ix (const (val, ()))
 --
--- /Unimplemented/
 {-# INLINE putIndex #-}
-putIndex :: -- (MonadIO m, Unbox a) =>
-    Ring a -> Int -> a -> m ()
-putIndex = undefined
+putIndex :: forall m a. (MonadIO m, Unbox a) => Ring a -> Int -> a -> m ()
+putIndex ring i val =
+    liftIO $ pokeAt (normalizedI * SIZE_OF(a)) (ringContents ring) val
```
harendra-kumar, line 192: this should be `unsafePutIndex ring normalizedI val`

core/src/Streamly/Internal/Data/Ring.hs
adithyaov: Ah, I was going to implement it. Then decided against it.

core/src/Streamly/Internal/Data/Ring.hs
```diff
@@ -467,7 +467,7 @@ unsafeFoldRingM len f z rb = go z 0 len
     go !acc !start !end
         | start == end = return acc
         | otherwise = do
-            let !x = unsafeInlineIO $ unsafeGetIndex start rb
+            !x <- unsafeGetIndex start rb
```
harendra-kumar, line 471: We do not need strictness annotation anymore.

harendra-kumar (approved):

PR #2780, opened 2024-07-17
Multiple issue resolution targeting the 0.11 release

adithyaov:
**TODO**: Update the `Changelog` accordingly.

### DIFF
--- a/core/docs/Changelog.md
+++ b/core/docs/Changelog.md
@@ -2,10 +2,15 @@
 
 ## Unreleased
 
+* Add the following modules
+  - Streamly.Data.Scan
+  - Streamly.FileSystem.Path
+  - Streamly.FileSystem.Path.LocSeg
+  - Streamly.FileSystem.Path.FileDir
+  - Streamly.FileSystem.Path.Typed
 * Remove the `Storable` constraint from the following functions:
   - Streamly.Data.Stream.isInfixOf
   - Streamly.Data.Array.writeLastN
-
 * `Streamly.FileSystem.Dir` module is deprecated and replaced by
   `Streamly.FileSystem.DirIO` module. The new module has exact same
   APIs except that it uses the streamly native `Path` type instead
@@ -16,6 +21,18 @@
 
 * Remove the `Storable` constraint from several functions involving the ring
   buffer.
+
+### API Renaming
+
+* Rename `writeN`-like APIs to `createOf`-like in Array modules.
+* Rename `new`-like APIs to `emptyOf`-like in Array modules.
+* In the Fold module `indexGeneric`, `lengthGeneric`, and `foldlM1'` to
+   `genericIndex`, `genericLength`, and `foldl1M'` respectively.
+
+### Deprecations and API changes
+
+See [0.2.2-0.3.0 API Changelog](/core/docs/ApiChangelogs/0.2.2-0.3.0.txt) for a
+full list of deprecations, additions, and changes to the function signatures.
 
 ## 0.2.2 (Jan 2024)
 
--- a/core/src/Streamly/Data/Scan.hs
+++ b/core/src/Streamly/Data/Scan.hs
@@ -1,0 +1,29 @@
+-- |
+-- Module      : Streamly.Data.Scan
+-- Copyright   : (c) 2019 Composewell Technologies
+-- License     : BSD3
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+module Streamly.Data.Scan
+    (
+    -- * Type
+      Scan
+
+    -- * Primitive Scans
+    , identity
+    , function
+    , functionM
+    , filter
+    , filterM
+
+    -- * Combinators
+    , compose
+    , teeWithMay
+    , teeWith
+    )
+where
+
+import Streamly.Internal.Data.Scan
+import Prelude hiding (filter)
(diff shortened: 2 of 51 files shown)


core/docs/Changelog.md
```diff
@@ -2,15 +2,33 @@
...
 
+### API Renaming
+
+* Rename `writeN`-like APIs to `createOf`-like in Array modules.
+* Rename `new`-like APIs to `emptyOf`-like in Array modules.
+* Rename  `toFold` to `unsafeToFold` in the Parser module.
+* In the Fold module `indexGeneric`, `lengthGeneric`, and `foldlM1'` to
+   `genericIndex`, `genericLength`, and `foldl1M'` respectively.
+* In the scan module rename `map` and `mapM` to `function` and `functionM`
+  keeping it consistent with the naming in the Unfold module.
```
harendra-kumar, line 25: Scan module is new, this should not be in the changelog.

core/docs/Changelog.md
```diff
@@ -2,15 +2,33 @@
...
+
+### Deprecations and API changes
+
+See [0.3.0-0.2.2 API Changelog](/core/docs/ApiChangelogs/0.3.0-0.2.2.txt) for a
```
harendra-kumar, line 29: The naming convention is `lowerRelease-higherRelease`. See the 0.2.0 changelog.

core/src/Streamly/FileSystem/Path.hs
```diff
@@ -0,0 +1,212 @@
+-- |
+-- Module      : Streamly.FileSystem.Path
+-- Copyright   : (c) 2023 Composewell Technologies
+-- License     : BSD3
+-- Maintainer  : streamly@composewell.com
+-- Portability : GHC
+--
+-- = User Notes
```
harendra-kumar, line 8: Do not add the docs in the exposed modules, leave it in the internal modules for now. We have to write it better when moving it. Currently it is just random notes, and developer notes should anyway remain in the internal modules.

core/src/Streamly/Data/Scan.hs
```diff
@@ -0,0 +1,70 @@
+-- |
+-- Module      : Streamly.Data.Scan
+-- Copyright   : (c) 2019 Composewell Technologies
+-- License     : BSD3
+-- Maintainer  : streamly@composewell.com
+-- Stability   : experimental
+-- Portability : GHC
+--
+-- A Scan is half the pipe:
```
harendra-kumar, line 9: This documentation is totally internal.

core/src/Streamly/Internal/Data/Array/Generic.hs
```diff
@@ -146,7 +144,7 @@ fromPureStream x =
 fromCString# :: MonadIO m => Addr# -> m (Array Word8)
 fromCString# addr = fromStream $ D.fromCString# addr
 
-{-# DEPRECATED fromByteStr# "Please use fromCString instead" #-}
+{-# DEPRECATED fromByteStr# "Please use 'unsafeInlineIO . fromCString#' instead" #-}
```
harendra-kumar, line 147: unsafePerformIO, unsafeInlineIO is an internal function.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -2854,6 +2849,11 @@ castUnsafe (MutArray contents start end bound) =
 asBytes :: MutArray a -> MutArray Word8
 asBytes = castUnsafe
 
+-- | Cast an @MutArray Word8@ into an @MutArray a@.
+--
+fromBytes :: MutArray Word8 -> MutArray a
+fromBytes = castUnsafe
+
```
harendra-kumar, line 2856: why introduce this API? `castUnsafe` can be used directly.

core/src/Streamly/Internal/Data/Array/Generic.hs
```diff
@@ -118,9 +118,9 @@ createOf = fmap unsafeFreeze <$> MArray.createOf
 writeN :: MonadIO m => Int -> Fold m a (Array a)
 writeN = createOf
 
-{-# INLINE_NORMAL writeWith #-}
-writeWith :: MonadIO m => Int -> Fold m a (Array a)
-writeWith elemCount = unsafeFreeze <$> MArray.createWith elemCount
+{-# INLINE_NORMAL createWith #-}
+createWith :: MonadIO m => Int -> Fold m a (Array a)
+createWith elemCount = unsafeFreeze <$> MArray.createWith elemCount
```
harendra-kumar, line 123: How is this different from createOf? if it is different we should document it.

core/src/Streamly/Internal/Data/Array/Generic.hs
adithyaov: Then there should be a proper way to replicate the behaviour of `fromByteStr#`

core/src/Streamly/Internal/Data/MutArray/Type.hs
adithyaov: If I use `castUnsafe` I'll have to type it manually. Just like `asBytes`, `fromBytes` makes sense to have.
Especially since we are removing `newAsBytes`

core/src/Streamly/Internal/Data/Array/Generic.hs
adithyaov: `create = createWith (allocBytesToElemCount (undefined :: a) arrayChunkBytes)`

adithyaov: `createWith` is documented

adithyaov: Ah, it's documented in the `MutArray` module.

PR #2811, opened 2024-07-29
Make ParserD consistent with ParserK

adithyaov:
This needs to be reviewed very carefully.


core/src/Streamly/Internal/Data/Fold/Chunked.hs
```diff
@@ -150,7 +152,7 @@ fromParserD (ParserD.Parser step1 initial1 extract1) =
 
     step s (Array contents start end) = do
         if start >= end
-        then return $ Continue 0 s
+        then return $ Continue 1 s
```
adithyaov, line 155: This is incorrect. Revert this.

core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -748,21 +763,23 @@ toParser parser = ParserD.Parser step initial extract
...
 
     extract cont = do
         r <- cont None
         case r of
             -- This is extract so no input has been given, therefore, the
-            -- translation here is (0 - n) rather than (1 - n).
-            Done n b ->  assert (n <= 0) (return $ ParserD.Done (negate n) b)
+            -- translation here is (1 + n) rather than n.
+            -- XXX How should we visualize (1 + n)?
+            -- XXX Go forward by 1 element on extract?
```
adithyaov, line 777: Go back by n but consume the current element?
Is this correct? Should it be `n - 1` here, not `1 + n`?
How are the tests passing? Check if we test this.

adithyaov: Even though I put `n - 1` here, or `n + 1` here, the test cases are passing. We need to write a more elaborate test case for `fromParserK`

adithyaov: Ah, we have some fusion `RULES` due to which the test case isn't run like it should.

core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -1238,17 +1257,19 @@ adaptCWith pstep initial extract cont !offset0 !usedCount !input = do
...
             -- it is the backtrack element count and not the stream position
             -- index into the current input array.
-            ParserD.Done n b ->
-                assert (n >= 0)
-                    (cont (Success (- n) b) (count - n) None)
-            ParserD.Continue n pst1 ->
-                assert (n >= 0)
-                    (return $ Continue (- n) (parseCont (count - n) pst1))
+            ParserD.SDone m b ->
+                let n = 1 - m
```
harendra-kumar, line 1261: This looks incorrect based on the `IMPORTANT` note above. But why is no test failing?

core/src/Streamly/Internal/Data/Array/Generic.hs
```diff
@@ -631,17 +634,19 @@ adaptCGWith pstep initial extract cont !offset0 !usedCount !input = do
...
             -- it is the backtrack element count and not the stream position
             -- index into the current input array.
-            ParserD.Done n b ->
-                assert (n >= 0)
-                    (cont (Success (- n) b) (count - n) None)
-            ParserD.Continue n pst1 ->
-                assert (n >= 0)
-                    (return $ Continue (- n) (parseCont (count - n) pst1))
+            ParserD.SDone m b ->
+                let n = 1 - m
```
harendra-kumar, line 638: This looks incorrect based on the `IMPORTANT` note above. But why is no test failing?

core/src/Streamly/Internal/Data/Array.hs
adithyaov: Was the `IMPORTANT` note incorrect?

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -1218,7 +1218,19 @@ takeEndBy cond (Parser pstep pinitial pextract) =
         res <- pstep s a
         if not (cond a)
         then return res
-        else extractStep pextract res
+        else
```
adithyaov, line 1221: Was the previous code incorrect?

core/src/Streamly/Internal/Data/Parser.hs
```diff
@@ -2432,55 +2454,56 @@ takeP lim (Parser pstep pinitial pextract) = Parser step initial extract
...
-                        Continue n s1 -> Continue n (Tuple' (cnt1 - n) s1)
+                        SDone n b -> SDone n b
+                        SContinue n s1 -> SContinue n (Tuple' (cnt1 + n) s1)
                         Error err -> Error err
-                        Partial _ _ -> error "takeP: Partial in extract"
-            Partial n s -> do
-                let taken = cnt1 - n
+                        SPartial _ _ -> error "takeP: SPartial in extract"
+            SPartial n s -> do
+                let taken = cnt + n
```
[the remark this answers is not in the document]
adithyaov: Oh nvm. `cnt1 == cnt + 1`

adithyaov (commented):
1 bug. The rest looks good.

core/src/Streamly/Internal/Data/Stream/Eliminate.hs
```diff
@@ -290,23 +284,26 @@ parseBreakD (PRD.Parser pstep initial extract) stream@(Stream step state) = do
     goExtract !_ buf (List (x:xs)) !pst = do
         pRes <- pstep pst x
         case pRes of
-            PR.Partial 0 pst1 ->
+            PR.SPartial 1 pst1 ->
```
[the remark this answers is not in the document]
harendra-kumar: We are running `pstep pst x` above feeding it one element, therefore we need to take into account the element we fed. In goStop we are calling `extract` instead which is different.

PR #2815, opened 2024-07-30
Add a test case for ParserK.toParser

adithyaov:


core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -725,10 +725,12 @@ adaptCG (ParserD.Parser step initial extract) =
 -- | A continuation to extract the result when a CPS parser is done.
 {-# INLINE parserDone #-}
 parserDone :: Monad m => ParseResult b -> Int -> Input a -> m (Step a m b)
-parserDone (Success n b) _ None = return $ Done (negate n) b
-parserDone (Success n b) _ (Chunk _) = return $ Done (1 - n) b
-parserDone (Failure n e) _ None = return $ Error (negate n) e
-parserDone (Failure n e) _ (Chunk _) = return $ Error (1 - n) e
+parserDone (Success n b) _ _ = do
+    assertM(n <= 0)
```
adithyaov, line 729: Shouldn't `n >= 0`?

adithyaov: The `n` is `ParserK.Success` and the `n` in `ParserK.Done` should have opposite signs.
The `n` is `ParserK.Success` is similar to `n` in `ParserD.Done`

harendra-kumar: n in `Success n` is the index in the array which is same as `Done n`. See the comments in the definitions:

```
-- | The parser's result.
--
-- Int is the position index into the current input array. Could be negative.
-- Cannot be beyond the input array max bound.
--
-- /Pre-release/
--
data ParseResult b =
      Success !Int !b      -- Position index, result
    | Failure !Int !String -- Position index, error
```

```
data Step a m r =
    -- The Int is the current stream position index wrt to the start of the
    -- array.
      Done !Int r
    | Partial !Int (StepParser a m r)
    | Continue !Int (StepParser a m r)
    | Error !Int String
```

Issue #2699, opened 2024-01-25
Unreproducible failure in Unicode.Stream test suite


harendra-kumar:
On GHC-8.6.5, in a PR based on 5d096dfad

```
Failures:

  Streamly/Test/Unicode/Stream.hs:177:9: 
  1) Unicode.Stream, UTF8 - Encoding / Decoding, Arrays Streamly.Data.String.lines == Prelude.lines
       Assertion failed (after 766 tests):
      zipAsyncly Compose three - empty in the middle [✔]
         [48,89,237,128,185,236,161,51,229,204,167,5,154,15,252,80,71,15,179,204]

  To rerun use: --match "/Unicode.Stream/UTF8 - Encoding / Decoding/Arrays Streamly.Data.String.lines == Prelude.lines/" --seed 1380626044
        +++ OK, passed 100 tests.

Randomized with seed 1380626044
```

The relevant test is:

```
testLinesArray :: Property
testLinesArray =
    forAll genWord8 $ \list ->
        monadicIO $ do
            xs <- Stream.toList
                    $ fmap A.toList
                    $ AS.splitOnSuffix 10
                    $ Stream.fromPure (A.fromList list)
            assert (xs == map (map (fromIntegral . ord))
                              (lines (map (chr .  fromIntegral) list)))
```

Could be an issue with AS.splitOnSuffix?

```
{-# INLINE splitOnSuffix #-}
splitOnSuffix
    :: (MonadIO m)
    => Word8
    -> Stream m (Array Word8)
    -> Stream m (Array Word8)
splitOnSuffix = Array.compactOnByteSuffix
```

```
{-# INLINE compactOnByteSuffix #-}
compactOnByteSuffix
    :: (MonadIO m)
    => Word8
    -> Stream m (Array Word8)
    -> Stream m (Array Word8)
compactOnByteSuffix byte =
    fmap unsafeFreeze . MA.compactOnByteSuffix byte . fmap unsafeThaw
```

```
{-# INLINE compactOnByteSuffix #-}
compactOnByteSuffix
    :: (MonadIO m)
    => Word8
    -> Stream m (MutArray Word8)
    -> Stream m (MutArray Word8)
compactOnByteSuffix byte =
        -- XXX use spliceExp and rightSize?
        Stream.splitInnerBySuffix
            (\arr -> byteLength arr == 0) (breakOn byte) splice
```

`breakOn` uses unpinned pointer in memcpy, could it be because of that?

[...]

PR #2818, opened 2024-08-02
Fix unsafe use of FFI for unpinned arrays

harendra-kumar:
Pass the MutableByteArray# directly to the FFI functions to avoid garbage collector moving it between the time we get a pointer and then call the FFI function.

### DIFF
--- a/core/src/Streamly/Internal/Data/MutByteArray/Type.hs
+++ b/core/src/Streamly/Internal/Data/MutByteArray/Type.hs
@@ -12,7 +12,6 @@ module Streamly.Internal.Data.MutByteArray.Type
     (
     -- ** MutByteArray
       MutByteArray(..)
-    , MutableByteArray
     , getMutByteArray#
 
     -- ** Pinning
@@ -38,6 +37,7 @@ module Streamly.Internal.Data.MutByteArray.Type
     , unsafeAsPtr
 
     -- ** Deprecated
+    , MutableByteArray
     , getMutableByteArray#
     , newBytesAs
     , sizeOfMutableByteArray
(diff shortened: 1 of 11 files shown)


harendra-kumar:
Fixes #2699 .

core/src/Streamly/Internal/Data/MutByteArray/Type.hs
```diff
@@ -101,58 +102,46 @@ touch (MutByteArray contents) =
...
-asPtrUnsafe :: MonadIO m => MutByteArray -> (Ptr a -> m b) -> m b
-asPtrUnsafe = unsafePinnedAsPtr
-
--- | For use with unsafe FFI functions. Does not force pin the array memory.
 {-# INLINE unsafeAsPtr #-}
 unsafeAsPtr :: MonadIO m => MutByteArray -> (Ptr a -> m b) -> m b
 unsafeAsPtr arr f = do
-  let !ptr = Ptr (byteArrayContents#
+    -- XXX Make this an error instead?
+    assertM(isPinned arr)
```
adithyaov, line 137: You can make this an error here.
I'm very sure, this will be misused.

adithyaov: Why not pin it here, if unpinned?

adithyaov: Oh, we shouldn't pin it. It'll be a different array then. This will cause programmer confusion.
Can error out though.

core/src/Streamly/Internal/Data/MutByteArray/Type.hs
```diff
@@ -101,58 +102,46 @@ touch (MutByteArray contents) =
...
---
--- From the GHC manual:
---
--- GHC, since version 8.4, guarantees that garbage collection will never occur
--- during an unsafe call, even in the bytecode interpreter, and further
--- guarantees that unsafe calls will be performed in the calling thread. Making
--- it safe to pass heap-allocated objects to unsafe functions.
+-- | NOTE: this is deprecated because it can lead to accidental problems if the
+-- user tries to use it to mutate the array because it does not return the new
+-- array after pinning.
```
adithyaov, line 107: Aha, I was reading the changes from bottom-up hence the irrelevant comments later.

adithyaov (approved):

PR #2849, opened 2024-10-03
Consistent naming and behavior for split/compact operations

harendra-kumar:


core/src/Streamly/Internal/FileSystem/Handle.hs
```diff
@@ -456,7 +456,10 @@ chunkWriter = Refold.drainBy putChunk
 {-# INLINE writeChunksWith #-}
 writeChunksWith :: (MonadIO m, Unbox a)
     => Int -> Handle -> Fold m (Array a) ()
-writeChunksWith n h = A.lCompactGE n (writeChunks h)
+-- writeChunksWith n h = A.lCompactGE n (writeChunks h)
+writeChunksWith n h =
```
adithyaov, line 460: Review this.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -3323,40 +3395,110 @@ lCompactGeAs ps minElems (Fold step1 initial1 _ final1) =
...
     => Int -> Fold m (MutArray a) () -> Fold m (MutArray a) ()
 lPinnedCompactGE = lCompactGeAs Pinned
 
--- | @compactGE n stream@ coalesces adjacent arrays in the @stream@ until
--- the size becomes greater than or equal to @n@.
+data CompactMinState arr =
+    CompactMinInit | CompactMinIncomplete arr | CompactMinComplete arr
+
+{-# INLINE_NORMAL scanCompactMinAs #-}
+scanCompactMinAs :: forall m a. (MonadIO m, Unbox a)
```
adithyaov, line 3419: Review this!

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -3323,40 +3395,110 @@ lCompactGeAs ps minElems (Fold step1 initial1 _ final1) =
...
+    => PinnedState -> Int -> Scanl m (MutArray a) (Maybe (MutArray a))
```
adithyaov, line 3420: Instead of `Maybe (MutArray a)`, Why not have `Either (MutArray a) (MutArray a)`?
The use case being: we might have some error handling such as adding padding if `Left`.
Using this, is there a way to handle that?

adithyaov: The above statements apply to the last element.
Or the last accumulated chunk

PR #2851, opened 2024-10-04
Fix wordWithQuotes for a case where the string ends with an esc char

adithyaov:


test/Streamly/Test/Data/Parser.hs
```diff
@@ -1271,6 +1273,33 @@ takeStartBy_ =
             where
                 predicate = odd
                 parser = P.takeStartBy_ predicate FL.toList
+
+quotedWordTest :: String -> [String] -> IO ()
+quotedWordTest inp expected = do
+    res <-
+        S.fold FL.toList
```
harendra-kumar, line 1280: We have S.toList as well. We can use that.

harendra-kumar (approved):

PR #2859, opened 2024-10-07
Add an elaborate parser driver sanity tests

adithyaov:


core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -931,7 +931,7 @@ parseBreakChunksK (Parser pstep initial extract) stream = do
                 let n = Prelude.length backBuf
                     arr0 = fromListN n (Prelude.reverse backBuf)
                     arr1 = Array contents cur end
-                    str = StreamK.cons arr0 (StreamK.cons arr1 stream)
+                    str = StreamK.cons arr0 (StreamK.cons arr1 st)
```
adithyaov, line 934: On an error shouldn't you get the entire stream back?

harendra-kumar (approved):

PR #2869, opened 2024-10-15
Add File internals that work directly with Path

adithyaov:


adithyaov:
Ensure the change made is tested in the CIs

harendra-kumar (changes requested):
Change the module names as follows:

```
FileSystem.Windows.File => FileSystem.File.Window
FileSystem.Posix.File => FileSystem.File.Posix
```

core/src/Streamly/Internal/FileSystem/FileIO.hs
```diff
@@ -0,0 +1,682 @@
...
+    , fromChunks
+
+    -- ** Append To File
+    , writeAppend
+    , writeAppendWith
+    -- , appendShared
+    , writeAppendArray
+    , writeAppendChunks
+
+    -- * Deprecated
```
adithyaov, line 82: You can remove this section, as this is a new module.

core/src/Streamly/Internal/FileSystem/Posix/File.hsc
```diff
@@ -0,0 +1,318 @@
...
+
+##define MK_MODE_API(name1,name2,x) \
+{-# INLINE name1 #-}; \
+name1 :: FileMode -> FileMode; \
+name1 (FileMode mode) = FileMode (x .|. mode); \
+{-# INLINE name2 #-}; \
+name2 :: FileMode -> FileMode; \
+name2 (FileMode mode) = FileMode (x .|. mode)
+
+{-
```
adithyaov, line 118: Can we even comment CPP definitions?

core/src/Streamly/Internal/FileSystem/Posix/File.hsc
```diff
@@ -0,0 +1,318 @@
...
+#include <fcntl.h>
+
+-------------------------------------------------------------------------------
+-- Create mode
+-------------------------------------------------------------------------------
+
+-- | Open flags, see posix open system call man page.
+newtype FileMode = FileMode CMode
+
+##define MK_MODE_API(name1,name2,x) \
```
adithyaov, line 110: Is it `##define` and not `#define` because of `hsc`?

adithyaov (commented):
I've reviewed the PR. I've added some comments.

core/src/Streamly/Internal/FileSystem/FileIO.hs
harendra-kumar: I though it will be easier to replace the module if we keep them. But we can remove it too, because if someone is anyway making changes then they can change this as well.

core/src/Streamly/Internal/FileSystem/Posix/File.hsc
harendra-kumar: Yes, it was failing with an error and ##define worked.

core/src/Streamly/Internal/FileSystem/Posix/File.hsc
```diff
@@ -0,0 +1,318 @@
...
+{-# INLINE name1 #-}; \
+name1 :: FileMode -> FileMode; \
+name1 (FileMode mode) = FileMode (x .|. mode); \
+{-# INLINE name2 #-}; \
+name2 :: FileMode -> FileMode; \
+name2 (FileMode mode) = FileMode (x .|. mode)
```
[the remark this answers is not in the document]
harendra-kumar: Yes, that's a bug. copy, paste and forgot to fix.

core/src/Streamly/Internal/FileSystem/Posix/File.hsc
harendra-kumar: No, we are using the definitions directly from the header files rather than defining ourselves. I kept this code commented, just in case there is any issue in using the headers directly.

adithyaov: But if it did not error out because you've not used `##` it likely works as expected.

harendra-kumar: This code is actually commented, it does not work in any way.

PR #2872, opened 2024-10-17
Replace Enum usage for IoSubSystem with manully written functions

adithyaov:


.github/workflows/haskell.yml
```diff
@@ -95,6 +95,16 @@ jobs:
             cabal_project: cabal.project.ghc-head
             disable_sdist_build: "y"
             ignore_error: true
+          - name: 9.12.1-alpha
+            ghc_version: head
```
adithyaov, line 99: This `head` `ghc_version` has a special meaning in packcheck.

harendra-kumar (approved):

PR #2876, opened 2024-10-30
Change the representation of Generic.(Mut)Array to match (Mut)Array

adithyaov:


core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -423,9 +420,9 @@ reallocWith label sizer reqSize arr = do
 {-# INLINE unsafeSnoc #-}
 unsafeSnoc, snocUnsafe :: MonadIO m => MutArray a -> a -> m (MutArray a)
 unsafeSnoc arr@MutArray {..} a = do
-    assert (arrStart + arrLen < arrTrueLen) (return ())
-    let arr1 = arr {arrLen = arrLen + 1}
-    unsafePutIndex arrLen arr1 a
+    assert (arrEnd < arrCapacity) (return ())
+    let arr1 = arr {arrEnd = arrEnd + 1}
+    unsafePutIndex arrEnd arr1 a
```
harendra-kumar, line 425: BUG: arrEnd is not the index.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -479,9 +476,9 @@ snoc = snocWith (* 2)
 {-# INLINE uninit #-}
 uninit :: MonadIO m => MutArray a -> Int -> m (MutArray a)
 uninit arr@MutArray{..} len =
-    if arrStart + arrLen + len <= arrTrueLen
-    then return $ arr {arrLen = arrLen + len}
-    else realloc (arrLen + len) arr
+    if arrEnd + len <= arrCapacity
```
harendra-kumar, line 479: Looks like the old code was wrong. If the meaning of arrTrueLen was length instead of index.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -778,7 +784,7 @@ chunksOf n (D.Stream step state) =
                 putIndexUnsafeWith end contents x
                 let end1 = end + 1
                 return $
-                    if end1 >= bound
+                    if start + end1 >= bound
```
harendra-kumar, line 787: BUG: this should be `end1 >= bound`. end1 is the end index and bound is the max index + 1.

harendra-kumar (changes requested):

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -223,15 +223,11 @@ data MutArray a =
...
+        , arrEnd :: {-# UNPACK #-}!Int
           -- ^ The length of this slice.
 
-        , arrTrueLen :: {-# UNPACK #-}!Int
-          -- ^ This is the true length of the array. Coincidentally, this also
-          -- represents the first index beyond the maximum acceptable index of
-          -- the array. This is specific to the array contents itself and not
-          -- dependent on the slice. This value should not change and is shared
-          -- across all the slices.
+        , arrCapacity :: {-# UNPACK #-}!Int
```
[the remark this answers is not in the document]
adithyaov:
```
-- XXX rename to arrCapacity to be consistent with ring.
```
Was the comment in `Data.MutArray`.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -380,21 +376,22 @@ modifyIndex i arr@MutArray {..} f = do
...
         !(I# arr1S#) = arrStart arr1
         arrC# = arrContents# arr
         arr1C# = arrContents# arr1
+        !newEnd = arrStart arr1 + newLen
     liftIO
         $ IO
         $ \s# ->
               let s1# = copyMutableArray# arrC# arrS# arr1C# arr1S# newLen# s#
-               in (# s1#, arr1 {arrLen = newLen, arrTrueLen = n} #)
+               in (# s1#, arr1 {arrEnd = newEnd, arrCapacity = n} #)
```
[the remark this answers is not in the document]
adithyaov: `arrCapacity` does not depend on `arrStart`, it is always relative to `0`
`arrEnd` on the other hand will depend on `arrStart` and length.
Yes, `arrayStart` here is 0, but conceptually, what is written above is correct.
`arrCapacity = arrStart + n` would conceptually be incorrect.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
adithyaov: Why is the old code wrong? I did not understand.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
adithyaov: Looks like the previous code was bugged.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -223,15 +223,11 @@ data MutArray a =
         , arrStart :: {-# UNPACK #-}!Int
           -- ^ The starting index of this slice.
 
-        , arrLen :: {-# UNPACK #-}!Int
+        , arrEnd :: {-# UNPACK #-}!Int
           -- ^ The length of this slice.
```
harendra-kumar, line 227: Fix the comment.

core/src/Streamly/Internal/Data/MutArray/Generic.hs
```diff
@@ -421,12 +426,8 @@ reallocWith label sizer reqSize arr = do
...
 -- /Internal/
 {-# INLINE unsafeSnoc #-}
-unsafeSnoc, snocUnsafe :: MonadIO m => MutArray a -> a -> m (MutArray a)
-unsafeSnoc arr@MutArray {..} a = do
-    assert (arrStart + arrLen < arrTrueLen) (return ())
-    let arr1 = arr {arrLen = arrLen + 1}
-    unsafePutIndex arrLen arr1 a
-    return arr1
+snocUnsafe, unsafeSnoc :: MonadIO m => MutArray a -> a -> m (MutArray a)
+unsafeSnoc arr = snocNewEnd (arrEnd arr + 1) arr
```
harendra-kumar, line 430: would be better to expand snocNewEnd code right here instead of using that function.

harendra-kumar (approved):

PR #2897, opened 2024-12-09
Handle the unknown d_type field in posix readdir

adithyaov:


core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -163,6 +180,10 @@ readDirStreamEither (DirStream dirp) = loop
         -- fromPtrN, but it is not straightforward because the reclen is
         -- padded to 8-byte boundary.
         name <- Array.fromCString (castPtr dname)
+
+        when (dtype == #const DT_UNKNOWN)
+            $ throwIO (RDE_DT_UNKNOWN (mkPath name))
```
harendra-kumar, line 185: We cannot throw an exception, we need to use stat to find out the file type in this case. It will be less efficient but it will work. We cannot fail based on filesystem features.

harendra-kumar:
Need to check performance impact on ListDir example before merging.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -142,16 +142,31 @@ isMetaDir dname = do
...
-            -- EACCES.
-            _ -> throwErrno "checkIfDirectory"
+-- | Checks if dname is a directory and additionaly returns if dname is a meta
+-- directory.
+checkDirStatus
+    :: PosixPath -> Ptr CChar -> #{type unsigned char} -> IO (Bool, Bool)
+checkDirStatus parent dname dtype =
+    if dtype == #const DT_UNKNOWN
+    then do
+        isMeta <- liftIO $ isMetaDir dname
```
harendra-kumar, line 152: We can factor out this branch to a separate function and INLINE checkDirStatus.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -164,7 +164,7 @@ lstatDname parent dname = do
 checkDirStatus
     :: PosixPath -> Ptr CChar -> #{type unsigned char} -> IO (Bool, Bool)
 checkDirStatus parent dname dtype =
-    if dtype == #const DT_UNKNOWN
+    if True
```
harendra-kumar, line 167: Can we put this under a CPP macro, so that we can just enable/disable that for testing?

harendra-kumar (approved):

Issue #2804, opened 2024-07-27
Make demux and classify consistent


harendra-kumar:
Classify does not restart a fold after it terminates whereas demux does. We should perhaps use the classify behavior in demux as well. This can be done easily by installing a drain fold after a fold is done.

adithyaov:
Fix `demux` in both `Fold` and `Scan` module.
Their fold termination behaviour should be similar to classify.

harendra-kumar:
Looking at the implementation of `demuxGeneric` and `classifyGeneric` they should both have similar performance characteristics as the number of Map lookups in both is the same, and the rest of the code is same, in fact simpler in demux.

However, demux is more powerful combinator as the fold being used for a particular key can be modified dynamically. Thus the oneShot behaviour of `classify` which we have in the existing code can be emulated by `demux` by modifying the underlying state such that the next time getFold returns a different fold which ignores the input.

Since `classify` neither has a performance advantage nor a functionality advantage, it can possibly be left alone as it is today and may be deprecated later on or implemented in terms of `demux`. Maybe we can supply some idioms to emulate the behavior.

harendra-kumar:
We should verify the performance assumption of `demux` vs `classify` by comparing the benchmarks.

adithyaov:
See #2913

PR #2902, opened 2024-12-16
Make demux and classify consistent

adithyaov:
Closes: https://github.com/composewell/streamly/issues/2804


adithyaov:
- [ ] Add the previous functions as separate APIs

adithyaov:
See #2913

PR #2912, opened 2024-12-20
Make empty stream behavior of splitSepBySeq_ similar to splitSepBy_

adithyaov:
- [x] Check the benchmarks


test/Streamly/Test/Data/Stream.hs
```diff
@@ -403,7 +409,12 @@ intercalateSplitOnId x desc =
 groupSplitOps :: String -> Spec
 groupSplitOps desc = do
     -- splitting
-    splitOnSeq splitOnSeqFold
+
+    -- The foldManyPost implementation on an empty stream produces a single
+    -- value. The behaviour of foldManyPost implementation and the direct stream
+    -- implementation is not different.
```
adithyaov, line 415:
```suggestion
    -- implementation is different.
```

harendra-kumar (commented):
Can add some doctests in `splitSepBySeq_` and `splitEndBySeq` similar to splitSepBy_ ? Also enable the doctests in `splitBeginBy_`.

adithyaov:
> Also enable the doctests in splitBeginBy_.
`splitBeginBy` is undefined.

core/src/Streamly/Internal/Data/Stream/Nesting.hs
```diff
@@ -2892,6 +2941,23 @@ splitSepBySeq_ patArr (Fold fstep initial _ final) (Stream step state) =
...
+                if pat == x
+                then final fs >>= yieldReinit jump
+                else do
+                    r <- fstep fs x
+                    case r of
+                        FL.Partial fs1 ->
+                            pure $ Skip $ SplitOnSeqSingle fs1 s pat
+                        FL.Done b -> yieldReinit jump b
+            Skip s -> pure $ Skip $ SplitOnSeqSingle0 fs s pat
+            Stop -> final fs >> pure Stop
```
harendra-kumar, line 2959: See splitSepBy_ . We should have the same behavior. If we have not consumed any element we should not initialize the fold. We will be perform initialization and finalization unnecessarily.

harendra-kumar (changes requested):
Keep only interesting and minimal doctests for understanding the behavior. Move the rest to actual tests.

harendra-kumar (approved):

PR #2918, opened 2024-12-29
Path refactoring

harendra-kumar:


core/src/Streamly/Internal/Data/Stream/Type.hs
```diff
@@ -2114,25 +2115,27 @@ indexerBy (Fold step1 initial1 extract1 _final) n =
 
     extract (Tuple' i s) = (i,) <$> extract1 s
 
--- XXX rename to indicesEndBy
-
--- | Like 'splitEndBy' but generates a stream of (index, len) tuples marking
+-- | Like 'splitEndBy_' but generates a stream of (index, len) tuples marking
 -- the places where the predicate matches in the stream.
 --
+-- >>> Stream.toList $ Stream.indexEndBy_ (== '/') $ Stream.fromList "/home/harendra"
```
adithyaov, line 2121: `/home/harendra`
Now the world knows where your home directory is located.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -244,23 +252,26 @@ dropTrailingBy p arr@(Array barr start end) =
...
+------------------------------------------------------------------------------
+-- Drive parsing
+------------------------------------------------------------------------------
+
+-- | @C:...@, does not check array length.
+{-# INLINE unsafeHasDrive #-}
+unsafeHasDrive :: (Unbox a, Integral a) => Array a -> Bool
+unsafeHasDrive a
     -- Check colon first for quicker return
     | unsafeIndexChar 1 a /= ':' = False
```
adithyaov, line 276: `a /= b == False` is a complex way of saying `a == b`

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -269,63 +280,148 @@ hasDrive a
...
+------------------------------------------------------------------------------
+-- Relative or Absolute
+------------------------------------------------------------------------------
+
+-- | A path relative to cur dir it is either @.@ or starts with @./@.
+isRelativeCurDir :: (Unbox a, Integral a) => Array a -> Bool
+isRelativeCurDir a
     -- Assuming the path is not empty.
-    isSeparator Windows (wordToChar (Array.getIndexUnsafe 0 a))
+    | wordToChar (Array.getIndexUnsafe 0 a) /= '.' = False
```
adithyaov, line 299: This is unsafe. We don't check the length before indexing.

adithyaov: Can be written as `wordToChar (Array.getIndexUnsafe 0 a) == '.'`
It's more easier to understand.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -269,63 +280,148 @@ hasDrive a
...
-    | unsafeIndexChar 1 a /= ':' = False
-    | not (isSeparator Windows (unsafeIndexChar 2 a)) = False
-    -- XXX If we found a colon anyway this cannot be a valid path unless it has
-    -- a drive prefix. colon is not a valid path character.
-    -- XXX check isAlpha perf
-    | not (isAlpha (unsafeIndexChar 0 a)) = False
-    | otherwise = True
+-- | The path starting with a separator. On Windows this is relative to current
+-- drive while on Posix this is absolute path as there is only one drive.
+isRelativeCurDrive :: (Unbox a, Integral a) => OS -> Array a -> Bool
```
adithyaov, line 305: Can you add some examples for this?
`isRelativeCurDrive` is any path that starts with a seperator?

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -269,63 +280,148 @@ hasDrive a
...
+
+-- | Returns () if the path can be a valid file, otherwise throws an
+-- exception.
+maybeFile :: (MonadThrow m, Unbox a, Integral a) => OS -> Array a -> m ()
+maybeFile os arr = do
+    s1 <-
+            Stream.toList
+                $ Stream.take 3
+                $ Stream.takeWhile (not . isSeparator os)
+                $ fmap wordToChar
```
adithyaov, line 403: We are converting all words to characters.
If the input is UTF16 encoded, will all the Word16 translate into proper characters?
We should instead use `isSeperatorIntegral,` which checks the underlying `Word` instead of converting it into a `Char. '

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -269,63 +280,148 @@ hasDrive a
...
+                $ Array.readRev arr
+    -- XXX On posix we just need to check last 3 bytes of the array
+    -- XXX Display the path in the exception messages.
+    case s1 of
+        [] -> throwM $ InvalidPath "A file name cannot have a trailing separator"
```
adithyaov, line 408: Even an empty path could result in this. The error message does not indicate that.
So, `maybeFile` would fail if the path ends with
1. `.`
2. `..`

core/src/Streamly/Internal/FileSystem/Path/Common.hs
harendra-kumar: > This is unsafe. We don't check the length before indexing.

It assumes path is not empty. This assumption is there at some places already. If it is legal to have non-empty path can you raise an issue to fix these?

core/src/Streamly/Internal/FileSystem/Path/Common.hs
harendra-kumar: Empty path is not a valid path. 

> So, maybeFile would fail if the path ends with

yes.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
harendra-kumar: I do not see a real problem here.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
harendra-kumar: "/home/user" is relative to current drive.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
adithyaov: Ah, yes. The problem might arise when we are printing it. You can ignore this comment.

adithyaov (approved):

PR #2919, opened 2024-12-29
Add splitPath operation

harendra-kumar:


core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -463,3 +484,42 @@ append :: (Unbox a, Integral a) =>
...
+-- components in the path are ignored except in the leading position. Multiple
+-- consecutive separators are ignored.
+--
+-- >>> :{
+--  splitPath Common.Posix = Stream.toList . fmap unpackPosix . Common.splitPath Common.Posix . packPosix
+--  splitPath Common.Windows = Stream.toList . fmap unpackWindows . Common.splitPath Common.Windows . packWindows
+-- :}
+--
+-- >>> splitPath Common.Posix "."
+-- ["."]
```
adithyaov, line 498: Do the doctests pass.
How do we keep the leading `.`?
We are filtering them out right?

Perhaps `off /= 0` should be a part of `shouldFilterOut`
```
    shouldFilterOut (off, len) =
        len == 0 ||
            (len == 1 && unsafeIndexChar off arr == '.' && off /= 0)
```

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -463,3 +538,284 @@ append :: (Unbox a, Integral a) =>
...
+-- ["/","home"]
+--
+-- >>> splitPosix "home//user/./..////\\directory/."
+-- ["home","user","..","\\directory"]
+--
+-- >>> splitWin "home//user/./..////\\directory/."
+-- ["home","user","..","directory"]
+--
+{-# INLINE splitPath #-}
+splitPath
```
adithyaov, line 785: `splitPath` will ignore the trailing separator.
Is there a loss of information here?

adithyaov: `a/b/` is the same as `a/b/.`
`a/b` is different from `a/b/`

harendra-kumar: Ideally the components should not have separators. We can keep the trailing slash but that would require special casing it and not sure how useful that will be.

PR #2921, opened 2025-01-04
Fix FingerPrint serialization for ghc 8.6.5

adithyaov:


adithyaov:
The FingerPrint test works as expected.

PR #2950, opened 2025-02-05
Rename multiple array APIs based on certain conventions

adithyaov:
break -> into 2 parts
split -> into many parts
strip* -> dropWhile*

### DIFF
--- a/core/src/Streamly/Internal/Data/Array/Type.hs
+++ b/core/src/Streamly/Internal/Data/Array/Type.hs
@@ -37,7 +37,7 @@ module Streamly.Internal.Data.Array.Type
     , unsafeAsForeignPtr
 
     -- * Subarrays
-    , unsafeGetSlice
+    , unsafeSliceOffLen
 
     -- ** Construction
     , empty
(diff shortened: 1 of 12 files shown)


core/src/Streamly/Internal/Data/Array/Type.hs
```diff
@@ -36,23 +36,26 @@ module Streamly.Internal.Data.Array.Type
     , unsafePinnedAsPtr
     , unsafeAsForeignPtr
 
+    -- * Subarrays
+    , unsafeGetSlice
```
harendra-kumar, line 40: Should be `unsafeSliceAtLen`.

core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -68,17 +68,17 @@ module Streamly.Internal.Data.MutArray.Type
     -- | Get a subarray without copying
     , unsafeGetSlice -- XXX unsafeSliceAtLen
     , getSlice -- XXX sliceAtLen
```
harendra-kumar, line 70: Rename the above two as well. 

Also, do not deprecate if the API was introduced in this release itself. You can check the packdiff output.

harendra-kumar (approved):

PR #2953, opened 2025-02-09
Refactor, add some Array/MutArray APIs, add appendCString to Path, INLINE some Path functions

harendra-kumar:


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -204,28 +219,64 @@ module Streamly.Internal.Data.MutArray.Type
...
+    -- | TODO: We can replace memcpy with stream copy using Word64. Arrays are
+    -- aligned on 64-bit boundaries on 64-bit CPUs. A fast way to copy an
+    -- array is to unsafeCast it to Word64, read it as a stream, write the
+    -- stream to Word64 array and unsafeCast it again. We can use SIMD
+    -- read/write as well.
+
+    , spliceCopy -- XXX freeze and splice instead?
     , splice
-    , spliceExp
+    , spliceWith -- XXX spliceGrowWith
```
adithyaov, line 276: Is `Grow` required in the name.
We're making the name longer.
Isn't `Grow` implicit?

adithyaov (approved):

core/src/Streamly/Internal/Data/MutArray/Type.hs
harendra-kumar: spliceWith sounds like splice an array with another - which is also implicit. So spliceGrowWith is more explicit to say that we are specifying a function for grow behavior.

PR #2967, opened 2025-02-11
Add lazy pinned mutvar primitives

harendra-kumar:


core/src/Streamly/Internal/Data/IORef/Unboxed.hs
```diff
@@ -59,6 +62,27 @@ newIORef x = do
...
+-- benefits. It is pure so that we can declare it within a global or local
+-- scope without a monadic context. In local scopes it may be better to use the
+-- IO monad instead of declaring mutvars.
+--
+-- Also see "Streamly.Internal.Data.MutByteArray" for more versatile but less
+-- safe mutable cells for variable length data structrures. See
+-- "Streamly.Internal.Data.MutArray" for array type mutvars.
+--
+{-# NOINLINE mutVar' #-}
+mutVar' :: Unbox a => a -> IORef a
```
adithyaov, line 76: We use `'` for pinned in the context of arrays and friends.
After looking at `'`, my first question was "Is this a pinned IORef"?

adithyaov: Now even `modifyIORef'` feels wrong.

harendra-kumar:
Needs more thought for multi-threaded use.

PR #2973, opened 2025-02-13
 Add some API design notes and API skeletons for Dir module 

harendra-kumar:


core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -78,8 +158,47 @@ foreign import capi unsafe "dirent.h fdopendir"
...
-foreign import ccall unsafe "lstat_is_directory"
-    c_lstat_is_directory  :: CString -> IO CInt
+--------------------------------------------------------------------------------
+-- Stat
+--------------------------------------------------------------------------------
+
+foreign import ccall unsafe "stat.h lstat"
+    c_lstat :: CString -> Ptr CStat -> IO CInt
+
+foreign import ccall unsafe "stat.h stat"
```
harendra-kumar, line 168: wondering if this should be `capi` instead?

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -78,8 +158,47 @@ foreign import capi unsafe "dirent.h fdopendir"
...
+    :: (CString -> Ptr CStat -> IO CInt) -> CString -> IO (Either Errno CMode)
+statWith f cstr =
+    allocaBytes #{size struct stat} $ \p_stat -> do
+        resetErrno
+        result <- f cstr p_stat
+        if result /= 0
+        then do
+            errno <- getErrno
+            pure $ Left errno
+        else do
```
harendra-kumar, line 195: we need to handle EINTR here?

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -138,41 +257,64 @@ isMetaDir dname = do
...
             case res of
-                x | x == 1 -> pure (True, False)
-                x | x == 0 -> pure (False, False)
-                -- XXX Need to check if and how we should handle some errors
-                -- like EACCES.
-                _ -> throwErrno "checkIfDirectory"
+                Left errno ->
+                    if    (errno == eNOENT && _ignoreNonExistingFiles conf)
+                       || (errno == eACCES && _ignoreInAccessibleFiles conf)
+                       || (errno == eLOOP && _ignoreSymlinkLoopErrors conf)
```
harendra-kumar, line 286: this is technically not efficient - even if errno was eNOENT  it cannot be eACCES but you might still check whether it is eACCES or not.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -138,41 +257,64 @@ isMetaDir dname = do
...
     else do
         -- XXX We can create a pinned array right here since the next call pins
         -- it anyway.
         path <- appendCString parent dname
         Array.asCStringUnsafe (Path.toChunk path) $ \cStr -> do
-            res <- c_lstat_is_directory cStr
+            res <-
+                if _followSymlinks conf
+                then statWith c_stat cStr
+                else statWith c_lstat cStr
```
harendra-kumar, line 281: you can just pass the flag to statWith and decide to use c_stat or c_lstat inside it. that will generate slightly better code.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -138,41 +257,64 @@ isMetaDir dname = do
...
+checkDirStatus conf parent dname dtype =
     if dtype == (#const DT_DIR)
     then do
         isMeta <- liftIO $ isMetaDir dname
-        pure (True, isMeta)
-    else if dtype /= #const DT_UNKNOWN
-         then pure (False, False)
-         else lstatDname parent dname
+        pure $ if isMeta then GSIsMetaDir else GSIsRegDir
+    else if dtype == (#const DT_LNK) && _followSymlinks conf
```
harendra-kumar, line 313: keep _followSymlinks inside the if statement instead of using `&&`.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -205,13 +350,13 @@ readDirStreamEither (curdir, (DirStream dirp)) = loop
...
-            if isMeta
-            then loop
-            else return (Just (Left (mkPath name)))
-        else return (Just (Right (mkPath name)))
+        gsRes <- checkDirStatus conf curdir dname dtype
+        case gsRes of
+            GSIsRegDir -> return (Just (Left (mkPath name)))
+            GSIsMetaDir ->  loop
+            GSIsNotDir -> return (Just (Right (mkPath name)))
+            -- We can possibly handle the error here
```
harendra-kumar, line 358: why this comment? we have already handled/thrown the error.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
adithyaov: `||` is lazy in nature right?
`True || x` = True

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
adithyaov: Any particular reason why?

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
adithyaov: Re-handle it in the context of the function.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
harendra-kumar: That is not the point, the point is - If errno == eNOENT is True and _ignoreNonExistingFiles conf is False we still check the other cases but we should not.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
harendra-kumar: Same reason as the errno case earlier. If dtype == (#const DT_LNK) is True and _followSymlinks conf is False we still go check the else case but we should not, we know it cannot be DT_UNKNOWN.

adithyaov: Now that I look at it again. What was written is incorrect logic. Will push a fix.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -148,41 +269,76 @@ isMetaDir dname = do
...
+                        if _ignoreInAccessibleFiles conf
+                        then pure $ GSIgnoreError errno
+                        else throwErrno (errMsg path)
+                    else if errno == eLOOP
+                    then
+                        if _ignoreSymlinkLoopErrors conf
+                        then pure $ GSIgnoreError errno
+                        else throwErrno (errMsg path)
+                    else throwErrno (errMsg path)
+                Right mode -> pure $
```
harendra-kumar, line 311: In general, in if or case statements, better keep the most common case as the first case, in case the compiler generates code assuming that the first branch is more likely.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -148,41 +269,76 @@ isMetaDir dname = do
...
+                    then
+                        if _ignoreInAccessibleFiles conf
+                        then pure $ GSIgnoreError errno
+                        else throwErrno (errMsg path)
+                    else if errno == eLOOP
+                    then
+                        if _ignoreSymlinkLoopErrors conf
+                        then pure $ GSIgnoreError errno
+                        else throwErrno (errMsg path)
+                    else throwErrno (errMsg path)
```
harendra-kumar, line 310: You could use `when` to throw the errno. And in the end use `pure $ GSIgnoreError errno`, that will generate just one branch in the generated code instead of three branches.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -148,41 +269,76 @@ isMetaDir dname = do
...
+                Right mode -> pure $
+                    if (mode == s_IFDIR)
+                    then GSIsRegDir
+                    else GSIsNotDir
+    where
+    errMsg path =
+        let pathStr = Path.toString path
```
harendra-kumar, line 317: this may fail on rogue paths or in case of bugs. use Path.toString_ instead.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
harendra-kumar: is there an example case and possible re-handling?

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.c
```diff
@@ -1,18 +0,0 @@
-#include <sys/stat.h>
-
-int lstat_is_directory(const char *path) {
-    struct stat statbuf;
-
-    // XXX Using fstatat with a dirfd and relative path would be faster.
```
harendra-kumar, line 6: Did we lose this comment? can you transfer at some relevant place.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -360,7 +529,8 @@ readEitherChunks alldirs =
                             (ChunkStreamLoop curdir xs dirp [] 0 files nfiles)
                          else return $ Skip
                             (ChunkStreamLoop curdir xs dirp dirs1 ndirs1 files nfiles)
-            else do
+                GSIsMetaDir ->  return $ Skip st
```
harendra-kumar, line 533: We can use a wildcard case at the end for skipping. In fact we can even combine the two cases of GSIsMetaDir and GSIsNotDir in to a single GSIgnoreEnt.

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
adithyaov: I planned to do this initially but it felt non-functional so I let it be.

core/src/Streamly/Internal/FileSystem/Dir.hs
```diff
@@ -123,14 +123,20 @@ pMapUnfoldE = fmap ePathMap . Unfold.lmapM Path.fromString
 -- Functions
 --------------------------------------------------------------------------------
 
+#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
+#define CONF id
```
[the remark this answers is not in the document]
adithyaov: See #2991

core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -148,41 +273,70 @@ isMetaDir dname = do
...
+         in "statDname: " ++ pathStr
 
 -- | Checks if dname is a directory and additionaly returns if dname is a meta
 -- directory.
 {-# INLINE checkDirStatus #-}
 checkDirStatus
-    :: PosixPath -> Ptr CChar -> #{type unsigned char} -> IO (Bool, Bool)
+    :: ReadOptions
+    -> PosixPath -> Ptr CChar -> #{type unsigned char} -> IO GStatRes
 #ifdef FORCE_LSTAT_READDIR
```
harendra-kumar, line 323: We can remove this flag now.

PR #2981, opened 2025-02-16
Add CI workflow that ensures released pkg builds with latest GHC

adithyaov:


adithyaov:
Running this on every commit is costly, we can possibly set up a cron job?

adithyaov:
```
ghcup list -t ghc -r | tail -n 1 | cut -d ' ' -f2
```
This gives us the latest ghc version.

PR #2985, opened 2025-02-17
Fix readdir FFI issue on macOS

harendra-kumar:


core/src/Streamly/Internal/FileSystem/Posix/ReadDir.hsc
```diff
@@ -59,9 +59,13 @@ data {-# CTYPE "struct dirent" #-} CDirent
 
 newtype DirStream = DirStream (Ptr CDir)
 
+-------------------------------------------------------------------------------
+-- readdir operations
 -------------------------------------------------------------------------------
 
-foreign import ccall unsafe "closedir"
+-- IMPORTANT NOTE: Use capi FFI for all readdir APIs. This is required at
```
adithyaov, line 66: You can explain why a little more here.

harendra-kumar: that may make the note too long and specific, one can read the difference between capi and ccall in GHC manual.

PR #2994, opened 2025-02-19
Use GHC2024 and GHC2021 extensions

harendra-kumar:


core/src/Streamly/Internal/FileSystem/File.hs
```diff
@@ -204,15 +205,15 @@ usingFile3 = UF.bracketIO before after
 -- /Pre-release/
 --
 {-# INLINABLE putChunk #-}
-putChunk :: FilePath -> Array a -> IO ()
+putChunk :: forall (a :: Type). FilePath -> Array a -> IO ()
```
adithyaov, line 208: Why is this important? Can we ignore this using `-Wno-`?

harendra-kumar: It is an error, not a warning.

PR #3007, opened 2025-03-21
Fix a bug in parDistributeScan and block on buffer in sendToWorker_

adithyaov:


src/Streamly/Internal/Data/Fold/Channel/Type.hs
```diff
@@ -510,10 +510,12 @@ sendToWorker_ chan a = go
                     (inputItemDoorBell chan)
                     (ChildYield a)
         else do
-            error "sendToWorker_: No space available in the buffer"
+            -- XXX: Is there a reason for us to error out here instead of
+            -- blocking?
```
harendra-kumar, line 514: this might have been introduced to know that we are actually hitting this condition instead of the possibility of a deadlock due to blocking. If there is a possibility of deadlock then you better ensure that the buffer is always available.

src/Streamly/Internal/Data/Scanl/Concurrent.hs
```diff
@@ -242,7 +242,8 @@ parDistributeScan cfg getFolds (Stream sstep state) =
                 then do
                     liftIO $ takeMVar db
                     return $ Skip (ScanDrain q db running)
-                else return $ Yield outputs (ScanDrain q db running)
+                else return $ Yield outputs ScanStop
```
harendra-kumar, line 245: There may be more outputs pending.

harendra-kumar (changes requested):
Remove the comment at line 348 in Scanl/Concurrent.hs
can we also add a test for parDemuxScan?

harendra-kumar (approved):

PR #3017, opened 2025-05-03
Array api - rename one and add some todos

harendra-kumar:


core/src/Streamly/Internal/Data/MutArray/Type.hs
```diff
@@ -60,27 +61,58 @@ module Streamly.Internal.Data.MutArray.Type
 
     -- *** Slicing
     -- | Get a subarray without copying
+
+    -- Element agnostic.
```
adithyaov, line 65: These are dev comments, right?

adithyaov (approved):
Looks good. This PR is mostly geared towards the dev, except for the renaming.

Issue #2663, opened 2024-01-03
Deprecate the array stream parsing functions from StreamK module


harendra-kumar:
Move them to the Array module.

PR #3025, opened 2025-05-14
Move array stream parsing funcs to array modules

harendra-kumar:
https://github.com/composewell/streamly/issues/2663

### DIFF
--- a/core/src/Streamly/Internal/Data/Array.hs
+++ b/core/src/Streamly/Internal/Data/Array.hs
@@ -70,10 +70,14 @@ module Streamly.Internal.Data.Array
     , compactEndByByte_
     , compactEndByLn_
 
-    , foldBreakChunks
+    -- * Parsing Stream of Arrays
+    , foldBreakChunks -- Uses Stream, bad perf on break
     , foldChunks
-    , foldBreakChunksK
-    , parseBreakChunksK
+    , foldBreakChunksK -- XXX rename to foldBreak
+    , parseBreakChunksK -- XXX uses Parser. parseBreak is better?
+    , parserK
+    , parseBreak
+    , parse
 
     -- * Serialization
     , encodeAs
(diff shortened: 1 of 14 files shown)


core/src/Streamly/Internal/Data/Array.hs
```diff
@@ -70,10 +70,14 @@ module Streamly.Internal.Data.Array
...
+    -- * Parsing Stream of Arrays
+    , foldBreakChunks -- Uses Stream, bad perf on break
     , foldChunks
-    , foldBreakChunksK
-    , parseBreakChunksK
+    , foldBreakChunksK -- XXX rename to foldBreak
+    , parseBreakChunksK -- XXX uses Parser, should use parseBreak instead
+    , parseBreak
+    , parse
+    , adapt
```
adithyaov, line 80: `adapt` seems like an incorrect name now that it is in `Array` module.

core/src/Streamly/Internal/Data/ParserK.hs
```diff
@@ -10,7 +10,30 @@
...
+import qualified Streamly.Internal.Data.Array.Generic as GenArray
+
+#include "inline.hs"
+
+{-# DEPRECATED adaptC "Use Streamly.Data.Array.adapt" #-}
+{-# INLINE_LATE adaptC #-}
+adaptC :: (Monad m, Unbox a) => Parser a m b -> ParserK (Array a) m b
+adaptC = Array.adapt
+
+{-# DEPRECATED adaptCG "Use Streamly.Data.Array.Generic.adaptCG" #-}
```
adithyaov, line 35:
```suggestion
{-# DEPRECATED adaptCG "Use Streamly.Data.Array.Generic.adapt" #-}
```

core/src/Streamly/Internal/Data/ParserK/Type.hs
```diff
@@ -27,13 +27,13 @@ module Streamly.Internal.Data.ParserK.Type
     , Input (..)
     , ParseResult (..)
     , ParserK (..)
-    , adaptC
-    , adapt
-    , adaptCG
+    , adapt -- XXX rename to parserK, like in Arrays?
```
adithyaov, line 30: `ParserK.parserK` sounds weird. In this case, we can name it to `fromParser`

adithyaov (approved):

PR #3034, opened 2025-05-18
Use separate Config type for stream/fold channels

harendra-kumar:


adithyaov:
Shouldn't `Fold.Concurrent` module also have changes?
You'll need to deprecate the exposed functions.

harendra-kumar:
> Shouldn't `Fold.Concurrent` module also have changes? You'll need to deprecate the exposed functions.

what changes? It is using the Config type from Fold.

Which function is to be deprecated? We are still using the same functions except that Fold and stream are using separate definitions.

adithyaov:
Oh nvm. I misunderstood. This looks good to me.

adithyaov (approved):

Issue #2964, opened 2025-02-10
Constraint the type in `pollIntIORef` to `Int`


adithyaov:
The `Int` in `pointIntIORef` signifies that if the size of the datatype is `<=
Int` then the operations are safe as they are atomic in nature.

Any type bigger than `Int` can corrupt memory if writing and polling is done
concurrently.

A better way is to restrict the type to `Int`. Any type with size `<= Int` can
be coerced from/to `Int` safely.

PR #3036, opened 2025-05-20
Change the signature of pollIntIORef

harendra-kumar:
issue https://github.com/composewell/streamly/issues/2964

### DIFF
--- a/core/src/Streamly/Internal/Data/IORef.hs
+++ b/core/src/Streamly/Internal/Data/IORef.hs
@@ -32,11 +32,16 @@ module Streamly.Internal.Data.IORef
 
     -- Read
     , readIORef
+    , pollGenericIORef
+    , pollIORefInt
+
+    -- Deprecated
     , pollIntIORef
     )
 where
 
 #include "inline.hs"
+#include "deprecation.h"
 
 import Control.Monad.IO.Class (MonadIO(..))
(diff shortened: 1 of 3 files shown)


core/src/Streamly/Internal/Data/IORef.hs
```diff
@@ -88,20 +93,30 @@ modifyIORef' var g = do
...
--- access the memory. It may read stale values though until caches are
--- synchronised in a multiprocessor architecture.
---
--- /Pre-release/
-{-# INLINE_NORMAL pollIntIORef #-}
-pollIntIORef :: (MonadIO m, Unbox a) => IORef a -> D.Stream m a
-pollIntIORef var = D.Stream step ()
+-- | Internal, do not use.
+{-# INLINE_NORMAL pollGenericIORef #-}
+pollGenericIORef :: (MonadIO m, Unbox a) => IORef a -> D.Stream m a
```
adithyaov, line 98: Why do we need a generic version? Just the `Int` version should be enough. Any acceptable generic value should be coerced to `Int`.

harendra-kumar: To suppress the deprecation warning in the MutArray module. This definition is being used there.

adithyaov (approved):

PR #3050, opened 2025-06-21
Split the test:Data.Parser into 2 modules

adithyaov:


test/streamly-tests.cabal
```diff
@@ -288,6 +288,8 @@ test-suite Data.Parser
   import: test-options
   type: exitcode-stdio-1.0
   main-is: Streamly/Test/Data/Parser.hs
+  hs-source-dirs: .
+  other-modules: Streamly.Test.Data.Parser.Common
   if flag(limit-build-mem)
     ghc-options: +RTS -M4000M -RTS
```
harendra-kumar, line 294: this 4GB limit should be reduced after the split.

PR #3060, opened 2025-06-25
Add more powerful cleanup functions

harendra-kumar:
So that we can release resources any time.
Fix atomicity issue during allocation.
Rename the APIs.


core/src/Streamly/Internal/Data/Stream/Exception.hs
```diff
@@ -350,17 +354,81 @@ bracketIO3 bef aft onExc onGC =
...
+        (r, index) <- liftIO $ mask_ $ do
+            r <- alloc
+            idx <- atomicModifyIORef ref (\(i, mp) ->
+                ((i + 1, Map.insert i (void $ free r) mp), i))
+            return (r, idx)
+
+        let free1 =
+                atomicModifyIORef ref (\(i, mp) ->
+                    ((i, Map.delete index mp), ()))
+        return (r, free1)
```
adithyaov, line 401: This only deletes the registered free function at the index.
Shouldn't it also run the `free` function?

adithyaov: Also, I don't see how we can call the free function manually. Can you give an example or use it in the tests?

adithyaov (approved):

core/src/Streamly/Internal/Control/Exception.hs
```diff
@@ -165,54 +186,110 @@ newtype AllocateIO = AllocateIO
...
-            return (r, idx)
-
-        let modify (i, mp) =
-                let res = Map.lookup index mp
-                 in ((i, Map.delete index mp), res)
-
-            release = do
-                f <- atomicModifyIORef ref modify
-                sequence_ f
+    -- XXX can we ensure via GC that the resources that we are freeing are all
```
adithyaov: So we don't support parallel release yet.
PS: I'm reviewing the commits one by one. This commit may be obsolete.

core/src/Streamly/Internal/Control/Exception.hs
```diff
@@ -165,54 +186,110 @@ newtype AllocateIO = AllocateIO
...
+                    Priority1 ->
+                        let res = Map.lookup index mp1
+                         in ((i, Map.delete index mp1, mp2), res)
+                    Priority2 ->
+                        let res = Map.lookup index mp2
+                         in ((i, mp1, Map.delete index mp2), res)
+
+            release =
+                -- IMPORTANT: do not use interruptible operations in this
+                -- critical section. Even putStrLn can make tests fail.
```
adithyaov: What exactly are interruptible operations?
Operations that allow the thread to sleep?

core/src/Streamly/Internal/Control/Exception.hs
```diff
@@ -165,54 +186,110 @@ newtype AllocateIO = AllocateIO
...
+                liftIO $ mask_ $ do
+                    -- tid <- myThreadId
+                    -- liftIO $ putStrLn $ "releasing index: " ++ show index
+                    --      ++ " " ++ show tid
+                    f <- atomicModifyIORef' ref deleteResource
+                    -- restoring exceptions makes it non-atomic, tests fail.
+                    -- Can use allowInterrupt in "free" if desired.
```
adithyaov: I don't follow this.

core/src/Streamly/Internal/Data/Stream/Exception.hs
```diff
@@ -650,26 +650,35 @@ withAllocateIO action = do
...
 
         let modify (i, mp) =
                 let res = Map.lookup index mp
                  in ((i, Map.delete index mp), res)
             free1 = do
-                res <- atomicModifyIORef ref modify
+                -- We run it without an exception mask unlike withAllocateIO in
+                -- the Streamly.Control.Exception module. We can add it if
+                -- needed.
+                res <- atomicModifyIORef' ref modify
```
adithyaov: So, async exceptions can trigger this?

src/Streamly/Internal/Data/Stream/Channel/Type.hs
```diff
@@ -619,7 +625,7 @@ _getBound = _bound
 -- simpler code.
 --
 setReleaseCb :: AllocateIO -> Config -> Config
-setReleaseCb f cfg = cfg { _release = Just (register f) }
+setReleaseCb f cfg = cfg { _release = Just (registerWith f Priority1) }
```
adithyaov: The callback registered in the channel config will always be called first, and then the other callbacks.
The user has to be wary about resource cleanup and decide what should go where.

src/Streamly/Internal/Data/Stream/Channel/Type.hs
```diff
@@ -820,13 +826,30 @@ dumpChannel sv = do
...
+        -- prompt them to be cleaned up quickly.
+        performMajorGC
+
+-- XXX Handle ThreadAbort while this is running?
+
+-- | Never called from a worker thread. This can be called multiple times and
+-- only one will succeed others will block until it finishes.
+--
+cleanupChan :: Channel m a -> String -> IO ()
+cleanupChan chan reason = do
```
adithyaov, line 844: There will always be a major GC performed before the channel stops.
It may not be required in many cases where there isn't any resource allocation.
Is it possible to `performMajorGC` only when required?

core/src/Streamly/Internal/Control/Exception.hs
```diff
@@ -103,6 +106,85 @@ data Priority = Priority1 | Priority2 deriving Show
...
+--
+-- The only other possibility, could be user issued forkIO not being
+-- tracked by us, however, that would be a programming error and any such
+-- threads could misbehave if we freed the resources from under them.
+--
+-- We use GC based hooks in Stream.bracketIO' so there could be async threads
+-- spawned by GC, releasing resources concurrently with us. For that reason we
+-- need to make sure that the "release" in the bracket end action is executed
+-- only once in that case.
+--
```
adithyaov: Is it possible to remove GC hooks? Now that we have prompt cleanup why do we still have them?

core/src/Streamly/Internal/Control/Exception.hs
```diff
@@ -86,11 +86,12 @@ verifyM predicate = verify predicate (pure ())
 -- a limited GC to free them? We can then just put gc sync barriers at points
 -- where we want to ensure that resources are freed.
 
--- | Resources with Priority1 are freed before Priority2. This is especially
--- introduced to take care of the case where we need to free channels, so that
--- all the workers of the channel are cleanup before we free the resources
--- allocated by the workers of the channel. Otherwise we might free the
--- resources and workers may be trying to use them and start misbehaving.
+-- | Resources with 'Priority1' are freed before 'Priority2'. Priority is
```
adithyaov, line 92: Can you also give a small example on how this helps in freeing concurrency channels?

harendra-kumar: A stream is being evaluated concurrently by multiple workers, each worker allocates some resource and is working with those resources. An interrupt comes and cleanup starts. Assume, we cleanup the resources of the workers first and then the worker. Now the worker is still working so it might crash or misbehave of because of the resource slipping from under it. So we need ensure that we call the cleanup of the workers before we cleanup their resources.

src/Streamly/Internal/Data/Stream/Channel/Type.hs
harendra-kumar: performMajorGC is done only in inspect mode so that logs come properly.

src/Streamly/Internal/Data/Stream/Channel/Type.hs
harendra-kumar: only streamly developers need to worry about the priority. For normal users there is one priority.

core/src/Streamly/Internal/Control/Exception.hs
harendra-kumar: to cleanup even before the "prompt" release happens. prompt release can be later than GC.

PR #3088, opened 2025-07-08
Remove cabal flags while releasing

adithyaov:


.github/workflows/haskell.yml
```diff
@@ -207,6 +212,7 @@ jobs:
             cabal_version: 3.6.2.0
             subdir: core
             ignore_error: false
+            rewrite_flags: dev use-unliftio debug
```
harendra-kumar, line 215: should not remove `debug`, it can be used for debugging problems in production. Remove dev flags from both streamly and streamly-core. Can also remove `inspection`, `limit-build-mem`, and `opt`, `use-unfolds`, `use-folds`, `force-lstat-readdir`.

harendra-kumar (changes requested):
Instead of keeping it in streamly repo, we can put it in a separate repo and host a static binary in that repo, download and use it from there.

adithyaov:
This is making the maintenance process unnecessarily complex. Ensuring the build works with the new generated cabal file involves building all tests and benchmarks, as well as verifying that existing flags are functional.

There is no easy way to make sure that the newer cabal file is the same as the stripped version of the older cabal file.

Instead, it may be easier to prefix the flags with `INTERNAL_`. This makes it explicit and cautions the user, albeit it is a little ugly.

adithyaov:
Too complex for something so simple :-)

PR #3090, opened 2025-07-09
Make some Posix/Windows specific APIs portable in Path

harendra-kumar:


core/src/Streamly/Internal/FileSystem/PosixPath.hs
```diff
@@ -234,16 +258,21 @@ For APIs that have not been released yet.
...
--- A OS_PATH is validated before construction unless unsafe constructors are
+-- A OS_PATH_TYPE is validated before construction unless unsafe constructors are
 -- used to create it. For validations performed by the safe construction
 -- methods see the 'fromChars' function.
 --
 -- Note that in some cases the file system may perform unicode normalization on
 -- paths (e.g. Apple HFS), it may cause surprising results as the path used by
 -- the user may not have the same bytes as later returned by the file system.
-newtype OS_PATH = OS_PATH (Array WORD_TYPE)
+newtype OS_PATH = OS_PATH (Array OS_WORD_TYPE)
```
adithyaov, line 275: Do we define a newtype in the `Path` module? The portable module should not have this right?

adithyaov: Oh nvm. This is under `else` branch.

core/src/Streamly/Internal/FileSystem/PosixPath.hs
```diff
@@ -289,14 +319,14 @@ adapt p = fromPath (toPath p :: OS_PATH)
 -- "/"
 --
 {-# INLINE dropTrailingSeparators #-}
-dropTrailingSeparators :: OS_PATH -> OS_PATH
+dropTrailingSeparators :: OS_PATH_TYPE -> OS_PATH_TYPE
```
adithyaov, line 322: The portable module could have used the definition directly, but it may not be possible to implement that in an easy way.

adithyaov (approved):
LGTM. Don't think there is a better way to do this.

PR #3093, opened 2025-07-11
Rename some and fix docs, doctests of Path APIs

harendra-kumar:


bench-test-lib/src/BenchTestLib/DirIO.hs
```diff
@@ -165,7 +164,7 @@ listDirChunkedWith
     -> [Char] -> Stream IO Word8
 listDirChunkedWith act inp = do
      Stream.unfoldEachEndBy 10 Array.reader
-        $ fmap (Array.asBytes . Path.toChunk)
+        $ fmap (Array.asBytes . Path.toArray)
```
adithyaov, line 167: `Chunk` is analogous to `Array` in the library.

core/src/Streamly/Internal/FileSystem/Path/Common.hs
```diff
@@ -19,8 +19,8 @@ module Streamly.Internal.FileSystem.Path.Common
     , validateFile
 
     -- * Construction
-    , fromChunk
-    , unsafeFromChunk
+    , fromArray
+    , unsafeFromArray
```
adithyaov, line 23: I think we should leave `*Chunk` as is.

PR #3113, opened 2025-07-22
Expose Cross and FairCross types from StreamK

harendra-kumar:


core/src/Streamly/Internal/Data/StreamK/Type.hs
```diff
@@ -1505,6 +1599,153 @@ mergeMapWith combine f str = go (leafPairs str)
...
+-- Compare with 'concatMapWith interleave' which explores the depth
+-- exponentially more compared to the breadth, such that each stream yields
+-- twice as many items compared to the next stream.
+--
+{-# INLINE concatMapDiagonal #-}
+concatMapDiagonal ::
+       (a -> StreamK m b)
+    -> StreamK m a
+    -> StreamK m b
+concatMapDiagonal f m1 = go id m1
```
adithyaov, line 1711: This is a pretty clever implementation for a diagonal traversal.

adithyaov (changes requested):
We should remove the monad instances wherever associativity isn't followed.

PR #3119, opened 2025-07-25
Add more general version of onException

jerbaroo:


adithyaov:
```
onExceptionE f = handle (\e -> f e >> Stream.nil)
```

adithyaov:
@jerbaroo `onExceptionE` does not appear to be a standard exception handling function and may not be a suitable addition to the library.

jerbaroo:
Solution: https://github.com/composewell/streamly/discussions/3117#discussioncomment-13894572

PR #3217, opened 2026-05-13
Fix race condition in sendExceptionToDriver

harendra-kumar:
We were ringing the doorbell too early even before we sent the message, so the doorbell could go waste if the reader woke up before the message was sent.


harendra-kumar:
@johnhampton the deadlock issue was not completely fixed earlier. The test was running into a deadlock in the CIs, this PR fixes the issue. You may want to use this if you are using it in production.

johnhampton:
@harendra-kumar thank you!

harendra-kumar:
Uploaded 0.11.1 on hackage with the fix.

PR #3223, opened 2026-05-27
Make the inject function of Unfold pure

harendra-kumar:


harendra-kumar:
lmapM shows big performance regression, dropping the idea of pure inject in Unfold.

